PR #9, opened 2015-03-31
Swift 1.1

petester42:
Since Xcode 6.3 and Swift 1.2 are still in beta this lovely library cannot be used when submitting to the store. This is a stopgap before Xcode 6.3 is final and could be added as a legacy branch.

Changes:
- Changed `@objc public enum` to `NS_Enum`
- Changed 1.2 specific syntax such as `as!`
- Removed `__nonnull` annotations
- Updated readme


danielgindi:
.Well this is nice for those who are eager to send it to the AppStore already,
But will mess up the code for backwards support for an incomplete language...

I will keep this PR live so other people know about your fork, but in the meantime I will continue developing so bear in mind you'll have to update to Xcode 6.3 in April 24. That's the release date for Xcode 6.3 :)

Btw, I would love to know about apps using this library!

petester42:
Sounds good.

The reason I did this is that I plan on working on a project that will use this but Xcode 6.3 might not be out by the time I need a release. I doubt 6.3 will be available on the 24th, wouldn't make much sense. Xcode 6.2 and iOS 8.2 were for the watch and are already live. I hope I'm wrong but without knowing for sure better safe than sorry.

danielgindi:
Oh they did make a lot of important improvements for WatchKit in Xcode 6.3
so it is most likely tied with the Watch release :-)

Also it fits the usual timeline for Xcode releases of about a month between
beta releases :-)

vishw3:
@danielgindi   Are u sure 6.3 will be released on april 24th.

danielgindi:
Well, known facts are on my side:
- They do have an average of ~30 days per beta:  http://www.thinkybits.com/blog/iOS-versions/
- And it's not a major version, which means they are not doind 7 betas, just about 3-4
- It has improvements for WatchKit. i.e:
  -   A crucial bugfix for Xcode: "When launching a WatchKit app on the iOS Simulator, the Watch Simulator may not automatically launch."
- Most of the bug fixes in the OS are minor: https://developer.apple.com/library/prerelease/ios/releasenotes/General/RN-iOSSDK-8.3/index.html

So I don't know if it's April 24th, but I'm pretty sure it's April-something.

bach942:
Be able to use Swift 1.1 was also a concern when I was looking for charting tools for iOS. This seems to have many features that I'm looking for and as a bonus there is an Android version.

petester42:
Since 6.3 is out. There is no need for this anymore. I will keep the branch in case anyone needs it but they really shouldn't anymore.

PR #13, opened 2015-04-07
Added a podspec

petester42:
I've added a podspec since I will be using this with cocoapods. I have not pushed it to the repo that manages pods just yet since the current specification is not complete. Cocoapods requires that a version be tagged in order to work properly and be able to distinguish between versions. I have done this in my own fork and used my fork as the source for the pod but I'd like it to point to the original source.

Once you tag a version, I will update this line: `s.source = { :git => "https://github.com/petester42/ios-charts.git", :tag => "0.0.1" }` to point to here and then you can merge.

I also tried making sure carthage support works but the way the current project is setup makes it so that the schemes are not there when running carthage.

### DIFF
--- a/Charts.podspec
+++ b/Charts.podspec
@@ -1,0 +1,12 @@
+Pod::Spec.new do |s|
+  s.name = "Charts"
+  s.version = "2.0.9"
+  s.summary = "ios-charts is a powerful & easy to use chart library for iOS"
+  s.homepage = "https://github.com/danielgindi/ios-charts"
+  s.license = { :type => "Apache License, Version 2.0", :file => "LICENSE" }
+  s.authors = "Daniel Cohen Gindi", "Philipp Jahoda"
+  s.ios.deployment_target = "8.0"
+  s.source = { :git => "https://github.com/danielgindi/ios-charts.git", :tag => "v2.0.9" }
+  s.source_files = "Classes", "Charts/Classes/**/*.swift"
+  s.frameworks = "Foundation", "UIKit", "CoreGraphics"
+end


danielgindi:
Please clean this pull request first

petester42:
all cleaned up

danielgindi:
Ok man now you can update the spec to this:
```
Pod::Spec.new do |s|
  s.name         = "Charts"
  s.version      = "2.0.9"
  s.summary      = "ios-charts is a powerful & easy to use charting library for iOS"
  s.homepage     = "https://github.com/danielgindi/ios-charts"
  s.license      = { :type => "Apache License, Version 2.0", :file => "LICENSE" }
  s.authors      = "Daniel Cohen Gindi", "Philipp Jahoda"
  s.platform     = :ios
  s.ios.deployment_target = '7.0'
  s.source       = { :git => "https://github.com/danielgindi/ios-charts.git", :tag => "v2.0.9" }
  s.source_files = "Classes", "Charts/Classes/**/*.swift"
  s.frameworks = 'Foundation', 'UIKit', 'CoreGraphics'
end
```
Squash the commits together and I'll pull it.

petester42:
Sadly iOS 7 is not supported for swift frameworks. It gives the following error:

` -> Charts (2.0.9)
    - ERROR | [iOS] Swift support uses dynamic frameworks and is therefore only supported on iOS > 8.`

It is also detailed on the cocoa pods blog under the availability section of this article http://blog.cocoapods.org/Pod-Authors-Guide-to-CocoaPods-Frameworks/

I will make the other changes but it has to stay iOS 8 for the pod.

danielgindi:
http://stackoverflow.com/questions/24001778/do-swift-based-applications-work-on-os-x-10-9-ios-7-and-lower

Are you sure it's not supported?

petester42:
It's not a swift issue. Swift will work fine on iOS 7. The problem is the framework. Dynamic frameworks are only supported as of iOS 8. This builds a dynamic framework so no iOS 7 love :(. If someone want to use this on iOS 7 they will need to copy the files to their project and not use the framework.

danielgindi:
Ok I see!
Thanks!
Merged.

petester42:
I've pushed the pod to trunk. If you want to be added as an owner of the pod just register with cocoapods `pod trunk register email 'name'` and i'll add you as an owner.

PR #20, opened 2015-04-10
Add missing UIKit

mkalmes:
Importing this class into an iOS 7 project raised an error for unresolved
identifier `CGFloat`.

This pull request fixes [the bug](https://github.com/danielgindi/ios-charts/issues/19).


danielgindi:
Actually the correct framework to import there is CoreGraphics, as there's no use of UIKit in the data classes

danielgindi:
But I've decided to accept it as it is- as there is usage of the UIColor class in some of those classes!

PR #22, opened 2015-04-10
Shared Charts.framework scheme

zenkimoto:
Added Charts.framework shared scheme for Carthage support.


danielgindi:
What are those "version" properties on the root node? Shouldn't they match the framework version and build version?

zenkimoto:
Hi @danielgindi 
I'm going to guess that it's Xcode's scheme file format version.  Also, the LastUpdateCheck is the last known object version of the xcodeproj.  See here:  https://github.com/CocoaPods/Xcodeproj/pull/253/files

danielgindi:
I'm not familiar with Carthage - Just wanted to make sure that it knows what version it is and if/how we should update version numbers for it :-)

PR #25, opened 2015-04-13
Added Bubble Chart Type

petester42:
I saw that this was one of the issues in the android version and I need it for my project. Might think about porting it to android as well.


danielgindi:
Thank you for this contribution! :-)

Please make a PR against the bubble-chart branch, so I can investigate it and make adjustments as necessary...

danielgindi:
OK I've pulled it into a separate branch - will work on it and push to the master when I see that it's ready!

danielgindi:
An update: 
It actually seems pretty clean to me - good work! I haven't touched a line.
But there's a showstopper - your files are copyrighted. They need to be under a compatible open source license...

I'm deleting my branch, waiting to pull this request when it's ready...

petester42:
Will try to do this today. I really wish there was an easy way to create new files in Xcode without these comments.

petester42:
done ;)

danielgindi:
Just one question - Why did you choose to set a separate property for bubble size? Isn't value always the same as size?

petester42:
I can understand your confusion. The difference is that the size is the number of items that belong to the corresponding x and y values while the bubble size is the actual size of the bubble on screen. I could have named size better I guess since it actually represents the weight of that data. 

Lets say you have data x, y, z where x is the survey questions 1-5, y is the score from 1-5 and z is the number of people that answered. So if there is 2 data points of (1,5,10) and (2,1,100) then what would the graph look like? It should look like a point at (1,5) and a point at (2,1) where the point at (2,1) is 10 times bigger than (1,5). Since there is no z axis you can't say that the size of the bubble will be 10 and 100 because what does that mean (10 and 100 px or something else). The bubble chart is supposed to indicate visually the different weights for the data points. So what I did was consider that the width and height of the chart would be the scale for the bubble. If you take that into consideration making the size 10 px and 100 px would make them very small on the chart. I used bubble size instead so that the data can be represented as big as possible within the graph while keeping the ratio between them the same. So for this example the bubble size might be 30px (for the 10 value) and 300px (for the 100 value). This represents the values much better visually and it abstracts away the concept that the size is linked to pixels on screen which shouldn't be the case anyways. Calculating the bubble size makes sure that they will always fit on screen even if the values have a big difference. If you have 1 and 10 or 100 and 1000 the bubble sizes should be the same on the graph since the second value is 10 times bigger in both cases.

Hopefully that helps. Size should probably be changed to weight. Makes it easier to understand.

danielgindi:
Well I understand the way of calculating the sizes - and we have actually
improved that a little since you have divided the chart's viewport size by
the visible x-values, which would not always give an exact value (Dividing
a float by a whole number), and we took the width of one X value from the
viewport transformer.

Anyway you mean that size is an extra dimension for the value? As in having
a value rendered as a label, and another value as the bubble?
Because if that's not the case, I don't see a use case for `size`, as in
the example we pass `size` the same value as we pass to `value`.
And if we want two different DataSets, then the size is normalized to the
dataset anyway...

petester42:
Yes. Size is a third dimension to the data. It should not just be the value of `value`. You can just read the intro paragraph on wikipedia (http://en.wikipedia.org/wiki/Bubble_chart). It it explains it fairly well.

danielgindi:
Okay I got it, that's what I thought. The Combine Chart example should be
adjusted to reflect the third dimension, not just duplicate the y-vals. I
just remembered the Phil is the one who added the bubbles to Combined
chart, and he said he is not very familiar with the bubbles.
I just got confused for a second :-)

petester42:
I was a bit confused too when I was trying to implement it :p

kajensen:
I'd like to use the bubble chart as a sort of scatter plot with a 3rd dimension, the size of the bubbles being its own separate value. But this is not possible with my data because the number of bubbles in each x category must be the same. Why is that?

petester42:
That isn't the case. If you look at the screenshot for bubble chart in the readme it clearly shows different bubble sizes for the same x value.

kajensen:
Sorry my question is poorly worded, I see that now. I want the ability to have a different # of bubbles for each category (color in the example)- 5 red, 3 yellow, 2 green. Right now each dataset must have the same number of y values. It actually looks like this is also the case for the scatter plot. Just curious why that is.

PR #66, opened 2015-05-05
Fix x-axis limit line render issue.

mkubenka:
Bottom position of limit line on x-axis was incorrectly set to y coordinate instead of x.

Current:
out

Fixed:
out-correct


danielgindi:
Thanks! :-)

PR #74, opened 2015-05-06
Changing iOS deployment target to 8.0 from 8.1

MichaelMcGuire:
The project has an iOS Deployment Target set to 8.1 which prevented me from using in a project targeting 8.0 via [Carthage](https://github.com/Carthage/Carthage).  I tested the library on a device using iOS 8.0 and it seemed to work fine.  In addition, the podspec specifies 8.0, so this updates the project deployment target to 8.0 from 8.1.


petester42:
Why was the project target changed from 8.0 to 8.1. Couldn't the iOS Deployment Target be set to 8.0 instead of 8.1?

MichaelMcGuire:
@petester42 You have it backwards.  It used to be set to 8.1, now it is set to 8.0 (on master, at least)

petester42:
Oops. Ya you're right.

PR #76, opened 2015-05-07
added requieredAxisLabelModulus to ChartXAxis

leoMehlig:
the requieredAxisLabelModulus variable makes sure that labels at certain indexes
are always drawn if possible.
E. g. if the chart wants to show a few
days with hours, but always wants to show the start of a day. 
In this
case requieredAxisLabelModulus would be set to 24 so that at least
every 24th label is drawn.


danielgindi:
I cannot accept this PR.

A) It's not clear this actually solves your problem. It still doesn't allow showing "a few days + all start of days" as I understand as your requirement.
B) Whatever you are doing there - isn't really needed as a feature of the Chart, as you could just calculate whatever modulus you want and set to the property.
C) Your code styling is bad and does not match it's surroundings.
D) It's a huge typo...

PR #93, opened 2015-05-19
Allow access to setLabelsToSkip from Objective-C.

mkubenka:


danielgindi:
Thanks :+1: 
I hope that the next Swift update will prevent the need for these bridging hacks...

Issue #89, opened 2015-05-15
calcMinMax() during scrolling / panning


AlBirdie:
For finance charts it is basically the industry standard to constantly adjust the chart's min and max values for the currently visible X axis range.

That would mean that we'd need to call calcMinMax with the chart's lowestVisibleXIndex and highestVisibleXIndex instead of just looping from 0 to data.count. 
In order not to break any existing implementations, that would ideally be configurable with a boolean.

@danielgindi In case you consider this to be included into the library, I'd be happy to implement it on the current master branch and create a pull request for it (same for Android, but that would have to wait a couple of days). I'm just hoping that this doesn't put too much effort on the CPU, but I reckon that'll be fine as long as the datasets aren't that large.

danielgindi:
@AlBirdie That sounds great!

If you could do a PR just a few notes:
1. Keep it as a feature toggled by a boolean, as you said
2. Do performance benchmarks on different datasets - before and after, so we'd know it it has affected performance and how... It would be great to include notes about performance in the documentation of the feature!

I appreciate your participation in this project very much! :-)

AlBirdie:
Sure, not a problem. Do you have any preferences as to how to test performance?
I'd probably do a loop that invalidates (hence calling drawRect on the chart eventually) and measure the time it takes to execute drawRect and create an average of that. Sounds reasonable?
I intended to to this for a long time already (mainly for the line chart performance issue), but haven't gotten around it, sorry for that. Now that the main implementation is done I can finally work on that.

danielgindi:
That sounds good. Actually if you have the time to implement Tests based on
the test platform that's already in Xcode it will be the best as in one
click you could run a test without all the hassle..

AlBirdie:
Just a quick update, unfortunately it's gonna take a while for me to finish work on this, got a lot of other things to do this week. But we'll get there eventually. ;)

danielgindi:
Ok it was merge, along with some bugfixes.
Can you make sure that everything is working ok according to your expectations? Before I go ahead and append to the Android version...

AlBirdie:
Sure, just give me a few minutes. The other changes I've talked about in #92 will then be based on the new master branch. I'm going to make a separate pull request for each feature, that'll make it much easier to spot bugs and make it easier to merge.

danielgindi:
Yes that's a good idea :-)

AlBirdie:
[...]

PR #95, opened 2015-05-20
AutoScaling yAxis during panning / zooming

AlBirdie:
This adds the auto scaling we've talked about in #89 . There is an issue that I haven't been able to solve yet, though. Sometimes during panning, the chart is kind of stuck in a re-drawing cycle which. I've got no idea why this is happening, but maybe you've got an idea, so please don't merge this as it is.

With autoscaling enabled, the drawRect method runs about 10% slower for the datasets of the ChartsDemo application. In our finance chart with 250 candles, this slow down isn't noticeable on an actual device.

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -24,6 +24,7 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
     private var _pinchZoomEnabled = false
     private var _doubleTapToZoomEnabled = true
     private var _dragEnabled = true
+    private var _autoScaleMinMaxEnabled = false;
     
     private var _scaleXEnabled = true
     private var _scaleYEnabled = true
@@ -153,6 +154,12 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
         _leftYAxisRenderer?.renderAxisLine(context: context);
         _rightYAxisRenderer?.renderAxisLine(context: context);
 
+        if (_autoScaleMinMaxEnabled)
+        {
+            calcMinMax();
+            calculateOffsets();
+        }
+        
         // make sure the graph values and grid cannot be drawn outside the content-rect
         CGContextSaveGState(context);
 
@@ -256,6 +263,11 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
     
     internal override func calcMinMax()
     {
+        if (_autoScaleMinMaxEnabled)
+        {
+            _data.calcMinMax(start: lowestVisibleXIndex, end: highestVisibleXIndex);
+        }
+        
         var minLeft = _data.getYMin(.Left);
         var maxLeft = _data.getYMax(.Left);
         var minRight = _data.getYMin(.Right);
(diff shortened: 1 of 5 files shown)


danielgindi:
Cool! I'll look into it
Just note that Xcode has a bug that causes the indentation of opening braces in properties, something really weird, I always un-indent it manually

AlBirdie:
Yes, I've noticed some inconsistencies during committing the files. I hope I kept all the formatting identical to the coding style of the library here on GitHub.

dorsoft:
Nice :)

I've managed to fix the re-drawing cycle issue on the candle chart be making the following changes:
ChartData.swift (lines 139,140) change:
 _yMin = _dataSets[0].yMin;
_yMax = _dataSets[0].yMax;
to
 _yMin = _dataSets[0].yMax;
_yMax = _dataSets[0].yMin;

CandleChartDataSet.swift (line 74) change
for (var i = 0; i <= endValue; i++)
to
for (var i = start; i <= endValue; i++)

BarLineChartViewBase.swift - call calcMinMax() only upon index change
1. Declare vars:
internal var _lastLowestVisibleXIndex: Int!
internal var _lastHighestVisibleXIndex: Int!

2. zero vars in initialize()
_lastLowestVisibleXIndex = 0;
_lastHighestVisibleXIndex = 0;

3. check in drawRect() before calling calcMinMax() 
 if (_autoScaleMinMaxEnabled)
        {
            if (_lastLowestVisibleXIndex != lowestVisibleXIndex || _lastHighestVisibleXIndex != highestVisibleXIndex) {
                calcMinMax();
                calculateOffsets();
                _lastLowestVisibleXIndex = lowestVisibleXIndex;
                _lastHighestVisibleXIndex = highestVisibleXIndex;
            }
        }

4.  There is an issue calculating _rightAxis.axisMinimum in calcMinMax() (line 309)

_rightAxis.axisMinimum = !isnan(_rightAxis.customAxisMin) ? _rightAxis.customAxisMin : (minRight - bottomSpaceRight);

when
 A. _rightAxis.customAxisMin is undefined
 B. minRight == 0.25f
 C. bottomSpaceRight == 0.1f

_rightAxis.axisMinimum value jumps "sporadically" between 0.15f and -0.01f  

Because I'm using only the right axis and my y values are always > 0 I've added the following to calcMinMax (line 311) 

if (_rightAxis.axisMinimum < 0) {
            _rightAxis.axisMinimum = 0;
}

And fixed the re-drawing cycle issue for my data set

AlBirdie:
Let me try that @dorsoft . If that does the the issue for the demo data set as well (I haven't had this issue on a CandleStickChart using our DataSets either), I'll merge your changes into my fork and create a new pull request.

danielgindi:
I still haven't managed to debug and find the source for the render loop,
I'd like to try first before adding so many untested changes

dorsoft:
I believe that the root cause for the render loop issue is an axisMinimum miscalculation in  BarLineChartViewBase, calcMinMax(), line 309 (@AlBirdie pull request)

AlBirdie:
Yes, seems like that did the trick.

danielgindi:
I'm testing it at this moment, and it doesn't compile because you forgot the bubble charts...
But do not touch I'm in a merge with the bugfixes right now

AlBirdie:
Whoops, sorry, totally forgot about those.

danielgindi:
Well @dorsoft it doesn't seem to be the root cause, as @AlBirdie did not actually touch anything around line 309. Line 309 derives from `minRight` which then comes from `data.getYMin()`. So it's actually the reborn `calcMinMax` functions. 

Something there seems weird to me - they are initialized with values that are not calculated yet (ymin/ymax of dataset[0], not calculated yet) and then iterating over all datasets. It should actually initialize with FLT_MIN/FLT_MAX.

I'm trying this, along with your fixes

Issue #92, opened 2015-05-19
General discussion & news


danielgindi:
Hi everybody!

Instead of announcing stuff through README updates, or speaking to you through personal email, I want  to have this here as a place for communications.

In general, you should know that everything is done on both iOS and Android platforms so sometimes it requires Phil and I thinking about stuff together. Phil is wonderful and very cooperative so I do not feel like the communication adds any overhead. Of course sometimes we are working on different features at the same time and it takes time to review, merge etc. *(And btw, we also have personal lives! I mean I don't know about Phil, but I certainly do :-) )*

----

Just so you all know, I've had a few busy days and although I did pay some attention I did not work on any major features. In the near few days I hope to get back to implementing important features or solving bugs.

As usual, anyone wanting to participate by helping with bugs or new features - is most welcome.
If you don't want to or don't feel that you are there yet, it's also ok to buy me a beer ;-)

AlBirdie:
I've got two more additions to the library that I'd like to contribute. 
The first thing is basically just a small adjustment to what's already there;
You can configure the xAxis to be enabled or not. If false, neither labels, nor grid lines will be rendered.
If true, you'll get both labels and the grid. That's fine. Now you can also set 'drawLabelsEnabled' on the xAxis. If true you'll see the xAxis labels. If false, you don't get labels, but the offset of the chart doesn't adjust. 
In order to fix this (and therefore not waste any space by assigning space to invisible xAxis labels), a tiny adjustment has to be made;
In 'BarLineChartViewBase.calculateOffsets()' change 'if(xAxis.isEnabled)' to if(xAxis.isEnabled && xAxis.isDrawLabelsEnabled)'. That'll cause the chart area to stretch to the available height.
The same applies for the Android implementation. 

The second addition is a getter and setter for the y axis width. That way you can easily adjust the axis width to a given value if you've got multiple charts in a single view. Simple use case, a finance chart with a main chart showing the instrument and several technical studies in separate charts below the main chart. Naturally they all share the same x values and zooming / panning is synched across all charts. In order to have the x grid at the same position all the time, you need the y axis to be the same width for all charts. During initialisation of the charts, just look for the widest y axis and adjust all other chart axes accordingly.
I can make a pull request for that if you like.

Third addition would be a right text alignment for the right y axis. That basically just requires a new property on the y axis and inside 'ChartYAxisRenderer.renderAxisLabels' something around the lines of this;

if (labelPosition == .OutsideChart)
            {
[...]

PR #106, opened 2015-05-26
Offset adjustment when drawLabels on the x axis is disabled.

AlBirdie:
As I said in issue #92 the x axis offset doesn't adjust when you disable x axis labels. With this change you can have the x grid without labels but the chart area still fills the entire viewport.


AlBirdie:
Perfect. Thanks Daniel!

PR #107, opened 2015-05-26
Added support for setting a custom width that is wider than the longe…

AlBirdie:
This adds support to set a custom width to an y axis as described in #92 .
Note that this only works with values that are greater than the longest label width. 
It takes the max from the longest label and the custom width.

What do we need this for? If you've got several charts that share the same x values vertically stacked in a view, setting a common y axis width based on the widest y axis not only looks nicer, it also allows you to get away with a single x axis (that is, in case the chart ranges are kept in sync of course).

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -1514,6 +1514,33 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
     /// returns true if autoScaleMinMax is enabled, false if no
     /// :default: false
     public var isAutoScaleMinMaxEnabled : Bool { return autoScaleMinMaxEnabled; }
+    
+    
+    /// Sets a custom width to the specified y axis.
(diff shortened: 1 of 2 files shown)


AlBirdie:
bildschirmfoto 2015-05-26 um 19 13 47

Like this, basically.

danielgindi:
I'm improving it to have a max and a min width, and then merging it in.
By the way, you should really make branches for your PRs, so you can have separate PRs. (Create a branch, make whatever changes, create PR. Then go back and branch from the same spot and make another)

AlBirdie:
Ok, sounds even better.
Regarding the branching, do you have any difficulties with merging these pull requests the way I do them now? I usually pull from your master branch before implementing my changes and create the pull request from that, so (ideally) there shouldn't be any conflicts when you merge them, at least I hope so.
However, if branching makes it easier for you, then I'll do that of course.

danielgindi:
No it is meant to make it easier for you, so you are not "stuck" waiting for a PR to go through...

PR #149, opened 2015-06-12
Swift 2.0

petester42:
Here is some initial work on converting the project to swift 2.0. Everything should still work as before. There are a couple of places that variable were created but never used so I didn't know what to do for some of those since they call a delegate method. Rather than breaking things I just replaced them with `_` like Xcode suggested. That is something that should be looked into and remove all unused code to make the codebase cleaner.

### DIFF
--- a/Charts/Classes/Charts/BarChartView.swift
+++ b/Charts/Classes/Charts/BarChartView.swift
@@ -48,7 +48,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
             return
         }
         
-        var barData = _data as! BarChartData
+        let barData = _data as! BarChartData
         
         // increase deltax by 1 because the bars have a width of 1
         _deltaX += 0.5
@@ -60,7 +60,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
         
         for (var i = 0, count = barData.dataSetCount; i < count; i++)
         {
-            var set = barData.getDataSetByIndex(i)
+            let set = barData.getDataSetByIndex(i)
             
             if (maxEntry < set!.entryCount)
             {
@@ -68,7 +68,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
             }
         }
         
-        var groupSpace = barData.groupSpace
+        let groupSpace = barData.groupSpace
         _deltaX += CGFloat(maxEntry) * groupSpace
         _chartXMax = Double(_deltaX) - _chartXMin
     }
@@ -78,7 +78,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
     {
         if (_dataNotSet || _data === nil)
         {
-            println("Can't select by touch. No data set.")
+            print("Can't select by touch. No data set.", appendNewline: true)
             return nil
         }
         
@@ -93,17 +93,17 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
     }
     
     /// Returns the correct Highlight object (including xIndex and dataSet-index) for the specified touch position.
-    internal func getHighlight(#xPosition: CGFloat, yPosition: CGFloat) -> ChartHighlight!
+    internal func getHighlight(xPosition xPosition: CGFloat, yPosition: CGFloat) -> ChartHighlight!
     {
         if (_dataNotSet || _data === nil)
         {
             return nil
         }
         
-        var barData = _data as! BarChartData!
+        let barData = _data as! BarChartData!
         
-        var setCount = barData.dataSetCount
-        var valCount = barData.xValCount
+        let setCount = barData.dataSetCount
+        let valCount = barData.xValCount
         var dataSetIndex = 0
         var xIndex = 0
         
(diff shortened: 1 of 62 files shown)


Charts/Classes/Charts/BarChartView.swift
```diff
@@ -93,17 +93,17 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
     }
     
     /// Returns the correct Highlight object (including xIndex and dataSet-index) for the specified touch position.
-    internal func getHighlight(#xPosition: CGFloat, yPosition: CGFloat) -> ChartHighlight!
+    internal func getHighlight(xPosition xPosition: CGFloat, yPosition: CGFloat) -> ChartHighlight!
```
MGaetan89: I don't know Swift, but is it normal to have `xPosition ` twice here?

liuxuan30:
In swift method, the second and later paramaters by default will have the same name for internal and external parameters, only the first on is exception. adding a '#' means to force set the external name same as internal name for the first parameter. So you don't have to add another xPosition again.

petester42:
I believe the swift 2.0 converter changed the `#` to the duplicated `xPosition` because I didn't change that.

liuxuan30:
perhaps, because it is said using ‘#’ is not recommended.

danielgindi:
Apple removed the # from the specification, they said it wasn't clear
enough. They prefer duplicate code...

petester42:
I agree with the decision. It makes more sense to have to specify internal and external parameters. Makes you chose the interface you want to expose better.

danielgindi:
I don't see why you've invested so much in removing semicolons........... It just messes up history by changing every single line of code. Semicolons are not forbidden in Swift 2.0, you can choose whether to use them or not :-)

petester42:
It really didn't take any time with the handy replace all feature. I get that semicolons are not forbidden but they serve no purpose how they are currently used. In my opinion swift code should not have semicolons unless you are doing two operations on one line which follows Apple's own code style.

I really don't agree that it messes up the history. The point of the history is to track changes not to look pretty. If the focus was keeping the history clean then no one would ever refactor or make any changes.

I think matching Apple's code style is a good thing and makes the code cleaner. If everyone disagrees with this then I'll gladly revert the change.

danielgindi:
Apple's code style is brackets up, which is counter productive, so... not
everything Apple is sacred ;-)

I do not object to removing the brackets, but these are not part of moving
to Swift 2.0, and it makes it harder to see and review the actual changes

petester42:
Ah ok. I see your point now. Makes sense.

petester42:
I was just looking at this again and I think that in the move to swift 2.0 the code base should be significantly updated to be more swift. There are a lot of explicitly unwrapped optionals and nil checks instead of `if let`. Moving to a more swift syntax would reduce the unnecessary checks and it would make the codebase safer. What does everyone think about that?

danielgindi:
I think it's a good idea.
Think about the process of migrating from Android to Swift:
1. Copy all .java files to .swift
2. Fix basic syntax problems (Most syntax match as Swift is flexible to accept whatever syntax you like. There's no specific Swift syntax)
3. Re-write drawing code for Core Graphics
4. Write animation classes
5. Write touches and gesture handlers
6. Write the demo app
7. Testing

What happens in this case is that code is very similar to its Java counterpart, also allowing for easy comparison to check for errors in translation.

At this stage, the two are identical. And it's a piece of cake to take a commit or two from the Android repo and implement in the iOS version, so it's not a problem to refactor the Swift code.
I actually do more code translating new iOS features into the Android repo :-) It's not a mirror anymore, it's a self evolving project, since after 3 weeks from when it was borne. And that is thanks to all of the people involved here!

There are a few fixes/improvements that we are planning to do right before the soon to come version release. We want that release to be stable of course, as there are people using CocoaPods or Gradle to fetch a specific version.

Right after that, I want to do some refactoring for the Swift code. @petester42 I guess you can come in on that :-)
Based on that - a branch for Swift 2.0. As Swift 2.0 isn't released yet.

I'm still not sure about making the Swift refactoring part of the next release. But I tend to think that it should be *more* stable than the current version, by moving to optionals where necessary. And most refactoring is just removing extraneous syntax. So there's a chance that it will make it to the next release...

danielgindi:
BTW this PR is already too far from the repo to be automatically merged... If it were rebased to just include the Swift 2.0 transition it would be easily integrated to an updated branch...

petester42:
The thing is that this branch does only include the swift 2.0 changes. I basically ran the converter and updated the documentation to the new syntax.

I think that either refactoring before or during the 2.0 transition would be fine. Really depends on when the next release is planned. 2.0 does include certain simplifications that would make the refactor a little cleaner but overall it shouldn't matter. The 2.0 changes could just be applied after the refactor in that case.

danielgindi:
@petester42 could you rebase your Swift 2.0 changes on `master`?

danielgindi:
@petester42 actually the doc updates as a separate PR would also be nice. This could easily be achieved through rebasing or cherry-picking :-)

danielgindi:
I would have done this myself but I'd hate to steal your credit... Cherry-picking will set my name beside the commit.

danielgindi:
Or do you mean that Xcode 7 did all of it? Including converting the docs?

petester42:
Xcode did all the code by itself. The documentation I needed to update myself.

danielgindi:
Ok there's now a swift-2.0 branch, including your changes to the comments which I successfully pulled from your branch. It's updated to beta 2 already.

danielgindi:
I'm closing this, as this is on the swift-2.0 branch already. @petester42 thanks for you help!

Issue #122, opened 2015-06-01
Candle chart - make the shadow same color as an increasing/decreasing candle color 


dorsoft:
I'm developing a financial app, it is customary to color the candle shadow the same color as the candle itself. e.g. green/red for increasing/decreasing.

I've managed to accomplish that by adding the following code:
A. CandleChartDataSet line 31
/// use candle color for the shadow
    public var makeShadowSameColorAsCandle = false

B. CandleStickChartRenderer - drawDataSet (line 104) changed from:

CGContextSetStrokeColorWithColor(context, (dataSet.shadowColor ?? dataSet.colorAt(j)).CGColor);

to

           var shadowColor = dataSet.shadowColor ?? dataSet.colorAt(j);
            if (dataSet.makeShadowSameColorAsCandle)
            {
                if (e.open >= e.close)
                {
                    shadowColor = dataSet.decreasingColor ?? dataSet.colorAt(j);
                }
                else
                {
                    shadowColor = dataSet.increasingColor ?? dataSet.colorAt(j);
                }
            }
            
            CGContextSetStrokeColorWithColor(context, shadowColor.CGColor);    

Could you please add it to the master ?

AlBirdie:
I've had the same requirement and altered the CandleStickChartRenderer almost identically to your solution. However, what is missing is the case for e.open == e.close, so a neutral shadow color.

dorsoft:
Nice catch @AlBirdie, I guess we could use something like:

var shadowColor = dataSet.shadowColor ?? dataSet.colorAt(j);
        if (dataSet.makeShadowSameColorAsCandle)
        {
            if (e.open > e.close)
            {
                shadowColor = dataSet.decreasingColor ?? dataSet.colorAt(j);
            }
            else if (e.open < e.close) 
            {
                shadowColor = dataSet.increasingColor ?? dataSet.colorAt(j);
            }
        }
      CGContextSetStrokeColorWithColor(context, shadowColor.CGColor);

AlBirdie:
Yes, that's it. :)

dorsoft:
On top of that it is customary to draw a horizontal line when the candle height is zero, I've added this feature as well

PR #150, opened 2015-06-14
Candle chart - make the shadow same color as an increasing/decreasing candle color #122

dorsoft:
I'm developing a financial app, it is customary to color the candle shadow the same color as the candle itself. e.g. green/red for increasing/decreasing.

On top of that it is customary to draw a horizontal line when the candle height is zero, I've added this feature as well

### DIFF
--- a/Charts/Classes/Data/CandleChartDataSet.swift
+++ b/Charts/Classes/Data/CandleChartDataSet.swift
@@ -28,6 +28,9 @@ public class CandleChartDataSet: BarLineScatterCandleChartDataSet
     /// the color of the shadow line
     public var shadowColor: UIColor?
     
+    /// use candle color for the shadow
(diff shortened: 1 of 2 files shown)


danielgindi:
This seems incomplete. As you are missing the other case.
Also what about just letting the user choose the colors for shadows?...

dorsoft:
What other case ?

My implementation works like this:
1. makeShadowSameColorAsCandle = false
    Works exactly like your current implementation, the shadow color is determined by CandleChartDataSet.shadowColor 

2. makeShadowSameColorAsCandle = true
    The shadow color is determined by the candle color

3. If a candle height is zero (open==close), draw a horizontal line using CandleChartDataSet.shadowColor

danielgindi:
You have this in another PR. You should move them to separate branches.

PR #152, opened 2015-06-15
add NaN check to allow non-digits handling for radar chart

liuxuan30:
The background is that not every data in every dataSet is valid number in real life.

add NaN check to allow non-digits handleing. if a CGPoint contains NaN, simply ignore it, this will ignore current point and continue the path with next valid point.

This is helpful, where dataSet 1 contains invalid data but dataSet 2 has valid data. People can have a chance to insert NaN if user found the data is nil/null/NSNotFound/ifinity/DOUBLE_MAX.... any number that cannot be drawn on screen or even exists in real life, and still get the chart.

for example, 

xValues: [1,2,3,4,5]
dataSet 1: [5,6,NaN,8,9]
dataSet 2: [11,12,13,14,15]

With NaN support, the radar chart will connect dataSet 1 values: 5-6-8-9 as a web, and dataSet 2 as 11-12-13-14-15 as another web.

**Attention:**
I only add the code for radar chart, since I am more familiar with it and gives the demo. If it is a valid scenario, you can expand it to other charts accordingly.

### DIFF
--- a/Charts/Classes/Charts/PieRadarChartViewBase.swift
+++ b/Charts/Classes/Charts/PieRadarChartViewBase.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 import CoreGraphics
-import UIKit.UIGestureRecognizer
+import UIKit
 
 /// Base class of PieChartView and RadarChartView.
 public class PieRadarChartViewBase: ChartViewBase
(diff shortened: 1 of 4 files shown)


liuxuan30:
radarnan

Check out the little white dot with yellow, its corresponding blue line has NaN so we don't have to highlight it and draw it.

PR #165, opened 2015-06-19
Fix a silly bug. should check if first is -0.0

liuxuan30:
My bad, made a terrible mistake. the purpose is to check if first is -0.0, turn it into 0.0. I forgot to copy it from my project...


danielgindi:
:-)
And I was there scratching my head...

liuxuan30:
FYI Please take a look at issue #166. I am trying to fix it, but it causes some side effect. still looking for a perfect solution

Issue #166, opened 2015-06-19
Radar chart + Bar Line chart Base + startAtZeroEnabled = false + negative values


liuxuan30:
When turning off startAtZeroEnabled, the negative values are not proper drawn. the radius ratio for each inner web is not correct.

The main trouble is:
```swift
        var labelCount = _chart.yAxis.entryCount;
        
        for (var j = 0; j < labelCount; j++)
        {
            for (var i = 0, xValCount = _chart.data!.xValCount; i < xValCount; i++)
            {
                var r = CGFloat(_chart.yAxis.entries[j] - _chart.chartYMin) * factor;
...
```
Take an example if _chart.yAxis.entries is: [-100000.0, 0.0, 100000.0, 200000.0, 300000.0, 400000.0]

and _chart.chartYMin is -35991.2689931.

The minimum raw data is -7501.57 and max is 277395.38. 
![negabug](https://cloud.githubusercontent.com/assets/4375169/8249438/8fcd97c6-169c-11e5-98b9-d882ef37e943.png)

liuxuan30:
Provide data here:
data set 0:
[ChartDataEntry, xIndex: 0, value -7501.574034, ChartDataEntry, xIndex: 1, value 73587.954072, ChartDataEntry, xIndex: 2, value 68861.089668, ChartDataEntry, xIndex: 3, value 79499.134258, ChartDataEntry, xIndex: 4, value 53615.304423]

data set 1:
[ChartDataEntry, xIndex: 0, value 35790.059533, ChartDataEntry, xIndex: 1, value 20118.91748, ChartDataEntry, xIndex: 2, value 42567.845168, ChartDataEntry, xIndex: 3, value 38749.916356, ChartDataEntry, xIndex: 4, value 13551.626137]

data set 2:
[ChartDataEntry, xIndex: 0, value 163400.659256, ChartDataEntry, xIndex: 1, value 187068.443587, ChartDataEntry, xIndex: 2, value 277395.375557, ChartDataEntry, xIndex: 3, value 133373.871463, ChartDataEntry, xIndex: 4, value 132363.776504]

xVals could be any string.

danielgindi:
Well this is not very clear to me. Can you show a visual example of how negative values should behave in a Radar chart vs. how it is now?

liuxuan30:
[...]

PR #168, opened 2015-06-19
fix radar chart negative value rendering bug if startAtZeroEnabled is false for issue #166

liuxuan30:
For issue #166 
we have to check if Double(yMin) / interval will be  -1<x<0, if this is the case, ceil(x) would be 0. but I think we need it to be -1.

In order to render the negative values correctly, 
```swift
 _yAxis.axisMinimum = _yAxis.entries[0];
```
is the key. 

I have been thinking about different ways to fix. But I was suddenly realized the whole reason is the axisMinimum is not changed while the entries have been modified. Add this line solved my problem, and other radar charts remain good on my side. So I think this should be the correct way to fix it.

@danielgindi you must double check here, because I am not sure if you missed to set it, or you intend not to set _yAxis.axisMinimum.


danielgindi:
I still need to understand the different situations to see what you did here... Could you supply samples - before/after?

liuxuan30:
This is mainly about issue #166. Current logic for negative values has defect, especially for radar chart. Go to #166 to see all the details and the screenshot before and after.

You can try on your side, just turn off startAtZeroEnabled, and feed a nagative value. We need to use floor instead of ceil for negative values.

For radar chart, `_yAxis.axisMinimum = _yAxis.entries[0];` is missing for correctly rendering negative values;

For commit: `fix drawing out of bound negative value axis labels issue`:
If the axis contains negative values, it will draw out of the bound; I have a screenshot in issue #166 as well.

For commit: `drawExtras should also be included before restoring clipped context`:
I think drawExtras should be included in the clipped context; otherwise it will render out of the chart.

liuxuan30:
I will close this since this mixed up some different issue here, though it is quite slight. I will use different branch for this issue.

liuxuan30:
OK, split done. see pull request #207, #208, #209

PR #178, opened 2015-06-28
fixes #177

### LINKED ISSUE #177: Render Crash Due to Rect NAN from rectValueToPixel in BarChartRenderer.swift
I'd offer a fix for this but this is my very first look at ios-charts and I've only got about 20 minutes in on this project. With that said, the bug is exposed in the ChartsDemo using the Sinus Bar Chart but reducing the element count to 1 by dragging the control at the bottom of the chart as far to the left as possible.

BarChartRenderer.drawDataSet attempts to pixelize the rect to be drawn on line 124 by calling trans.rectValueToPixel(&barRect). When there is only one element to be drawn, barRect is (-0.3, 0.0, 0.6, 0.0) before the call and (nan, nan, nan, nan) after the call. Apparently, the affine transform performed in rectValueToPixel does not like something about the incoming rect. 

None of the other charts allow less than two elements to be presented so, without some additional futzing, I was unable to see if this bug is manifest for single-element bar charts.

(If I get time, I'll research further.)

petester42:
Fixes issue #177. I assumed that NaN would be resolved to 0.

### DIFF
--- a/Charts/Classes/Utils/ChartViewPortHandler.swift
+++ b/Charts/Classes/Utils/ChartViewPortHandler.swift
@@ -332,22 +332,34 @@ public class ChartViewPortHandler: NSObject
     
     public func isInBoundsLeft(x: CGFloat) -> Bool
     {
+        if isnan(x) {
+            return _contentRect.origin.x <= 0.0 ? true : false;
+        }
         return _contentRect.origin.x <= x ? true : false;
     }
     
     public func isInBoundsRight(x: CGFloat) -> Bool
     {
+        if isnan(x) {
+            return (_contentRect.origin.x + _contentRect.size.width) >= 0.0 ? true : false;
+        }
         let normalizedX = CGFloat(Int(x * 100.0)) / 100.0;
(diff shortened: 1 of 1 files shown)


danielgindi:
Did you test when is that value coming in as `NaN`? 
On big data sets we would prefer not to do extra checks in the bounds checking... So if we can prevent that from happening at the source it would be much better

petester42:
Yes. The value becomes NaN after this [transformation is applied](https://github.com/danielgindi/ios-charts/blob/a00453fadb0cdf2ba7076b7ddca17fe990975833/Charts/Classes/Renderers/BarChartRenderer.swift#L124) when the values are small. Since it affects small values I assumed that 0 would work. Either there is a check after every transform or there is a check in the bounds. I don't see a way around this. In both cases there is extra work that will need to be done in the loop.

danielgindi:
Well, I found it... It comes from the recent improvements to min/max calculations. Basically, one needs to remember that DBL_MIN is not the minimum value, it's the minimum positive value... It's a trap everybody falls into.

petester42:
:+1:

PR #179, opened 2015-06-29
The line charts have started to properly display balloon markers

Maxim-38RUS-Zabelin:
I would like to display several sequential data sets on the same line chart (two data sets are sequential if the last element of the 1st set is also the first element of the 2nd one e.g. [1,2,3] and [3,4,5]); unfortunately, set up so, the line chart stops properly display balloon markers (they appear only over some points of the plot). The attached commit fixes the described issue.


danielgindi:
Thanks!

danielgindi:
This seems to be caused because of a deeper reason, from one of the latest changes.. I'll address it immediately

Maxim-38RUS-Zabelin:
You're welcome!

Issue #29, opened 2015-04-14
Performance Enhancements


AlBirdie:
Are there any plans to enhance the performance of the library in a way Philip did with MPAndroidCharts?
With his enhancements it is now possible to render thousands of data points smoothly on Android. 
I've had a quick glance at the ios-charts implementation and from what I've seen it is based on the original MPAndroidCharts implementation without the latest performance enhancements.

Having it render thousands of points as well would be HUGE plus for the library as it would effectively render all commercial versions useless (most of which are a waste of time and money anyways).

danielgindi:
The original performance bottleneck in the Android version was many extra allocations (of arrays) inside the rendering code. Memory **should not be allocated** inside rendering code. So @PhilJay has moved that code into a Buffer that is preallocated, and delegating the rendering calculations to the Buffer classes while rendering.

I chose to not move the rendering calculations to the Buffer classes, but just preallocate the required memory the same way - but have the rendering calculations in the same loop as the rendering code. This way it is far easier to manage the code.

Also it IS a performance gain that the rendering calculations are not being done inside extra functions, and the data is looped only once inside the rendering code instead of twice.

Btw, in Swift performance is FAR FAR better than Java anyway, so you would have seen that performance gain even without pre-allocating those arrays.

In order to graphically explain it, this is what is/was being done in the rendering code:

* [BEGIN RENDERING]
*   Loop over DataSets
*   -- Allocate array for points of this DataSet -- **Now preallocated in both platforms**
*   -- Calculate points for rendering -- *Moved to Buffers in Android version*
*   -- Render those points
* [COMMIT RENDERING]

Also note that in Java, abstractions (functions, classes, inheritance) come with a great price. It is not for nothing the Google themselves wrote an [important recommendation](http://developer.android.com/training/articles/memory.html#Abstractions):

    Be careful with code abstractions

    Often, developers use abstractions simply as a "good programming practice," 
    because abstractions can improve code flexibility and maintenance. 
    However, abstractions come at a significant cost: 
        generally they require a fair amount more code that needs to be executed, 
    requiring more time and more RAM for that code to be mapped into memory. 
    So if your abstractions aren't supplying a significant benefit, you should avoid them.


Summing it up: 
* You can be comfortable with working with thousands of data points!
* I'm trying to convince Phil to move point calculations back to rendering loops to keep it simpler to maintain ;-)

AlBirdie:
[...]

PR #192, opened 2015-07-08
Performance Enhancements #29 - candle chart

dorsoft:
performance enhancement for the candle chart.

### DIFF
--- a/Charts/Classes/Data/CandleChartDataSet.swift
+++ b/Charts/Classes/Data/CandleChartDataSet.swift
@@ -62,7 +62,7 @@ public class CandleChartDataSet: BarLineScatterCandleChartDataSet
         }
         else
         {
-            endValue = end
+            endValue = end > (entries.count - 1) ? (entries.count - 1) : end
         }
         
         _lastStart = start
--- a/Charts/Classes/Renderers/CandleStickChartRenderer.swift
+++ b/Charts/Classes/Renderers/CandleStickChartRenderer.swift
@@ -67,11 +67,8 @@ public class CandleStickChartRenderer: ChartDataRendererBase
         
         var entries = dataSet.yVals as! [CandleChartDataEntry]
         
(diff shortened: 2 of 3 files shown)


danielgindi:
How exactly does this increase performance?
This actually makes the renderer loop over *more* labels than needed.
It also made scrolling jerkier when having 2.5K points

dorsoft:
1. I'd forgot to make the same change in drawDataSet(), I've pushed the fix to the PR. please pull and run the candle demo with 2500 candles.

2. dataSet.entryIndex becomes slower as the chart is scrolled from left to right for datasets > 300.
    I've added time measurements logs and ran the candle demo with 2500 candles (iPhone 5) and scrolled the chart all the way to the right. I've found out that the  4 lines of code with dataSet.entryForXIndex and  dataSet.entryIndex take about 70-90ms to complete and are called about 5-7 times a second i.e. (5 X 2 X 70ms) to (7 X 2 X 90ms) = 700ms to 1260ms per second are wasted.   
  
2. The render loops over the same number of labels and candles as:
var entryFrom = dataSet.entryForXIndex(_minX)
var entryTo = dataSet.entryForXIndex(_maxX) 
var minx = max(dataSet.entryIndex(entry: entryFrom!, isEqual: true), 0)
var maxx = min(dataSet.entryIndex(entry: entryTo!, isEqual: true) + 1, entries.count)

is always equal to 

var minx = max(_minX, 0)
var maxx = min(_maxX + 1, entries.count)

You can test this easily by adding some testing code like:

var minx2 = max(_minX, 0)
var maxx2 = min(_maxX + 1, entries.count)
if (maxx != maxx2 || minx != minx2) {
println("maxx2 or minx2 are different ! ! !")
}

danielgindi:
This introduces only a very minor improvement, even for 2500. You have probably measured in an un-optimized compilation. Anyway, I've tested and did not find any side effects.

PR #200, opened 2015-07-09
Fixes for beta 3

petester42:
This fixes some compiler errors in beta 3. Oops. Deleted my remote while trying to fix the history.


petester42:
As far as the `@obj` bridging is concerned you can't remove them in beta-3

PR #208, opened 2015-07-13
renderExtras is inside the chart contentRect, so renderExtras should be clipped

liuxuan30:
When I tried to add something more in renderExtras, it will draw outside of the chart contentRect. I think renderExtras should also be clipped. How do you think?

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -216,11 +216,11 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
         {
             renderer?.drawHighlighted(context: context, indices: _indicesToHightlight)
         }
-
-        // Removes clipping rectangle
-        CGContextRestoreGState(context)
         
         renderer!.drawExtras(context: context)
+        
+        // Removes clipping rectangle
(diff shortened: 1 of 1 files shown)


liuxuan30:
@danielgindi today I kind of know why you don't clip `drawExtras`, because you are drawing dots in the line chart view, and you don't want the dot to be clipped while the dot is at the edge.

However, when I tried to add some additonal stuff on the chart, I make use of drawExtras and see it draws beyond the rect, so I think it is reasonable to be clipped.

Maybe we should reconsider the structure here, as sometimes people want drawExtras to be clipped while sometimes not

danielgindi:
Actually you should make sure you don't draw something that is out of bounds...
You *could* clip yourself when you override drawExtras, but the best practice is to just *not draw* anything that shouldn't be visible

liuxuan30:
@danielgindi but you draw the line chart dots in drawExtras, and it is drawing out of bounds :-) 
though it will not exceeds the axis

danielgindi:
Yes, specifically those dots look bad when clipped, so we are drawing them even if they are bleeding a little bit. But eventually when you subclass it is up to you if you clip or not :-)

Take into consideration that the fact that you are clipping, in itself it requires more resources for drawing. It comes with a cost.

PR #209, opened 2015-07-13
when the point of label out of viewPortHandler.contentRect, we should not draw the label

liuxuan30:
when the point of label out of viewPortHandler.contentRect, we should not draw the label

1.for bar chart, if the label point is out of contentRect vertically, ignore it and continue
2.for horizontal bar chart, if the label point is out of contentRect horizontally, ignore it and continue


liuxuan30:
Add a horizontal bar chart example to show th reason. there is a -100 label drawn out bound of y axis range.

liuxuan30:
Hold on, today I met a bug introduced by my commit, the pt.y is actually less than contentRect.origin.y, and it should be drawn.

Wondering why. Still pending investigation.

Issue #214, opened 2015-07-15
[BUG] Bar Chart incorrect rendering for special data


liuxuan30:
Here's what I changed in `- (void)setDataCount:(int)count range:(double)range` for the demo multiple bar chart view controller, other remains the same:

```Objective-C
- (void)setDataCount:(int)count range:(double)range
{
    NSMutableArray *xVals = [[NSMutableArray alloc] init];
    for (int i = 0; i < 11; i++)
    {
        [xVals addObject:[@(i + 2015) stringValue]];
    }
    NSMutableArray *yVals1 = [[NSMutableArray alloc] init];
    NSMutableArray *yVals2 = [[NSMutableArray alloc] init];

    for (int i = 0; i < 11; i++)
    {
        if (i == 0) {
            [yVals1 addObject:[[BarChartDataEntry alloc] initWithValue:2354235 xIndex:i]];
            [yVals2 addObject:[[BarChartDataEntry alloc] initWithValue:3354235 xIndex:i]];
        } else if (i == 10) {
            [yVals1 addObject:[[BarChartDataEntry alloc] initWithValue:6354235 xIndex:i]];
            [yVals2 addObject:[[BarChartDataEntry alloc] initWithValue:7354235 xIndex:i]];
        } else {
            continue;
        }
    }
    
    BarChartDataSet *set1 = [[BarChartDataSet alloc] initWithYVals:yVals1 label:@"Company A"];
    [set1 setColor:[UIColor colorWithRed:104/255.f green:241/255.f blue:175/255.f alpha:1.f]];
    BarChartDataSet *set2 = [[BarChartDataSet alloc] initWithYVals:yVals2 label:@"Company B"];
    [set2 setColor:[UIColor colorWithRed:164/255.f green:228/255.f blue:251/255.f alpha:1.f]];
    
    NSMutableArray *dataSets = [[NSMutableArray alloc] init];
    [dataSets addObject:set1];
    [dataSets addObject:set2];
    
    BarChartData *data = [[BarChartData alloc] initWithXVals:xVals dataSets:dataSets];
    data.groupSpace = 0.8;
    [data setValueFont:[UIFont fontWithName:@"HelveticaNeue-Light" size:10.f]];
    _chartView.leftAxis.startAtZeroEnabled = NO;
    _chartView.rightAxis.startAtZeroEnabled = NO;
    _chartView.data = data;
}
```
[...]

PR #221, opened 2015-07-16
Try to fix wrong position bug for BarChartView + HorizontalBarChartView

liuxuan30:
try to fix wrong position bug for issue #214 

The idea is, because not every xIndex will have a dataEntry, if the data is invalid (nil/nan/null), the entryCount will be less than xValsCount. 

For example, xVals: [0,1,2,3,4,5,6,7,8,9,10], if there are two dataSets, like below:
xIndex:0 value:100, xIndex:10 value:200
xIndex:0 value:1000, xIndex:10 value:2000

Current logic will use entryCount instead of xValsCount to calculate the matrix and position, which will lead to shorter xPosition value on xAxis. issue #214 has the screenshot.

What I changed is I always use xValsCount and actual xIndex value to calculate.

@danielgindi There might be something I missed if I am not correct. On the other hand, if you think I am right, I would also miss places to apply the new logic, because while I am testing, I found I missed to change `getMarkerPosition`. There could be more.

I also feel it might has such issues for all bar-like charts, but I don't have time to dig out more charts. If I ran into this, I will continue trying to fix.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+# How to contribute
+
+Bug-fixes and features often come from users of the Charts framework, and improving it greatly. We want to keep it as easy as possible to contribute changes that improve the experience for users all around the world. There are a few guidelines that we
+need contributors to follow so that we can have a chance of keeping on
+top of things.
+
+## Simple issues and bug reports
+
+If you are reporting a bug which can be observed visually, please add to your issue either:
+
+* Screenshots, if the bug is easily explainable
+* A working sample project that we can compile, run, and immediately observe the issue
+
+## Getting Started with Contributions
+
+* Make sure you have a [GitHub account](https://github.com/signup/free)
+* Submit a ticket for your issue, assuming one does not already exist.
+  * Clearly describe the issue including steps to reproduce when it is a bug.
+  * Make sure you fill in the earliest version (or commit number) that you know has the issue.
+* Fork the repository on GitHub
+
+## Making Changes
+
+* Create a topic branch from where you want to base your work. This is usually the master branch.
+* Make commits of logical units.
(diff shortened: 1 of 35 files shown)


liuxuan30:
@danielgindi as you have changed lots since 2.1.1, this PR cannot be merged with green light. I won't be able to merge newer code to this branch until I got some time to resolve the conflicts... Also need you to review the whole logic. I might miss something somewhere.

liuxuan30:
alright, lucky night I got time to deal with the conflicts.. hope you don't mind a few more commits introduced.

danielgindi:
Do you know how to squash commits?

danielgindi:
If you are using SourceTree, you could right click the latest commit, Rebase, and let it rebase the code on the latest commit.
In order to squash, you need to right click the commit before your first commit, and Rebase Interativelty. There you have a squash option :-)

liuxuan30:
actually I don't know how to squash... What's SourceTree here?

liuxuan30:
Are you saying this?https://www.sourcetreeapp.com

I can take a look at it tomorrow. Need to handle some other stuff now.

liuxuan30:
@danielgindi I managed to squash the commits #247 on a new branch, though it did only have one commit that includes all your commits and mine, when trying to create a PR #247, it still shows red light:  Can’t automatically merge. And showing 35 files changed.

I am no expert of git, but to keeps it simple, I think I would stay unsquashed for this PR :( at least it's clear what I have changed.

liuxuan30:
Alright, thanks for the lesson! I managed to squash the commits.
The new PR is #248. I will close this.

ghizoulo:
I am using v3.0.1 but still have the same problem. the bars are misplaced. Please if u can help me with that check #3154

PR #225, opened 2015-07-16
Fix for calling PieChartData(xVals: nil, dataSet: dataSet) giving the…

toriaezunama:
… following warning:

"One or more of the DataSet Entry arrays are longer than the x-values array of this Data object."

Cause is that when xVals is used, if it's nil then it is set to an empty array. Later on when it is used, the warning is emitted because `dataSets[i].yVals.count > _xVals.count`

### DIFF
--- a/Charts/Classes/Data/ChartData.swift
+++ b/Charts/Classes/Data/ChartData.swift
@@ -127,7 +127,7 @@ public class ChartData: NSObject
         
         for (var i = 0; i < dataSets.count; i++)
         {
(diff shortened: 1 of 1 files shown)


danielgindi:
The warning is good; If you don't have x-indexes, the chart will not calculate positions correctly. That's a "limitation" currently, that you must specify the x-indexes, but we hope it will be lifted soon.
Anyway, your fix will just break other things :-)

PR #229, opened 2015-07-20
Fix "no data" label (with custom text).

willingheart:
If user set a custom text for «no data» label, it shows twice (default and custom).
This pull request fixes it.


danielgindi:
These are two separate properties, independent of each other. What you did does not fix anything, but just makes one of them redundant.
If you do not want one of them to draw, set it to an empty string

Issue #242, opened 2015-07-24
Grouped columns misplaced


JanciP:
Hey, I'm having troubles with grouped bar chart.
My barChartData are following:

x-values (array of dates):
02.júla.15,
23.júla.15,
24.júla.15

dataSets:
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 0, value 0.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 0, value 0.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 0, value 0.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 0, value 0.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 0, value 0.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 1, value 39.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 2, value 16.0,
ChartDataSet, label: tst, 1 entries:
ChartDataEntry, xIndex: 2, value 28.0

This produces almost correct result, however bars seems to have wrong position as showed in this image http://imgur.com/Vq43QFH

What am I doing wrong? :)

JanciP:
I've found out that adding another BarChartData for different x-cooridnate creates this Data for other indexes as well... that's why everything is moved by some offset, but I have not idea how fix it.

Data:
 BarChartDataEntry *entry = [[BarChartDataEntry alloc] initWithValues:@[@1] xIndex:0];
 BarChartDataEntry *entry2 = [[BarChartDataEntry alloc] initWithValues:@[@4] xIndex:0];    
    
 BarChartDataSet *set = [[BarChartDataSet alloc] initWithYVals:@[entry] label:@"tst"];
 BarChartDataSet *set2 = [[BarChartDataSet alloc] initWithYVals:@[entry2] label:@"tst"];

 BarChartDataEntry *entry4 = [[BarChartDataEntry alloc] initWithValues:@[@4] xIndex:2];
 BarChartDataSet *set4 = [[BarChartDataSet alloc] initWithYVals:@[entry4] label:@"tst"];
 
 NSArray *tstArr = @[set,set2, set4];
    
BarChartData *data = [[BarChartData alloc] initWithXVals:xVals dataSets:tstArr];
self.chart.data = data;

[...]

PR #248, opened 2015-07-28
try to fix bar chart + Horizontal Bar chart wrong render + highlight position bug for issue #214 and #242.

liuxuan30:
Alright, I managed to squash commits and merge v2.1.2 and #221. I will paste old message from #221 below.

Try to fix wrong position bug. use entry.xIndex instead of j of dataSet.entryCount to calculate the offset and position. Issue #214 + #242

The idea is, because not every xIndex will have a dataEntry, if the data is invalid (nil/nan/null), the entryCount will be less than xValsCount.

For example, xVals: [0,1,2,3,4,5,6,7,8,9,10], if there are two dataSets, like below:
xIndex:0 value:100, xIndex:10 value:200
xIndex:0 value:1000, xIndex:10 value:2000

Current logic will use entryCount instead of xValsCount to calculate the matrix and position, which will lead to shorter xPosition value on xAxis. issue #214 has the screenshot.

What I changed is I always use xValsCount and actual xIndex value to calculate.

@danielgindi There might be something I missed if I am not correct. On the other hand, if you think I am right, I would also miss places to apply the new logic, because while I am testing, I found I missed to change getMarkerPosition. There could be more.

merge from upstream master (+3 squashed commits)

Squashed commits:
[c861fea] add fix for horizontal bar chart
[2f75778] apply the same logic for getMarkerPosition
[baca6dd] try to fix wrong position bug. use entry.xIndex instead of j of dataSet.entryCount to calculate the offset and position

### DIFF
--- a/Charts/Classes/Charts/BarChartView.swift
+++ b/Charts/Classes/Charts/BarChartView.swift
@@ -55,17 +55,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
         // extend xDelta to make space for multiple datasets (if ther are one)
         _deltaX *= CGFloat(_data.dataSetCount)
         
-        var maxEntry = 0
-        
-        for (var i = 0, count = barData.dataSetCount; i < count; i++)
-        {
-            var set = barData.getDataSetByIndex(i)
-            
-            if (maxEntry < set!.entryCount)
-            {
-                maxEntry = set!.entryCount
-            }
-        }
+        var maxEntry = barData.xValCount
         
         var groupSpace = barData.groupSpace
         _deltaX += CGFloat(maxEntry) * groupSpace
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -467,9 +467,8 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
         {
             var bd = _data as! BarChartData
             var space = bd.groupSpace
-            var j = _data.getDataSetByIndex(dataSetIndex)!.entryIndex(entry: entry, isEqual: true)
             
-            var x = CGFloat(j * (_data.dataSetCount - 1) + dataSetIndex) + space * CGFloat(j) + space / 2.0
+            var x = CGFloat(entry.xIndex * (_data.dataSetCount - 1) + dataSetIndex) + space * CGFloat(entry.xIndex) + space / 2.0
             
             xPos += x
             
--- a/Charts/Classes/Renderers/BarChartRenderer.swift
+++ b/Charts/Classes/Renderers/BarChartRenderer.swift
@@ -92,8 +92,8 @@ public class BarChartRenderer: ChartDataRendererBase
             var e = entries[j]
             
             // calculate the x-position, depending on datasetcount
-            var x = CGFloat(e.xIndex + j * dataSetOffset) + CGFloat(index)
-                + groupSpace * CGFloat(j) + groupSpaceHalf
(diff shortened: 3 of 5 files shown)


PhilJay:
This looks pretty good to me. Is it tested?
 I'm not sure if there are any more things that need fixing concerning grouped bars.

liuxuan30:
I just tested on my side with a few bar charts. I am not sure about other parts either.

Charts/Classes/Charts/BarChartView.swift
```diff
@@ -55,17 +55,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
         // extend xDelta to make space for multiple datasets (if ther are one)
         _deltaX *= CGFloat(_data.dataSetCount)
         
-        var maxEntry = 0
```
danielgindi: Can you clarify why this change was made?

liuxuan30: Each dataSet will have up to xValsCount entries I think. If the specific xIndex don't have entry, the entry count would less than x values count.

Let's say, we have 4 x values [Jan,Feb,Mar,Apr], if every dataSet does not have entries at xIndex: 2 (Mar), then the maxEntry in old code would be 3. However, we actually need maxEntry to be 4, and we can have correct space to draw the bars. 

I am using below code to test in demo multiple bar charts. It only has data at first index and last index. If you use old logic to calculate maxEntry, then it only draws the first index bars. The last one is missing, and the xAxis is not rendered correctly as well. The maxEntry would be 2 in old logic, however I got 11 xIndex.

Acutally, this is the first place I changed when I found the issue. But not enough to fix the bug.
```Swift
{
    NSMutableArray *xVals = [[NSMutableArray alloc] init];
    for (int i = 0; i < 11; i++)
    {
        [xVals addObject:[@(i + 2015) stringValue]];
    }
    NSMutableArray *yVals1 = [[NSMutableArray alloc] init];
    NSMutableArray *yVals2 = [[NSMutableArray alloc] init];
    
    for (int i = 0; i < 11; i++)
    {
        if (i == 0) {
            [yVals1 addObject:[[BarChartDataEntry alloc] initWithValue:2354235 xIndex:i]];
            [yVals2 addObject:[[BarChartDataEntry alloc] initWithValue:3354235 xIndex:i]];
        } else if (i == 10) {
            [yVals1 addObject:[[BarChartDataEntry alloc] initWithValue:6354235 xIndex:i]];
            [yVals2 addObject:[[BarChartDataEntry alloc] initWithValue:7354235 xIndex:i]];
        } else {
            continue;
        }
    }
    
    BarChartDataSet *set1 = [[BarChartDataSet alloc] initWithYVals:yVals1 label:@"Company A"];
    [set1 setColor:[UIColor colorWithRed:104/255.f green:241/255.f blue:175/255.f alpha:1.f]];
    BarChartDataSet *set2 = [[BarChartDataSet alloc] initWithYVals:yVals2 label:@"Company B"];
    [set2 setColor:[UIColor colorWithRed:164/255.f green:228/255.f blue:251/255.f alpha:1.f]];
    
    NSMutableArray *dataSets = [[NSMutableArray alloc] init];
    [dataSets addObject:set1];
    [dataSets addObject:set2];
    
    BarChartData *data = [[BarChartData alloc] initWithXVals:xVals dataSets:dataSets];
    data.groupSpace = 0.8;
    [data setValueFont:[UIFont fontWithName:@"HelveticaNeue-Light" size:10.f]];
    _chartView.leftAxis.startAtZeroEnabled = NO;
    _chartView.rightAxis.startAtZeroEnabled = NO;
    _chartView.data = data;
}
```

danielgindi:
@PhilJay I went through it - looks like the change in BarChart view is legit!

PR #268, opened 2015-08-04
Minor refactor for BarLineChartViewBase

liuxuan30:
1. change all gestures and their handlers to internal
2. use import UIKit instead of UIKit.UIGestureRecognizer

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 import CoreGraphics
-import UIKit.UIGestureRecognizer
+import UIKit
 
 /// Base-class of LineChart, BarChart, ScatterChart and CandleStickChart.
 public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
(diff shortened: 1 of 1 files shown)


petester42:
What are the benefits of making them internal?

liuxuan30:
then you can override it with your desired actions.

petester42:
The reason I am asking is that those methods are functions specific to those gesture recognizers. Its not as simple as overriding the the subclass. I'd see a case for either making the just the variables internal or just the functions internal. I think both will lead to confusion when users override stuff.

liuxuan30:
If both internals makes confusion, I would like to make the functions internal. 

But sometimes, we may want to replace the gestures to be our sub classes. I used to write a multiple tap gesture recognizer, and when I tried to replace it, I cannot. This is the reason why I make both internal.

It is not simple to override, but they are for advanced users :)

petester42:
I think out of the two making the variables internal is better. The object defines what selector will get called. If you want to change the tap recognized then just replace whole object I your subclass and implement the selector. it also adds more flexibility like changing numberOfTaps etc.

liuxuan30:
You seems right. I don't come up this way. PR updated.

PR #274, opened 2015-08-04
Solved XCode7 abort trap:6 bug

eldafito:
Archiving the project in XCode 7 does not throw compiler exception
anymore

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -411,7 +411,8 @@
 		5BA8EC371A9D14DC00CE82E1 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 0610;
+				LastSwiftUpdateCheck = 0700;
+				LastUpgradeCheck = 0700;
 				ORGANIZATIONNAME = dcg;
 				TargetAttributes = {
 					5BA8EC3F1A9D14DC00CE82E1 = {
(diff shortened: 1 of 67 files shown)


liuxuan30:
Could share why you make the changes this way? What's the difference here.

petester42:
It's a compiler bug. Here is a radar explaining it. http://www.openradar.me/22138918

danielgindi:
Hi!
Thanks for sharing!
There are two problems:
1. The codestyling in your code does not match the code style of the library, and specifically changes existing code to a different, unreadable code style.
2. This is a compiler bug in a *beta* Xcode version, which should be actually fixed by Apple in the next betas...

eldafito:
I can change the codestyling or, if not interested, I'll close the request.

danielgindi:
Let's wait a little for Xcode 7 beta 5 - see if it is fixed. If Apple does not handle that then we'll have to make the workarounds...

eldafito:
ok

petester42:
I actually like this syntax better. It is very clear what the edge conditions are and the variable is only set once so it's easier to follow the assignment.

danielgindi:
We could change the flow to this kind of flow as a refactoring commit....
But it must be clean, not a degradation in quality :-)

danielgindi:
Anyone knows if this was fixed in Xcode 7 beta 5?

petester42:
yes it is. tried the code in the radar on beta-4 and it failed. Beta 5 it compiled fine. :+1:

Issue #284, opened 2015-08-08
[BarChart] single dataSet + force groupSpace=0.8 leads to wrong xAxis label rendering and highlight


liuxuan30:
Bar chart with single dataSet and force groupSpace to be 0.8 will leads to strange behavior for xAxis label rendering and highlight when scaleX > 1.0.

To reproduce, simply return a number say 0.8 for groupSpace, like below, and zoom in and scroll, you will see some xAxis label disappeared, and highlighting also highlight wrong bar.
```Swift
    /// The spacing is relative to a full bar width
    public var groupSpace: CGFloat
    {
        get
        {
            if (_dataSets.count <= 1)
            {
                return 0.8
            }
            return _groupSpace
        }
        set
        {
            _groupSpace = newValue
        }
    }
```
I am not sure if it's intended to return groupSpace 0.0 for single dataSet. It seems like we don't want to support groupSpace for single dataSet?

liuxuan30:
Closing as with single dataSet we should use barSpace directly.

PR #285, opened 2015-08-08
Pending confirmation: Try to fix single dataSet with custom groupSpace issue #284

liuxuan30:
for issue #284 

How to test the difference:
Force groupSpace to return custom value or 0.8 like below:
```Swift
    /// The spacing is relative to a full bar width
    public var groupSpace: CGFloat
    {
        get
        {
            if (_dataSets.count <= 1)
            {
                return 0.8
            }
            return _groupSpace
        }
        set
        {
            _groupSpace = newValue
        }
    }
```

### DIFF
--- a/Charts/Classes/Charts/BarChartView.swift
+++ b/Charts/Classes/Charts/BarChartView.swift
@@ -104,7 +104,7 @@ public class BarChartView: BarLineChartViewBase, BarChartRendererDelegate
     public override var lowestVisibleXIndex: Int
     {
         let step = CGFloat(_data.dataSetCount)
-        let div = (step <= 1.0) ? 1.0 : step + (_data as! BarChartData).groupSpace
+        let div = step + (_data as! BarChartData).groupSpace
         
         var pt = CGPoint(x: _viewPortHandler.contentLeft, y: _viewPortHandler.contentBottom)
         getTransformer(ChartYAxis.AxisDependency.Left).pixelToValue(&pt)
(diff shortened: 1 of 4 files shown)


danielgindi:
@PhilJay do you see any possible sideeffects to this?

liuxuan30:
Why it forces to return 0.0 if only up to 1 dataSet. Is it allowed to return other value instead of 0.0 for this scenario?
I am not sure if it makes sense to return positive values. If not, then we can close it. I got this issue by QA testing.

danielgindi:
@liuxuan30 I haven't had the time yet to look at the issue with grouped bars in combined charts... Did you have success with it? What's the status of this PR - does it correct it?

liuxuan30:
@danielgindi I am occupied with some dev work recently, I will update as soon as I got some time.

liuxuan30:
get some time to merge with master. The only problem still remains:
```
Why it forces to return 0.0 if only up to 1 dataSet. Is it allowed to return other value instead of 0.0 for this scenario?
I am not sure if it makes sense to return positive values. If not, then we can close it. I got this issue by QA testing.
```

liuxuan30:
Today I have been thinking about this, groupSpace should be used when dataSets count > 1. Only one dataSet and setting groupSpace seems not make sense?

Besides, I found a bug that when trying to highlight the bars, the xIndex is not correct, due to the groupSpace is 0.8, not 0. Using 0.0 is correct highlighting. So I am thinking there must be some reasons to separate them. Closing this PR right now.

PR #291, opened 2015-08-10
Fixes xcode beta 5 compile errors

### LINKED ISSUE #287: ios-charts for swift 2.0 cannot compile on XCode7 beta 5 - self.dynamicType.allocWithZone(zone) issue
Hi

Thank you for your hard word and fantastic ios-charts.

I noted that the latest ios-charts for swift 2.0 cannot compile on XCode7 beta 5 due to the following issue.
The self.dynamicType.allocWithZone(zone) produce the following error:
'allocWithZone' is unavailable in Swift: use object initializers instead

There are a couple of error occurrences.
1- ChartDataEntry.swift
2- ChartDataSet.swift

// MARK: NSCopying

public func copyWithZone(zone: NSZone) -> AnyObject
{

    let copy = self.dynamicType.allocWithZone(zone) as ChartDataSet
    -------
    --------
    return copy
}
I would love finding the trick on my own but.. :-(

Kind Regards,
Patrick

petester42:
I know its a lot of changes but this beta made the `UIGraphicsGetCurrentContext()` now return an optional value. So instead of adding `if let` or `guard` everywhere I decided to refactor the methods that took `CGContext` to `CGContext?` and add extensions to methods that didn't take optional values. Hopefully apple moves the 2 methods that I extended to take in optional values as well so the extensions can be removed. (fixes #287)

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		06A5D1861B78675500915098 /* UIGraphics+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06A5D1851B78675500915098 /* UIGraphics+Extensions.swift */; };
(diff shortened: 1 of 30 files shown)


AJ9:
+1

PR #297, opened 2015-08-12
Changed Release signing configuration to "iOS Developer" to improve Carthage compatibility

JaviSoto:
Carthage was failing to build this. I believe that with `iPhone Distribution` you need a special profile to sign this library. It works with simply `iPhone Developer`. This shouldn't be an issue for anyone because the frameworks end up being signed with your application's profile when submitting to Apple.

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -658,7 +658,7 @@
 		5BA8EC581A9D14DC00CE82E1 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CODE_SIGN_IDENTITY = "iPhone Distribution";
+				CODE_SIGN_IDENTITY = "iPhone Developer";
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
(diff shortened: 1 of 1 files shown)


danielgindi:
Why would carthage fail to build this? Doesn't it use your project's certificate?

JaviSoto:
I don't think so.
It fails because I don't have a distribution certificate with an associated profile for that bundle ID, but I do have a wildcard development one.

This is the error that building with Carthage outputs:

```
Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) were found.
CodeSign error: code signing is required for product type 'Framework' in SDK 'iOS 8.0'
A shell task failed with exit code 65:
** BUILD FAILED **
```

danielgindi:
So you do not have a distribution certificate for your project at all? Maybe that is what you should fix?..

danielgindi:
The thing is if I accept this PR then this might make trouble for those with in-house distribution

JaviSoto:
I think it's much more likely that people will develop apps, at least at the beginning, with only a development certificate. This is how all the other projects I've seen are set up, and it's the Xcode default I think.

danielgindi:
I agree

JaviSoto:
Thanks :)

PR #306, opened 2015-08-18
Fix for crash in the case all y values were negative and...

pajai:
... startAtZeroEnabled was set. I fixed the logic to compute the yAxis range. Previously we were getting a range of 0, and since we use it when computing the transformation matrix as divisor we were getting a matrix with +Inf and NaN values which was causing a crash later on.

Cheers,
Patrick

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -277,6 +277,10 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
         setNeedsDisplay()
     }
     
+    private func sign(n: Double) -> Double {
+        return n > 0 ? 1 : (n < 0 ? -1 : 0)
+    }
+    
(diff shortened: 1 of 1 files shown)


liuxuan30:
Is this fix for recent changes about startAtZeroEnabled and range calculation? Seems like older versions don't have this issue, if all y values negative.

pajai:
I am not sure if this patch fixes a recent change, because I have not followed ios-charts' history. I could just observe the bug on my side with the latest ios-charts version from CocoaPods or github's trunk.

danielgindi:
The logic for startAtZeroEnabled has changed, and this commit conflicts with it
Please check if your issue still happens with the latest changes

danielgindi:
If the issue still happens - please make a new PR, this can't be merged...

pajai:
Sorry, I was quite busy lately.

Ok, I checked the behavior of the latest ios-chart version for all negative y values and it seems fine so far. All good. I am switching the dependency back to the official version.

Thx!
Cheers,

PR #313, opened 2015-08-23
fix pie chart clipping

### LINKED ISSUE #309: Take selectionShift into account when displaying pie chart
For pie charts, the chart will get cut off when selected since the main circle does not take into consideration how far the selection will go. A fix would be to take the `selectionShift` into account when drawing the main circle. So instead of the current `width` and `height` it would be `width -  (selectionShift * 2)` and `height - (selectionShift * 2)`. This would make the chart inset by the `selectionShift` so that when selected the chart would fit inside the view. An example of this happening can be seen below.

clipped-pie-chart

petester42:
fixes #309

Takes the `selectionShift` into account when calculating the `circleBox`.

### DIFF
--- a/Charts/Classes/Charts/PieChartView.swift
+++ b/Charts/Classes/Charts/PieChartView.swift
@@ -84,11 +84,17 @@ public class PieChartView: PieRadarChartViewBase
         
         var c = centerOffsets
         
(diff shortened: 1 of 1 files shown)


danielgindi:
Thank you :-)

Issue #323, opened 2015-08-27
CombinedChart with BarChartData cuts the first and last bars in half


zamith:
Everything works fine apart from the fact that the first and last bars are cut in half, this might be something that is expected, if so, how can I disable it?

Here's how it looks:

<img width="305" alt="screen shot 2015-08-27 at 7 45 10" src="https://cloud.githubusercontent.com/assets/166304/9529792/2cb8599e-4cf4-11e5-9416-3748aef43930.png">

liuxuan30:
By default it should not be half.. I am not sure how much you changed? The code will leave some space for the first and last bar. I never see it on my side. You might need to provide more info. Have you changed any view port/offset properties or transformer matrix, etc. Is it zoomed or something.

You can also try your data with the ChartsDemo code to see if it is reproducible.

zamith:
I haven't changed any viewport things I think, nor should it be zoomed. However, we've decided to go with a linebar, which works fine, so I'm closing this issue.

Sorry about the trouble.

liuxuan30:
Well it never should be half a bar if you just setup the dataSets. If it is a bug (which I doubt), you could help provide some information.

zamith:
Ok, I was able to reproduce it on the Demo. If I comment the setting of the bubbleData, the same behavior occurs:

<img width="432" alt="screen shot 2015-08-28 at 3 31 59" src="https://cloud.githubusercontent.com/assets/166304/9537824/66cda9d6-4d35-11e5-9565-dff1149a1c28.png">

And with the bubbleData:

<img width="432" alt="screen shot 2015-08-28 at 3 32 56" src="https://cloud.githubusercontent.com/assets/166304/9537832/80bb7df0-4d35-11e5-9dd5-053ef259b858.png">

liuxuan30:
Are you using combined chart in your project? How about normal bar chart?

zamith:
Yes, I was using combined chart. The normal bar chart works fine.

liuxuan30:
Ah hah, it's actually a bug. if you don't have bubble chart, it will use another value to determine the chartXMin and chartXMax.

In CombinedChartView -> override func calcMinMax():
```Swift
            if (self.bubbleData !== nil)
            {
                for set in self.bubbleData.dataSets as! [BubbleChartDataSet]
                {
                    let xmin = set.xMin
                    let xmax = set.xMax
                    
                    if (xmin < chartXMin)
                    {
                        _chartXMin = xmin
[...]

PR #324, opened 2015-08-28
combined chart - seems we should use same chartXMin and chartXMax even there is no bubble data

liuxuan30:
For issue #323 
seems we should use same chartXMin and chartXMax
            _chartXMin = -0.5
            _chartXMax = Double(_data.xVals.count) - 0.5
even there is no bubble data


danielgindi:
Thanks :-)

PR #331, opened 2015-08-31
Fix a chart marker animation bug for horizontal bar view

liuxuan30:
fix an chart marker animation bug for horizontal bar view.

`getMarkerPosition` is calculating its own points, so we need to do the invertion for horizontal bar chart view. Maybe you can come up a better enhancement or structure.


danielgindi:
The whole marker thing is fixed in d163081, 64851c5. Thanks!

Issue #280, opened 2015-08-06
[feature request] support truncating lines if the xIndex in dataSet not continuous


liuxuan30:
There should be demand when we add xIndex but no entry, the line should be truncated, not goes to next entry.

danielgindi:
Currently the behaviour in line charts for missing y values is "undefined". We are going to define it soon, and possibly allow more control over that decision.

liuxuan30:
I tried to implement mine last week, it could break the lines. But for the filling function, it's still buggy. The filling rect is quite complicated when involving negative values. It seems the filling rect will always use y=0 axis line, so when there is only negative values, it fill the rect above line, below y=0 axis.

noais:
@liuxuan30 have you some news about this issue?

liuxuan30:
Well I could break the lines, but the rect filling part is still buggy and I gave up, too many corner cases for my implementations and I don't have much time to think carefully.

noais:
Do you have this alterations on you branch? Maybe I can try to help you.

liuxuan30:
I don't have yet, you can have your implementations and better than mine :)

Breaking the lines is simple. Just start another moveToPoint loop and close previous path properly. I feel the hard part is the filling function. Consider if there are all negative values, and the y axis is at the top... and half negative and half postive..

noais:
This week end I'm going to try implement this feature.

PR #335, opened 2015-08-31
Issue #280

noais:
To enable this feature needs to set lineBreakerEnabled on lineDataSet.

With this we can have break lines, the only thing we need to do is to jump a index.

For ex:
    [yVals addObject:[[ChartDataEntry alloc] initWithValue:5 xIndex:1]];
    [yVals addObject:[[ChartDataEntry alloc] initWithValue:5 xIndex:2]];
    [yVals addObject:[[ChartDataEntry alloc] initWithValue:5 xIndex:4]];
    [yVals addObject:[[ChartDataEntry alloc] initWithValue:5 xIndex:5]];

I change the first line char example have some example. (This alteration need to be deleted).

I made this pull request to have yours opinion, if you like like it was I go to clean the code.

### DIFF
--- a/Charts/Classes/Data/LineChartDataSet.swift
+++ b/Charts/Classes/Data/LineChartDataSet.swift
@@ -34,6 +34,9 @@ public class LineChartDataSet: LineRadarChartDataSet
     
     public var drawCircleHoleEnabled = true
     
+    // if true, line breaks on void values
+    public var lineBreakerEnabled = false
+    
     public override init()
     {
         super.init()
@@ -101,6 +104,8 @@ public class LineChartDataSet: LineRadarChartDataSet
     public var isDrawCubicEnabled: Bool { return drawCubicEnabled; }
     
     public var isDrawCircleHoleEnabled: Bool { return drawCircleHoleEnabled; }
+    
+    public var isLineBreakerEnabled: Bool { return lineBreakerEnabled; }
     
     // MARK: NSCopying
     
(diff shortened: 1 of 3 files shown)


liuxuan30:
Thanks for your effort! 

Some things needed to be improve:
1. the code style, especially the {} around if else
2. I just see you changed drawLinear, but drawCubic is not touched
3. drawLinearFill and drawCubicFill is not touched yet.

As a feature, I think it would be great to have them all in place (see the tough part? :-) )

noais:
Ok, I'm going to finish all around linear (drawLinearFill and better format). 
Do you know some tool to format automatically swift code?

This week end I start to implement to drawCubic.

I have done this pull request more to see if you all agree with my approach to implement this new feature.
I'm going to reject my pull request and when I'm really finish I create a new pull request.

liuxuan30:
I don't know the tool, I format manually since ios-charts has its own format. 
PS, the filling feature is tricky, remember to test 'pure negative values' and 'positive+negative values', these are the monsters when I tried to implement but got a lot of visual effect bugs.

noais:
http://giphy.com/embed/XpdYVfb6BEm4w

osolano:
@liuxuan30  @noais  I was able to make line gaps by setting the color array of the linedatachart set to the same color as the background for a missing data point.

For example, if there are 4 data points missing, and the 3rd one is 0 and we want to show a gap here, i would set the 3rd color in the color array to match the background of the grid. This make it look like there was a line break.

image

noais:
@osolano this is a work around. For instance this solution doesn't work for me, because I have a gradient colour on background...

osolano:
@noais yes purely workaround, just wanted to point it out in case you didn't have a solution before. Cheers!

liuxuan30:
@osolano I would start working on line breaking feature 'again' maybe next month and try to fix the buggy filling portion while line is broke. Once it's done, I will try file a PR for this. But still, huge work to do, so no ETA

noais:
@liuxuan30 I start already this implementation but I never had time to finishing. I don't know if you want to start from 0 or you want to see what I have made?

liuxuan30:
I had my own too :) it's just about fix those annoying bugs while filling

JCMcLovin:
Did this ever get merged and is the feature supported in version 4.0?

PR #343, opened 2015-09-07
Add option to set a color per label on xAxis for a RadarChart

slash7:
Related to issue https://github.com/danielgindi/ios-charts/issues/94

I've added an option to set different colors for each label on the xAxis for a RadarChart.

Please let me know what you think. Maybe this should handle yAxis too ?

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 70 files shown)


danielgindi:
Well, if this is only for RadarChart, then it should be only exposed in Radar chart related classes.
Otherwise, you should implement it in all charts...

Also, another way would be to only use an array, and use something like `labelTextColors[x % labelTextColorsCount]`.
But of course there are performance tests to be done ;-)

danielgindi:
This was closed automatically after merging `swift-2.0` to `master`.
Please rebase on `master` and reopen the PR :-)

lingzlu:
is this checked in? I don't see the labelTextColors properties under xAxis in version 3.0.2

PR #360, opened 2015-09-14
Fix default value of forceLabelsEnabled

yas375:
I believe the default value should be `false`. It's [false on Android version of the library](https://github.com/PhilJay/MPAndroidChart/blob/2e6bfeb87e2bb0a02fb6350b96a454d77f7c088b/MPChartLib/src/com/github/mikephil/charting/components/YAxis.java#L47). Also I've noticed documentation for `isForceLabelsEnabled` [says](https://github.com/danielgindi/ios-charts/blob/0f33d9fc1e06b0d3685539f3bea5d61bc63000ba/Charts/Classes/Components/ChartYAxis.swift#L238) it's false by default as well.

---

Without the fix:
without

With the fix:
with fix

Note that the values on y axis are nicely calculated.

As a temp workaround fix we can do in our code:

```objective-c
self.chartView.leftAxis.labelCount = self.chartView.leftAxis.labelCount;
```

### DIFF
--- a/Charts/Classes/Components/ChartYAxis.swift
+++ b/Charts/Classes/Components/ChartYAxis.swift
@@ -54,7 +54,7 @@ public class ChartYAxis: ChartAxisBase
     public var startAtZeroEnabled = true
     
     /// if true, the set number of y-labels will be forced
-    public var forceLabelsEnabled = true
+    public var forceLabelsEnabled = false
 
     /// the formatter used to customly format the y-labels
     public var valueFormatter: NSNumberFormatter?


danielgindi:
Thanks :-)

yas375:
@danielgindi thank you for the lib! ;)

Any chance it can be merged into Swift 2.0 branch too? What's current strategy on getting updates to that branch? Do you merge it with master from time to time or should I create a similar pull request with this fix to that branch? Since Xcode 7.0 is now available through AppStore, maybe it makes sense to merge Swift 2.0 branch into master now?

danielgindi:
Yes it will be merged very soon into master :-)
>
> Before that I need to go over the diff line by line and make sure that
nothing was broken by mistake.
It's not gonna take more than a day or two!

yas375:
Awesome! Thanks! :sunny:

PR #362, opened 2015-09-15
Allow graphs with a single axis

msanders:
This allows for the following graph:

screenshot 2015-09-14 17 21 48

As opposed to the current behavior, which results in this:

screenshot 2015-09-14 17 24 10

Which I do not want, since the values in this case are always integral.

### DIFF
--- a/Charts/Classes/Components/ChartYAxis.swift
+++ b/Charts/Classes/Components/ChartYAxis.swift
@@ -138,9 +138,9 @@ public class ChartYAxis: ChartAxisBase
         {
             _labelCount = 25
         }
-        if (_labelCount < 2)
+        if (_labelCount < max(1, Int(self.customAxisMin)))
         {
-            _labelCount = 2
+            _labelCount = max(1, Int(self.customAxisMin))
         }
     
         forceLabelsEnabled = force


danielgindi:
Hi :-)
I don't quite understand why are you using `customAxisMin` as a label count... It is a value for the axis minimum, not a value for a label count.
Any logic I'm missing here?

msanders:
I wasn't sure which property to use, another option would be to have a new property `minLabelCount` which defaults to 2 or to just change the default from 2 to 1 (not sure if that would break anything)

danielgindi:
Closing due to code being illogical...

msanders:
Hi Daniel,

Do you have any suggestions on how to improve this? Is there a reason the default is 2 instead of 1?

Thanks,
Michael

danielgindi:
This is in order to always give a reference regarding the Y axis scale. With only 1 label, you have no idea where you stand...

msanders:
Did you see the above screenshots? It is broken without this change. Do you have another suggestion on how to fix it?

danielgindi:
I do not see how the screenshot is incorrect...

danielgindi:
Please discuss this in the issues section, so we can understand your issue better and solve it

msanders:
I will write up some example code later today. The problem is the values in the y axis are supposed to be integral, and since the minimum label count is 2 ios-charts unnecessarily adds an intermediary axis. The same bug exists in Android as far as I can tell.

PR #367, opened 2015-09-16
Convert to swift 2 and compile with iOS 9 SDK

icecrystal23:
I ran the swift language converter and fixed a handful of compilation errors in Xcode 7


liuxuan30:
I remember there is a swift-2.0 branch already?

icecrystal23:
Oh, I didn't even notice.  Let me see if my tvOS branch works on top of that one instead.

PR #369, opened 2015-09-16
Feature/tv os swift 2.0

icecrystal23:
Add a target for tvOS and get it to compile with Xcode 7.1 beta


icecrystal23:
Meant to make this into swift-2.0 branch

PR #371, opened 2015-09-16
Option to start the y axis at at least zero

icecrystal23:
Add new option to yAxis "allowStartAtNegative", which will let you start the graph at a minimum value of 0 (but not enforce 0 if it would've started higher than zero)


danielgindi:
This was closed automatically after merging `swift-2.0` to `master`.
Please rebase on `master` and reopen the PR :-)

PR #374, opened 2015-09-17
Allow arbitrary line widths

icecrystal23:
We wanted a line chart with multiple lines and also with some points as circles (without lines in between the circles); this was the easiest way to do this as far as I could tell.


icecrystal23:
Hmm, maybe I can do this with a CombinedChartView... I will try that

Issue #289, opened 2015-08-09
Release Version 2.1.3 on CocoaPods


sgade:
I noticed the current release [v2.1.3](https://github.com/danielgindi/ios-charts/releases/tag/v2.1.3), although tagged, is not available on CocoaPods.
I think the problem is that the [PodSpec](https://github.com/danielgindi/ios-charts/blob/master/Charts.podspec#L3) is not up to date.

Would be great to use 2.1.3 as it has some nice fixes.

petester42:
fixed. enjoy :smile:

sgade:
Cocoapods.org seems to have some issues right now so I can't see the update but thanks for the effort. :+1:

Issue #283, opened 2015-08-06
installing with cocoapods


peteclark3:
Hi, I've read the various issues on this topic but haven't found a solution.  Closest I came was issue #87 but still didn't see a solution other than "read the instructions" which I swear I did a couple of times.. I apologize.

I simply added to my Podfile: pod 'ios-charts'

then did a pod install / pod update .. 

all seems good from a building perspective, but when I try to "import Charts" at the top of my swift file I get a compiler error, "No such module 'Charts'" .. similarly if I remove that and just try to add a BarChartView, e.g.  @IBOutlet weak var barChartView: BarChartView! , I get "use of undeclared type BarChartView"  

Is there a step I am missing for the cocoapods installation?  Tried multiple project cleans, stop and start xcode..

xrolfex:
You may need to make sure that you are using frameworks in cocoa pods.

peteclark3:
I was.. (I am) - gave up and installed via tutorial here: http://www.appcoda.com/ios-charts-api-tutorial/

khanxc:
![screen shot 2016-01-28 at 6 35 07 pm] Other framework in pod are working, trying to make Charts work for hours but no progress. Guidance needed
(https://cloud.githubusercontent.com/assets/5825355/12645103/b7a84646-c5ed-11e5-9afe-b0c2090c71b4.png)
Podfile:
```Swift
 platform :ios, '8.0'
use_frameworks!
target 'AppName' do

pod 'Alamofire', '~> 3.0'
pod 'Charts'

end

jhmacdon:
I am having the same issue

danielgindi:
You may want to post an Issue on the Issues page of CocoaPods. Or try to install it manually...

apprisingapps:
It looks as if you are using Swift.  If you are using Swift 3, you may need consider adding the following to your pod file.
post_install do |installer|
    installer.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        config.build_settings['SWIFT_VERSION'] = '3.0'
      end
    end
  end

H-Sha:
It worked for me - please try the following in your Podfile

source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '10.0'
use_frameworks!

target 'AppName' do
pod 'Alamofire', '~> 4.3.0'
pod 'Charts'
end

elchris78:
I've been dealing with the same problem since 2 days ago, with no success so far though!
[...]

Issue #277, opened 2015-08-05
Cocoapods support


gkaimakas:
Hello,

this is a great framework, been using it on Android and it is totally awesome! Can you please publish it on Cocoapods so it can be easier to use?

P.S. It is really easy to publish a project on Cocoapods

danielgindi:
Man it's already on...

gkaimakas:
Oh, didn't check cocoapods for it. Sorry.

Please update your README.md so that the support is clear.

Thanks

2015-08-05 18:01 GMT+03:00 Daniel Cohen Gindi <notifications@github.com>:

> Man it's already on...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/danielgindi/ios-charts/issues/277#issuecomment-128027332>
> .
>



-- 
Γιώργος

Issue #47, opened 2015-04-24
Cannot find this repo through Cocoapods.org


Apolotary:
edit: Please disregard the original question about Cocoapods support.

I can't seem to find this repo from the [Cocoapods website](https://cocoapods.org/?q=ios-charts), can anyone push it to the main trunk?

edit 2: found it, it's called [Charts](https://cocoapods.org/?q=charts). Sorry for the trouble guys

PR #376, opened 2015-09-17
be explicit on how to install 'Charts' when using CocoaPods since

codeHatcher:
I noticed many people were confused as how to use CocoaPods to install the framework (myself included). I accidentally added the 'ios-charts' pod instead of 'Charts' and spent 10 minutes trying to figure out why it wasn't working. I think a blurb would help many people out.

It's possible it was a contributing factor to these guys having an issue: #277 #283 #289 (mentioning CocoaPods.org having issues) and definitely the reason #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ If you want to compile for iOS 7:
 1. Drag the code itself (.swift files) to your project. As sadly, Swift currently does not support compiling Frameworks for iOS 7.
 2. Make sure that the files are added to the Target membership.
 
+## CocoaPods Install
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,10 @@ If you want to compile for iOS 7:
 1. Drag the code itself (.swift files) to your project. As sadly, Swift currently does not support compiling Frameworks for iOS 7.
 2. Make sure that the files are added to the Target membership.
 
+## CocoaPods Install
+
+Add `Pod 'Charts'` to your Podfile. Adding `Pod 'ios-charts'` will not work!!
```
yas375: It should `pod ...` not `Pod ...`

PR #377, opened 2015-09-17
Swift 2.0

fiona7:


danielgindi:
It will be merged manually after carefully going through the code and making last updates

Issue #315, opened 2015-08-24
Stop axis labels duplicating.


steprescott:
Within the App I have an issue where the labels on the left axis of either a `BarChartView` or a `LineChartView` repeats.

I first added the snippet below to ensure that the number of labels on the axis does not exceed the maximum value in the y data.

```
self.chartView.leftAxis.labelCount = self.chartView.data.yMax > YAxisMaxLabelCount ? YAxisMaxLabelCount : self.chartView.data.yMax;
```

This ensured that if the maximum value in the y data was `3` then there would only be `3` labels within the axis. Now this worked fine if the graph was not interactive, i.e. the user could not zoom the chart.

As soon as the user zooms obviously the count of the labels remains the same and the value formatter fills in the gaps allowing a value to repeat.

*Example of axis with no duplicated values*
![screen shot 2015-08-24 at 5 07 29 pm](https://cloud.githubusercontent.com/assets/1131458/9445187/b72e7c6c-4a82-11e5-934d-fb2a8a0fa691.png)

*Example of same chart but zoomed in*
![screen shot 2015-08-24 at 5 07 54 pm](https://cloud.githubusercontent.com/assets/1131458/9445197/c545bd92-4a82-11e5-9405-aac26c850be0.png)

The screenshots show the left axis repeating the value `2`. What is your best suggestion on solving this repeating label issue? Any help would be grateful. If you need any further information let me know.

Thanks

liuxuan30:
I think it is not the label duplicated, but it contains decimal values and a integer formatter? You need to check what's the raw value for each x axis label and how you format it.

steprescott:
I was able to do this by adding a check to ensure the interval each label is always a whole value.
The change is within the `computeAxisValues()` function of the `ChartYAxisRenderer` class

```
var interval = ChartUtils.roundToNextSignificant(number: Double(rawInterval))
if(_yAxis.valueFormatter?.maximumFractionDigits == 0)
{
    interval = interval < 1 ? 1 : floor(interval)
}
```

If the value formatter request there to be no fractional digits then it floors the intervals and if it is less then the whole number 1 it sets it the interval to `1`.

Would you like a pull request?

liuxuan30:
If you are having decimal values, why maximumFractionDigits is 0? I think it's the formatter issue.

steprescott:
I don't want decimal values that is why I set `maximumFractionDigits = 0`

liuxuan30:
[...]

Issue #256, opened 2015-07-30
viewPortHandler.setMaximumScaleX not working for horizontal bar chart


liuxuan30:
I am trying to set max scaleX for horizontal bar chart via `viewPortHandler.setMaximumScaleX`, let's say the maxScaleX to be 3.

However when zooming in on xAxis for horizontal bart chart, it can zoom in as much I want.

After the investigation, I found in below function `matrix.a` is always 1, but `matrix.d` is keeping increasing as the correct scale in `limitTransAndScale`. It seems like we forgot to switch them for horizontal bart chart, or do you intend to do so? 
```Swift
    private func limitTransAndScale(inout #matrix: CGAffineTransform, content: CGRect?)
    {
        // min scale-x is 1, max is the max CGFloat
        // for horizontal bar chart, matrix.a is always 1
        _scaleX = min(max(_minScaleX, matrix.a), _maxScaleX)
        
        // min scale-y is 1
        _scaleY = max(_minScaleY, matrix.d)
```
Also, I want to propose could we change the gestures and their handlers to be 'internal' rather than 'private'?

The pinch gesture handler is private and I cannot override it to adapt for horizontal bar charts, though we could use `is HorizontalBarChartView` in BarLineChartViewBase and make some changes, just like `performPanChange`

I tried to fix it but I found it's more complicated than I thought. I added setMaximumScaleY, and have to change `canZoomInMoreX` and have to get chart type in ChartViewPortHandler in order to read maxScaleY for horizontal bar chart, which seems not a good design pattern.

Needs your help how to fix this.

danielgindi:
About `limitTransAndScale`, the changes should probably be in a subclass of ViewPortHandler, flipping the functions. But I think we'll leave it be for now, as we may be changing the transformer's internal logic when switching to x-values instead of x-indexes

As for the gestures - I have no objection to making it internal...
We could go other way, checking with `is`, or making it in the subclass. If you see that there's too much overhead when doing it in a subclass, you could create a property for "rotated axis"

liuxuan30:
@danielgindi I was stuck in how to read correct _maxScaleX in ChartViewPortHandler. I have to obtain chart type in ChartViewPortHandler, which I think is not a good design?

danielgindi:
Yes it is, the better way should be that the ViewPort knows that the axis is rotated (x=y, y=x)

liuxuan30:
Got you. I will take a look at the subclassing way and 'rotated axis' way.

[...]

PR #388, opened 2015-09-18
Allow setting maximum y-scale factor

noais:


liuxuan30:
Hey noais, thanks for your PR! There is a issue about maxScaleY:https://github.com/danielgindi/ios-charts/issues/256. You will see Daniel said `maxScaleY is just a much more complex issue :-)`

So I hope you could read the post, and test more. It's not as easy as you would think  first about maxScaleY :)

noais:
I implement this because there aren't a way to stop the zoom in on the y axis, like the x axis. With this feature Is possible avoid labels duplicated. I implemented this because I have the same issue like #315

noais:
I need to read more carefully the issue #256 because I don't get the problem.

liuxuan30:
@noais, I remember the problem I had is for horizontal bar chart. I don't know why maxScaleY is complcated as @danielgindi said, but when I think I face the same problem, horizontal bar chart's xAxis is like yAxis, and the maxScaleX issue on horizontal bar chart is same as maxScaleY issue on bar chart I guess.

Anyway, you may want to test more on bar/line/horizontal bar chart for maxScaleY. @danielgindi, do you remember why maxScaleY is complicated?

noais:
@liuxuan30 I found this wrong behaviour:
http://imgur.com/chYcHLV

Is this one the problem?

This can be resolved when we have the max Y we can block the all zoom in events.

noais:
This problem happen with setMaximumScaleX and this is already in the main project.
I don't see the diference between setMaximumScaleX with setMaximumScaleY.

This behaviour must be fixed, for both cases beyond this problem I don't see what issue can have maxScaleY specifically?

liuxuan30:
if you are confident, just test different charts as many as possible with different situations.

danielgindi:
There were a few reasons:
1. That conversation at #256 was in the context of the plans to change x-indexes into x-values. Along with supporting logarithmic y-axis etc. We are looking at ways to implement those, and it requires mainly a rewrite of the Transformer and the ViewPortHandler. We are lucky that they are different classes and not just code mixed in random places in the project...
2. There's the possible collisions with auto-scaling and other features that work on the Y axis

But this is fine as a solution for the time being - as it seems to work in "normal" situations.

Thanks :-)

PR #392, opened 2015-09-18
Add target for tvOS and get it to compile

icecrystal23:


danielgindi:
Does tvOS has not gestures?

icecrystal23:
Not these ones, apparently. They are marked TV_OS_UNAVAILABLE or whatever the macro is in the latest beta

danielgindi:
Thanks :-)

PR #394, opened 2015-09-18
Allow for "circle holes" in scatter plots, too

icecrystal23:


danielgindi:
What do we want here exactly?
A circle with a different fill color? Or a circle with an actual hole?

Because I'm thinking maybe of maybe making this more generic (A square could also have a hole...)

icecrystal23:
Different fill color.  I basically just copied this code from the line graph into the scatter plot.  A more generic solution would be great, I just didn't have time to implement it.

PR #408, opened 2015-09-22
Repeated line removed from example

noais:
Repeated line removed from example


danielgindi:
It should actually be `[_chartView.viewPortHandler setMaximumScaleX: 2.f];`

noais:
:+1: Do you want I make a new pull request?

danielgindi:
That would be nice

PR #412, opened 2015-09-23
 Set maximumScaleY

noais:


noais:
closed needs to squash all commits

Issue #406, opened 2015-09-22
Crash by initialize() of CombinedChartView


venice81:
I don't speak English well.

CombinedChartView.initialize() -> () (CombinedChartView.swift:40)
CombinedChartView.initialize() -> () (CombinedChartView.swift:40)
ChartViewBase.init(frame : CGRect) -> ChartViewBase (ChartViewBase.swift:145)
BarLineChartViewBase.init(frame : CGRect) -> BarLineChartViewBase (BarLineChartViewBase.swift:75)
@objc CombinedChartView.init(frame : CGRect) -> CombinedChartView (CombinedChartView.swift:0)
-[PDNoteBMIChangeWeightDetailView setupChartView] (PDNoteBMIChangeWeightDetailView.m:66)

I found suspicious part in BarLineChartFillFormatter : 1650

private weak var _chart: BarLineChartViewBase!

I erased the 'weak'. so it's not crash anymore.
What right have I modified the way?

liuxuan30:
First, you should not delete 'weak', it is supposed to be weak I think.

I checked `getFillLinePosition`, it only has a function `getFillLinePosition`. 
```Swift
        let filled = generateFilledPath(
            entries,
            fillMin: delegate!.lineChartRendererFillFormatter(self).getFillLinePosition(
                dataSet: dataSet,
                data: delegate!.lineChartRendererData(self),
                chartMaxY: delegate!.lineChartRendererChartYMax(self),
                chartMinY: delegate!.lineChartRendererChartYMin(self)),
            from: minx,
            to: maxx,
            matrix: trans.valueToPixelMatrix)
```
So start from here.

Next time you see a crash, remember to post the back trace here, with the top most ones, that can be used to identify the issue.

KerryJava:
* thread #1: tid = 0x310af2, 0x00000001022368ef gpxj`gpxj.CombinedChartView.initialize (self=0x00007fcf385e4fb0)() -> () + 207 at CombinedChartView.swift:40, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x00000001022368ef gpxj`gpxj.CombinedChartView.initialize (self=0x00007fcf385e4fb0)() -> () + 207 at CombinedChartView.swift:40 [opt]
[...]

PR #416, opened 2015-09-24
fix crash issue, since there is getter to make sure it is not null (Lazy…

KerryJava:
… init), there is no need to keep this line

fix:
https://github.com/danielgindi/ios-charts/issues/406


liuxuan30:
Let's hold for a while until we know what happens here causing the crash.

liuxuan30:
Per the discussion in #406, I think it's better to assign it to a local variable and then assign it to `fillFormatter`. deleting the line is not good, since there is one method being used by line chart renderer

danielgindi:
This is clearly not a fix, but removing a feature. The problem is in the compiler.

PR #420, opened 2015-09-25
Start of pan gesture should not be cancelled by no drag

niraj-rayalla:
When checking for the start of a pan, the pan should not be cancelled when the recognizer's translationInView returns 0 in the direction of scrolling. Seems the iPad more easily returns a zero value in the beginning causing the drag to get cancelled immediately even though the actual drag hasn't started because performPanChange() thinks the chart can't be scrolled anymore. So check to see if the drag has vaules that are not actually 0 before calling performPanChange() to prevent immediate pan cancels.


danielgindi:
Nice catch! :+1:

PR #438, opened 2015-10-02
Only alow scaling further if the user can still zoom (Fixes #437)

### LINKED ISSUE #437: Minimum / maximum scale causes the chart to slide sideways when trying to pinch zoom further
Hi,

I've seen this bug when a maximum scale is configure and the user tries to pinch zoom the chart further.

To reproduce it, you can just set a maximum scale on the Candle Stick chart in the Demo and then try to zoom in being that maximum:
[[_chartView viewPortHandler] setMaximumScaleX:5];
[[_chartView viewPortHandler] setMaximumScaleY:5];

The chart just starts sliding (quite fast) sideways.

iangmaia:
Trying to fix issue when there's a maximum / minimum scale and the user tries to pinch the chart further, causing it to slide sideways.


iangmaia:
I've played around with it a bit more and I realized it fixes the issue only when pinchZoom is enabled. When it is disabled, unfortunately it is still there :-(
Ideas?

PR #457, opened 2015-10-07
fixed noDataText and NoDataTextDescription texts ovelapping issue

zntfdr:
At the moment, noDataText and NoDataTextDescription are drawn overlapped, this pull request fixes the issue.


petester42:
Nice catch. Thanks!

PR #464, opened 2015-10-12
Approved, pending styling: Fix Scroll issue when the graph is in a UITableView

coupgar:
foundScrollView  will be UITableWrepperView Not UITableView, which will cause TableView Scroll issue

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -981,6 +981,14 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
             {
                 scrollView = scrollView?.superview
             }
+            
+            // If there is two scrollview together, we pick the superview of the inner scrollview.
+            // In the case of UITableViewWrepperView, the superview will be UITableView
+            if let superViewOfScrollView = scrollView?.superview where superViewOfScrollView.isKindOfClass(UIScrollView)
+            {
+                scrollView = superViewOfScrollView
+            }
+
             var foundScrollView = scrollView as? UIScrollView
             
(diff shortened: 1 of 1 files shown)


danielgindi:
Nice catch!
Although this PR is mis-styled...
Please clean it up, match the code styling of the project, squash everything together and then we'll merge ;-)

danielgindi:
Although, wouldn't we want to check for the specific case of `UITableViewWrapperView` via the class name? or does this apply to all UITableViews which are wrapped in UIScrollViews?

coupgar:
I was worried that Apple won't be happy if we use `UITableViewWrapperView` class name to check this case. 
I will take the advice and update the pull request. :-)

petester42:
I don't think referencing a non public class is ideal if it works with UIScrillView. Is there scrolling issues within a collection view?

coupgar:
Actually you cannot scroll up/down when the chart view is in a `UITableView`,  Collection View has not been tested @petester42

petester42:
In that case I would assume that it would be the same for any subclass of `UIScrollView` such as `UITableView` and `UICollectionView`. Its worth to try it out to see.

danielgindi:
I know that inside a UIScrollView works, haven't tested UITableView.
It should be tested to make sure that this commit handles these cases
properly

coupgar:
@petester42 This hack only happens in a `UITableView`.  `UICollectionView` does not have a WrepperView,  so it works fine with ColloctionView.

petester42:
I just tested this and the fix works for UITableView. Without it the chart does not behave like if it was in a scrollview.

danielgindi:
@petester42 :+1: to the use of `where` here :-)

danielgindi:
@coupgar could you please squash everything together?

petester42:
awesome thanks!

PR #474, opened 2015-10-14
Refactor: move minOffset to ChartViewBase and assign it with 0 in PieRadarChartViewBase

liuxuan30:
I think it's better to put minOffset to ChartViewBase. It's a fundamental property for all chart views
Just assign 0 for PieRadarChartViewBase in initialize()


danielgindi:
I almost merged this, but then noticed that we actually have different defaults for this, and also getter/setters.
We may want to reconsider this in the future, but I'm going to leave it be for now

liuxuan30:
NP

PR #488, opened 2015-10-20
Fix issue related to PhilJay/MPAndroidChart#1121

PhilJay:
Setting default skip to 1 might result in unexpected behaviour.

### DIFF
--- a/Charts/Classes/Charts/RadarChartView.swift
+++ b/Charts/Classes/Charts/RadarChartView.swift
@@ -38,7 +38,7 @@ public class RadarChartView: PieRadarChartViewBase
     public var drawWeb = true
     
     /// modulus that determines how many labels and web-lines are skipped before the next is drawn
-    private var _skipWebLineCount = 1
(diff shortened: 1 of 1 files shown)


liuxuan30:
commit 088b10d actually want it to be 1

danielgindi:
If it will be zero then it will cause an infinite loop. Should be closed

PhilJay:
Should not cause infinite loop. You already fixed this: https://github.com/danielgindi/ios-charts/blob/master/Charts/Classes/Renderers/RadarChartRenderer.swift#L182

Currently it will (per default) skip 1 web line.

danielgindi:
Oh right. This property was very confusing to begin with...

PhilJay:
Yes it was :-)
I sure had my trouble with it :-D

danielgindi:
Yeah there was a good reason why I wanted to change its behavior to not add/subtract the 1 on the property. There were many bugs and misunderstandings. I hope its stable now...

PR #493, opened 2015-10-22
if only line data exists and no other data, turn _deltaX to 1.0

liuxuan30:
if only line data exists and no other data in combined chart, turn _deltaX to 1.0, just like what LineChartView did

This is for https://github.com/danielgindi/ios-charts/issues/329 and what @jwardle asked in my PR #269


danielgindi:
Thanks! Please have your descriptions more to-the-point next time :-) It's hard to understand what the change actually does...

liuxuan30:
OK. The problem is sometimes there is only line chart data, no bubble, no bar, no candle. So the _deltaX is not correct. This fix it, taking example from LineChartView

PR #498, opened 2015-10-23
Fixed drag offset panning bug

leoMehlig:
Fixed a bug where panning was disabled in BarLineChart if a drag offset
was set

Could also be merged manually since it is only one line/“!”


danielgindi:
Thanks!

leoMehlig:
Don't know if that is indented, but if the chart is fully zoomed out the drag offset still doesn't work

PR #499, opened 2015-10-23
Drag offset fix (again)

leoMehlig:
There was a bug in the previous fix. This one should now work correctly. Sorry for that.

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -928,9 +928,10 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
         
         if (gestureRecognizer == _panGestureRecognizer)
         {
-            if (_dataNotSet || !_dragEnabled || self.hasNoDragOffset ||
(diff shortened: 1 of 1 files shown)


danielgindi:
Please first match it with the logic on the Android part

Also please fix the code styling...

danielgindi:
Ok logic seems fine, but curly brackets should always be on a new line for clarity.
Also there should only be one commit in such a PR...

```
if _dataNotSet || !_dragEnabled || 
   (self.hasNoDragOffset && self.isFullyZoomedOut && !self.isHighlightPerDragEnabled)
{
  return false
}
```

danielgindi:
The code that I've pasted here (which is actually the way you wrote it, only styled) matches the android logic, I checked. Android and iOS thinks differently about gestures, but the logics of those flags stayed the same.

**Please squash everything into one commit.**

leoMehlig:
Shall I create a new PR with all the things I've changed in one commit?

codeHatcher:
Is this fixed in v2.1.5 @danielgindi? I'm running into this issue all of a sudden and am trying to troubleshoot.

danielgindi:
No its in `master`, a version release will take place soon.

> .
>

PR #500, opened 2015-10-23
Adds test support without cocoapods

petester42:
Addresses #414 

Like #422 this does not use cocoapods as to not create an xcworkspace. For now this uses `XCTest` but  could be moved to `Quick` and `Nimble` if everyone likes it better.


danielgindi:
Looks good!
I think we should add tests for the beziers too, and add edge cases (viewport encloses beginning/end of a large dataset, instead of being fully zoomed out)
But this is a good start!

petester42:
For sure. I was just putting the infrastructure in place.

danielgindi:
Can't auto-merge, I think it clashes with the tvOS project changes

petester42:
Alright. I'll rebase and make things work.

petester42:
That should fix it

danielgindi:
Compiles fine and runs just fine :-)

yas375:
@petester42 thanks! :+1: :rocket:

danielgindi:
How do we re-generate those images?

While adding rotation support to labels, I have removed some "predefined" values that were there to adjust drawing quirks. Now we control the drawing more accurately and do not need those. But things have changed on a pixel level and all tests fail :(

petester42:
In `setup` add `record = true`. An example can be seen [here](https://github.com/facebook/ios-snapshot-test-case/blob/master/FBSnapshotTestCaseDemo/FBSnapshotTestCaseDemoTests/FBSnapshotTestCaseDemoTests.m). It's in objective-c but it's the same thing.

petester42:
Possibly all test can have it written but set to no just to make it easier to understand.

danielgindi:
Thanks :-)

PR #502, opened 2015-10-24
Update LICENSE

kiliankoe:
According to the appendix on http://www.apache.org/licenses/LICENSE-2.0 this is enough
for licensing a project under the Apache license 2.0.

I included the current year and your name by hand, please make sure this is correct before merging, thanks :)


petester42:
its not the same thing. What you are talking about is how to apply it to your code files.

kiliankoe:
Oh, I must've misunderstood. Sorry. But at least the appendix at the bottom shouldn't be necessary then, right?

petester42:
I guess but there is really no point in removing it just because.

PR #503, opened 2015-10-24
draw the hollow candlesticks in the traditional way

jfcasley:
i.e individual upper and lower wicks


danielgindi:
1. Please show what you mean with images so we can compare
2. Changes like this will only be accepted if they are optional

jfcasley:
screen shot 2015-10-30 at 6 06 42 am

jfcasley:
https://en.wikipedia.org/wiki/Candlestick_chart

jfcasley:
I also noticed that the default fill logic is backwards, but didn't bother to create a pull request for something that trivial

danielgindi:
Since there were too many changes in the repo since this PR, I just recreated it over the latest code base - but without the need for a new View and Dataset :-)

Thanks!

PR #513, opened 2015-10-27
Implemented support for rotated labels on the x-axis

danielgindi:


liuxuan30:
do you mind explain some core parts here? sort of lost after looking at various offset and size and the radians... and of cause the scene behind rotation? :) Thanks a lot!

danielgindi:
Well:

1.
In some places the charts are trying to figure out a labels width/height,
or the "longest label".
This is invalid now because the label may be rotated, so I calculate the
width/height of the box enclosing the rotated label (some trigonometry
magic)

2.
The old drawing method always took a point, drawn the text *above* it
(that's android's behavior), and to the right/left/center of it depending
on the align property.
Now a label may be rotated, and after some trial and error, I figured out
that the label should *always* be rotated around its center. So the drawing
function takes a point, plays with matrices and renders the label.

3.
Now in some cases we still want alignment.
Like drawing everything to the left/right i.e in horizontal bar's axis.
Or draw th label below/above the point, becuase we don't want the labels to
be aligned vertically and then the axis looks wierd. I tried custom offsets
but it does not offer a solution.
So now instead of aligning the drawing on the corner of the label and
rotating around it- I align it to the corner of the outer box enclosing the
rotated label. That's a trick that took some time to figure out.

4.
To specify the alignment I'm using float x/y, where 0.0 x means the label
draws to the right of the specified point, 1.0 means to the left. y=0 means
below, 1.0 means above.
Basically the anchor point is the percentage on the outer rect, at which
the anchor point resides.

‏בתאריך יום רביעי, 28 באוקטובר 2015, Xuan <notifications@github.com> כתב:

liuxuan30:
Thanks for the details!

Issue #520, opened 2015-10-30
[BUG] highlight when triggering pan gesture in default scale not working


liuxuan30:
For commit https://github.com/danielgindi/ios-charts/commit/1d53151883e924d14c59c46ec77b7a402befc93a

It turned out that below cause the problem:
```Swift
if !_dataNotSet && _dragEnabled &&
    (!self.hasNoDragOffset || !self.isFullyZoomedOut || self.isHighlightPerDragEnabled)
{
    _isDragging = true
    ...
}
```
Why `|| self.isHighlightPerDragEnabled` here? It will make the whole if condition always true and lead  _isDragging to true. then
```swift
else if (recognizer.state == UIGestureRecognizerState.Changed)
{
    if (_isDragging)
    {
        ...
    }
```
will always goes in isDragging, not triggering the highlight code path.

liuxuan30:
correct me if I misunderstood or missed anything

PR #521, opened 2015-10-30
fix highlight while panning at default scale Fix issue #520

liuxuan30:
For issue #520:
self.isHighlightPerDragEnabled is making the whole if condition true in old code, so will lead to _isDragging = true

Correct me if I missed anything or misunderstood


danielgindi:
Thanks, I did something that will clarify this and prevent us from repeating those mistakes...

PR #558, opened 2015-11-19
round the float value before we cast to Int

liuxuan30:
Old code will cause the visible index incorrect, smaller by 1.

Int() simply cuts the fractional part, which usually isn't the desired behaviour ;) Int(round(f)) does the job

I met this when that the last label should be year 2015, but it actually shows 2014. It turned out that pt.x is 3.9999999999999987, and Int(pt.x) is 3

More information about the casting can be found here:http://stackoverflow.com/questions/24029917/convert-float-to-int-in-swift


danielgindi:
Thanks :-)

Issue #357, opened 2015-09-11
Module is not extensible


iheart2code:
I do not have a suggested solution for this issue, but the Swift module is not nearly as extensible as the Java counterpart. In Java, you can create a subclass of a chart renderer and take advantage of internal/protected methods & properties to create a custom chart renderer. This does not appear to be the case with the Swift module.

I was trying to create a bar chart with rounded corners on the bars. I needed to subclass BarChartRenderer, but was unable to do so without moving the ios-charts source code into my app's module.

Is there a way to accomplish a similar level of API-parity with the Java counterpart while at the same time allowing for extensibility? Perhaps take advantage of protocols or something.

danielgindi:
"protected" methods are not available in Swift. For some reason Apple think
that it's not a use case...

Maybe a post on the appropriate Swift forum will change something...

petester42:
I would suggest maybe writing an extension to `BarChartRenderer` and override the methods you want to change. Maybe worth a try.

martinnormark:
You cannot override methods from extensions:

> Extensions can add new functionality to a type, but they cannot override existing functionality.
Source: [Docs, first note](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Extensions.html#//apple_ref/doc/uid/TP40014097-CH24-ID151)

danielgindi:
@martinnormark it is not an extension but a module...

martinnormark:
Yes, I was just replying to @petester42 's suggestion.

petester42:
There is a distinction though. If the property or function comes from the subclass then its fine. 

```
extension UIViewController {
    /// doesn't work
    override var title: String? {
        get {
            return "Cool"
        } set(new) {
            
        }
    }
}

class VC: UIViewController { }

extension VC {
    /// works
    override var title: String? {
        get {
            return "Cool"
        } set {

        }
    }
}
```

I thought it might work since `BarChartRenderer` since it is a subclass of `ChartDataRendererBase` but I haven't tried it.

petester42:
@danielgindi I believe this kind of issue could be solved if the renderers were protocol oriented and not subclasses. Then the charts could then take any object that adheres to lets say `RenderProtocol` and make it easier for other to modify the rendering.

martinnormark:
[...]

PR #567, opened 2015-11-24
Make ChartHighlighter a protocol, publicly accessible

iheart2code:
Relates to #357

### DIFF
--- a/Charts/Classes/Charts/BarChartView.swift
+++ b/Charts/Classes/Charts/BarChartView.swift
@@ -33,7 +33,7 @@ public class BarChartView: BarLineChartViewBase, BarChartDataProvider
         renderer = BarChartRenderer(dataProvider: self, animator: _animator, viewPortHandler: _viewPortHandler)
         _xAxisRenderer = ChartXAxisRendererBarChart(viewPortHandler: _viewPortHandler, xAxis: _xAxis, transformer: _leftAxisTransformer, chart: self)
         
-        _highlighter = BarChartHighlighter(chart: self)
+        self.highlighter = BarChartHighlighter(chart: self)
         
         _chartXMin = -0.5
     }
@@ -69,7 +69,7 @@ public class BarChartView: BarLineChartViewBase, BarChartDataProvider
             return nil
         }
         
-        return _highlighter?.getHighlight(x: Double(pt.x), y: Double(pt.y))
+        return self.highlighter?.getHighlight(x: Double(pt.x), y: Double(pt.y))
     }
         
     /// - returns: the bounding box of the specified Entry in the specified DataSet. Returns null if the Entry could not be found in the charts data.
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -106,7 +106,7 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
         
         _xAxisRenderer = ChartXAxisRenderer(viewPortHandler: _viewPortHandler, xAxis: _xAxis, transformer: _leftAxisTransformer)
         
-        _highlighter = DefaultChartHighlighter(chart: self)
+        self.highlighter = DefaultChartHighlighter(chart: self)
         
         _tapGestureRecognizer = UITapGestureRecognizer(target: self, action: Selector("tapGestureRecognized:"))
         _doubleTapGestureRecognizer = UITapGestureRecognizer(target: self, action: Selector("doubleTapGestureRecognized:"))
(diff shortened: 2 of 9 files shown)


petester42:
Pretty sweet stuff man :+1:. Could this be used from any graph type and not just BarLineChartViewBase?

iheart2code:
Yes, it can. Although I had some difficulty figuring out how to get a downcast version of the renderer into some of the classes. See the TODO here: https://github.com/danielgindi/ios-charts/pull/567/files#diff-ac32fb888c389b4e17eca7e5ac42579dR20

petester42:
Something like this could possibly work
```
public protocol ChartHighlighter {
    public typealias ChartViewType
    weak var chart: ChartViewType? { get set }
}
```

danielgindi:
I'm going to hold this a bit, as Phil and I are working on some changes that make the library more interface-oriented, mainly to allow querying for data (instead of pushing an array of entries).
We're trying to keep it 100% backwards compatible, but things could happen.

Anyway, I'll try to add this into the mix before we merge to master!

iheart2code:
No worries. Good luck with your work. I'm excited to see what you've got in store.

danielgindi:
Don't you think that the use of an extension in this case is a little bit problematic?

I mean if we want to have an interface with some base functionality implemented, we should probably have an "abstract" base class...

What will happen if an implementation has a function with the same name as one implemented in the extension?

danielgindi:
@iheart2code  Actually I don't get why not make the original class' properties `public`..?

petester42:
defining a protocol and then defining an extension of that protocol is the same as an abstract base class. Like can be seen below.

```swift 
protocol CoolType {
    func superFunction() -> String
}

extension CoolType {
    func superFunction() -> String {
        return "blue"
    }
}

struct AType: CoolType {
    
}

struct BType: CoolType {
    
    func superFunction() -> String {
        return "red"
    }
}

let a = AType().superFunction()
let b = BType().superFunction()
```

The output looks like so:
screen shot 2016-01-10 at 1 10 10 am

danielgindi:
@petester42 thanks for explaining!

Is there an advantage of having this as an abstract class and not just making those properties `public`?
Because as I see it, if the class conforms to a protocol, then its methods are public anyway as those are protocol methods. Which means that just making the properties `public` in the first place will achieve the same desired effect.

danielgindi:
I just want to be finished with this one correctly on the iOS and Android side, before merging in `dynamic-preparations` branch which introduces interfaces between current dataset classes and their implementations, and easy Realm.io integration.

petester42:
The advantage is that you have a flatter hierarchy instead of more nested one with subclass of a subclass etc.

You can also set the access control to protocol methods like so:

```swift
public struct BType: CoolType {
    
    internal func superFunction() -> String {
        return "red"
    }
}
```

that function would only be public inside the library which for what this is trying to accomplish probably doesn't make sense but you can do it.

danielgindi:
Remember that the library is also used from ObjC, so if anyone tries to call a protocol method from ObjC, it will let him...

Also that protocol should actually be exposed to ObjC, which is not the case right now.

Anyway, I still don't see why we need the additional hierarchy if we just want to make things public!

petester42:
The point would be that a user would not have to subclass something from the library in order set their own chart highlighter. Subclassing add a level of complexity since you never know if you need to call super or not (unless the function is properly annotated). Having it be a protocol makes it more flexible for a user since it is more straight forward. Implement these methods and it will work.

danielgindi:
Well in such a case we would have to rename the protocol methods to specify the highlighter context, as one can (and often do) implement multiple protocols in one class.

@PhilJay what do you say about this? Basically this means making `ChartHighlighter` an interface, moving the implementation to `DefaultChartHighlighter`.

The thing that I mostly don't like about this is that this would mean a major inconsistency API with the Android side.
As an "extension" is something that only iOS allows, and so all current Highlighters would have to subclass `DefaultChartHighlighter`, while on the iOS side they just implement the base interface and get the functionality by default.
The iOS implementation is like a combination of an interface and an abstract class...

@petester42 I now get why these changes were implemented in the PR, but as you probably know the biggest strength of the framework is being cross-platform :-) We just need to carefully consider this before implementing.

petester42:
For sure, from the issues I know a bunch of people use this with objc and of course the API consistency with android is a priority.

Without having actually gone through it I would say that this should probably be rejected based on the things you mentioned. I don't know about protocol extensions in objc but I would assume they don't work.

danielgindi:
@petester42 Oh no, I haven't event thought about protocol extensions in ObjC... No they don't work. That's an issue on iOS even before considering Android.

So sadly this PR is going to be rejected... @iheart2code do you want to modify the PR to just make things public, make another PR, or just leave that to us?

PR #571, opened 2015-11-25
Make getBarBounds callable from Objective-C code (Fixes #570)

### LINKED ISSUE #570: Cannot call getBarBounds from objective-c
This is not a bug but it is a problem for me because I cannot get the size of the bars in any other way.

The problem at hand is the impossibility of calling this one method because the swift implementation returns nil when no data set is found. Since CGRect in obj-c is not an object and cannot be set to nil, the method is not callable.

There is an easy workaround for this which is returning CGRectNull instead of nil. I am creating a pull request for this.

This behaviour is there since the first implementation of the method which is in this commit: 19165ddacb2bbac9db0a541fd92f86b294147c11

adonaire-sage:
Relates to #570.


danielgindi:
Seems fine to me. :-)

PR #575, opened 2015-11-27
add the ability to choose whether to clip the chart when rendering

liuxuan30:
Kind of feeling we will need the ability to choose whether to clip the contentRect.

Currently we will clip the contentRect, but sometimes, if the user customize the highlight shape, or have bigger shape close to the boundary, it will be cut.

The default is to clip as current implementation.


liuxuan30:
hmm, seems if we con't clip drawData and drawHighlighted, the chart will be buggy when zooming/scrolling.
It appears not as simple as it looks like. Closing until I found a better way or people are asking for it.

PR #577, opened 2015-11-27
Fixes a crash caused by NaNs in CombinedChartView with empty data set

natethegreat44:
Fixes a crash caused by NaNs in CombinedChartView when an empty dataset is passed in.

Also fixes a simlar NaN-related crash when such a chart is touched.

A minimal view controller to reproduce is:

```
import UIKit
import Charts

class InformationViewController: UIViewController {
    @IBOutlet weak var chartView: CombinedChartView!

    override func viewDidLoad() {
        let xAxisData = [NSDate]()
        let bar1Data = BarChartDataSet(yVals: [BarChartDataEntry]())
        let bar2Data = BarChartDataSet(yVals: [BarChartDataEntry]())

        chartView.xAxis.labelPosition = .Bottom
        chartView.descriptionText = ""
        chartView.animate(xAxisDuration: 0.5, yAxisDuration: 1.0)

        let combinedData = CombinedChartData(xVals: xAxisData)
        combinedData.barData = BarChartData(xVals: xAxisData, dataSets: [bar1Data, bar2Data])

        chartView.data = combinedData
    }
}
```

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -1716,6 +1716,9 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
     {
         var pt = CGPoint(x: viewPortHandler.contentRight, y: viewPortHandler.contentBottom)
         getTransformer(.Left).pixelToValue(&pt)
+        
+        if (isnan(pt.x)) { return 0 }
+        
         return (_data != nil && Int(round(pt.x)) >= _data.xValCount) ? _data.xValCount - 1 : Int(round(pt.x))
     }
 }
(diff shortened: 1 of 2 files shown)


liuxuan30:
hold on, this seems a regression.
Missing data entries should be taken care of before we start rendering. There is a check when setting data:
```swift
        set
        {
            if newValue == nil
            {
                print("Charts: data argument is nil on setData()", terminator: "\n")
                return
            }
            
            _dataNotSet = false
            _offsetsCalculated = false
            _data = newValue
            
            // calculate how many digits are needed
            calculateFormatter(min: _data.getYMin(), max: _data.getYMax())
            
            notifyDataSetChanged()
        }
```
I checked the commits, @danielgindi delete one condition:
```swift
if (newValue == nil || newValue?.yValCount == 0)
```
in `drawRect`, if we detect dataNotSet is true,  we won't continue the drawing at all.

@danielgindi, seems we cannot delete newValue?.yValCount == 0

BTW, if it's nan, returning 0 is not a good choice here I think.

danielgindi:
Maybe keep the older condition - but only `print` is the `newValue` is `nil`? :-)
I think this would satisfy both issues...

liuxuan30:
I will close this PR and reverse 062a61b in PR #582
@natethegreat44  thanks for pointing out this issue!

PR #580, opened 2015-11-29
Bar chart tests

alvesjtiago:
Bar chart tests included mimicking previous line chart tests. Updated
ios-snapshot-test-case to 2.0.7 version.
Bar chart tests include:
- default chart
- hide values
- hide left axis
- hide right axis
- hide left axis gridlines
- hide right axis gridlines


Charts/ChartsTests/BarChartTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        dataSet = BarChartDataSet(yVals: entries, label: "Bar chart unit test data")
+        
+        chart = BarChartView(frame: CGRectMake(0, 0, 480, 350))
+        chart.data = BarChartData(xVals: xValues, dataSet: dataSet)
+    }
+    
+    override func tearDown()
+    {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
```
yas375: Nit picking: no need to override `tearDown` here to just call super. Boiler plate code probably ;)

yas375:
Cool! :+1:

danielgindi:
Thanks :-)

Charts/ChartsTests/BarChartTests.swift
alvesjtiago: You're right @yas375, I just reused the code from the LineChartTests but I will refactor it to a single data generation file and keep the specifics to each of the tests. As well as removing the ```tearDown``` override :) thank you

PR #586, opened 2015-12-03
Refactor charts tests

alvesjtiago:
Create base class for charts with:
- shared imports
- default values setup
- tearDown function


yas375:
Personally I would prefer to avoid subclassing in specs unless it's really needed. There are not so many code duplication in these specs. And for specs theirs readability is more important. Tests can be examples of how to use a library. Going through superclasses to see how the data is created seem to be more complicated.

If you'd like to reuse how you create a test data - you can add a helper method or a function and use it in multiple specs. Using different sets of data in different specs is not a bad idea too.

petester42:
I completely agree with @yas375

alvesjtiago:
Got it. Thank you for the feedback.

PR #614, opened 2015-12-15
Remove useless parentheses causing swift build error

chanil1218:
Error : ChartAnimationEasing.swift:345:16: Expression was too complex to be solved in reasonable time; consider breaking up the expression into distinct sub-expressions


petester42:
Thanks. Looks fine

PR #624, opened 2015-12-22
Bar and Pie Chart Icons [Dup #629?]

zmchristensen:
Bar and Pie Charts now support icons.

Added ability to draw icons on Bar and Pie charts when an image name is specified in ChartDataEntry.data. Added Boolean flags 'isDrawIconEnabled`and`isDrawIconAboveBarEnabled` (similar to the existing drawValues logic). Icon size is dependent on bar width or size of pie slice.

PR #629, opened 2015-12-28
LineChart: drawing image on nodal points [Dup #624?]

AnatolyRosencrantz:
Sometimes circle in the nodal point is not enough customisation. Here implemented add-on to LineChart renderer which allows to draw custom UIImage instead of (or on top of) circle.

For using: switch DataSet.drawNodalImageEnabled=true, set image by DataSet.nodalImage=UIImage() and that's it. See demo project also.


danielgindi:
It seems like both you and @zmchristensen are trying to implement icons, for different charts.
But the code differs a lot - and we would want a unified way of doing that across all charts.
Is there a chance that you guys will do some more work on merging those two?

zmchristensen:
@danielgindi: between the two implementations, do you have a preference for how icon are drawn?

danielgindi:
Not really- it just needs to be consistent as much as it can be with other APIs in the framework

AnatolyRosencrantz:
@zmchristensen finally I've got time to fix this issue. Which of your branches can I merge my changes with? master or chart-icons?

zmchristensen:
The chart-icons branch will be best.

abjurato:
Feature of this PR is included in #1261

Issue #138, opened 2015-06-10
Drawing circles for values on the radar chart


liuxuan30:
Say it is a radar chart view,  can ios-charts draw a dot circle at each data point (a specific yValue and xIndex)?

danielgindi:
Can you provide an example?

liuxuan30:
I capture this chart from a JS radar chart. For each data point, it has a doct circle. When highlighting, the dot circle can be hollow (filled by white color, stroke with the old color)

Current ios-chart radar chart don't have a dot. It not easy to see if there is a data point. Having some dots will make it easier.

liuxuan30:
![radar](https://cloud.githubusercontent.com/assets/4375169/8105647/7c054796-106d-11e5-8259-6332cc47ca97.png)

I capture this chart from a JS radar chart. For each data point, it has a doct circle. When highlighting, the dot circle can be hollow (filled by white color, stroke with the old color)

Current ios-chart radar chart don't have a dot. It not easy to see if there is a data point. Having some dots will make it easier.

liuxuan30:
@danielgindi Do you think this is worthy to have? I am about to finish implementing this feature on my side.

liuxuan30:
Seems only I have this requirement. I will close it. If anyone raise same request, i could share my code.

alexwillrock:
this is a very cool feature, it is strange that the developers did not include it by default, really need this feature
thx

liuxuan30:
@alexwillrock you are the first one to ask for a dot since Jun..

You could add the dots just like drawCircles in line chart renderer into Radar chart renderer.

shrikantwalekar-tudip:
@liuxuan30, I need the same feature you are talking about. would you please file PR.

liuxuan30:
I actually did it already, PR is right above your comment.

shrikantwalekar-tudip:
@liuxuan30 , I did that . and able to show circle on tap of point.
I want my chart to look like the image above.

but not able to plot single point or points like line chart, in case- if there are multiple radar Chart Data sets and if any data set contains only one point then that point is not visible. Also the circles are visible only after tap. Please guide me if I am doing something wrong.
![screen shot 2016-02-03 at 8 57 07 pm](https://cloud.githubusercontent.com/assets/14139567/12787220/72916214-caba-11e5-8acb-9d06e0c56e9b.png)

liuxuan30:
[...]

PR #630, opened 2015-12-29
add highlight circle for radar chart

liuxuan30:
For #138 
Add highlight circle for radar chart. It combined an inner filled circle and an outer stroke circle. inner circle fill color, inner circle radius, outer circle radius, stroke width and alpha can be customised.

I notice `drawHighlightLines` is in `LineScatterCandleRadarChartRenderer` now, no longer in the radar chart renderer. I am not sure if draw highlight circle should be here.

Since there are several properties can be asked to customized, I create new properties for radar chart view. 

Let me know if we want to add more customizable properties or it's too many, as well as if the structure needs to be improved.

### DIFF
--- a/Charts/Classes/Charts/RadarChartView.swift
+++ b/Charts/Classes/Charts/RadarChartView.swift
@@ -37,6 +37,19 @@ public class RadarChartView: PieRadarChartViewBase
     /// flag indicating if the web lines should be drawn or not
     public var drawWeb = true
     
+    /// flag indicating if highlight circle should be drawn or not
+    public var drawHighlightCircle = true
+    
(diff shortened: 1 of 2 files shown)


preetirani:
Hey @liuxuan30 
i want to draw circles in radar chart for the selected dataset. Can you please help me to understand what all properties needs to b implemented on radarchatview.
Here are the properties I have set but circles are not showing at all.
set1.drawHighlightCircleEnabled = true
            set1.highlightCircleStrokeAlpha = 1
            set1.highlightCircleFillColor = UIColor.green
            set1.highlightCircleOuterRadius = 10
            set1.highlightCircleInnerRadius = 8
            set1.highlightCircleStrokeColor = UIColor.white

PR #632, opened 2015-12-29
Use nil coalescing in ChartDataSet's entryCount (Fixes #631)

### LINKED ISSUE #631: ChartDataSet's entryCount computed property raises a signal when _yVals is nil
To reproduce:

```swift
let dataSet = ChartDataSet()
dataSet.entryCount
```

__Expected Behaviour__
entryCount returns 0

__Current Behaviour__
entryCount raises a signal

aarondaub:
The force unwrapping in entryCount caused a crash when yVals was nil. Now entryCount returns 0.

#631

### DIFF
--- a/Charts/Classes/Data/ChartDataSet.swift
+++ b/Charts/Classes/Data/ChartDataSet.swift
@@ -153,7 +153,7 @@ public class ChartDataSet: NSObject
         return yValueSum / Double(valueCount)
     }
     
-    public var entryCount: Int { return _yVals!.count; }
+    public var entryCount: Int { return _yVals?.count ?? 0 }
     
     public func yValForXIndex(x: Int) -> Double
     {


liuxuan30:
I'm not sure if we should initialize yVals in init()?

```swift
    public override required init()
    {
        super.init()
    }
```
It does not initialize yVals, so it will crash. Not sure which way is better.

BTW, _yVals?.count is Optional, while entryCount is Int type, though seems swift will help unwrap it. Not sure if we should make the return type consistent

aarondaub:
@liuxuan30 Having that initializer keep properties in the same state as other initializers would make sense.

I'm not sure I follow what you mean about making the return type consistent – making entryCount Int? requires changes throughout the rest of the codebase but having it fail safely (by using ??) only requires changing the getter itself.

liuxuan30:
@aarondaub, I am no expert of the details of swift, so what I want to ask is, _yVals?.count will return Int? instead of Int, while entryCount is declared as Int type. Let's say the yVals has 5 entries, so it will return `Optional(5)`. I tried the code, and swift will help unwrap, and entryCount is assigned as `5`. 
I am just not sure, if this is the right way to do so, or we should follow the style in documentation:
```swift
if let count = _yVals?.count 
{
   ...
} else {
   ...
}

**Comment by aarondaub** [comment]
@liuxuan30 ah, let me clarify.

You're right that

```swift
yVals?.count
```

is of type `Int?` so it is not a valid return type for a function (in this case entryCount) that returns `Int`. In fact it wouldn't compile if the implementation was just:

```swift
public var entryCount: Int { return _yVals? }
```

To get around this, I used the nil coalescing operator `??`. What `??` does is unwraps its left hand argument and return that if it __is not nil__, otherwise it returns its right hand argument.

So `nil ?? 0` is 0, and `Optional(1) ?? 0` is 1.

As far as style goes: I personally think the nil coalescing operator is fine, adding an explicit if-let/else would just be unnecessary bulk IMHO.

danielgindi:
Seems fine to me :-) Thanks!

PR #635, opened 2015-12-29
Add missing images for bar chart tests

petester42:
I noticed that there was missing images for 64 bit @3x and 32 bit @2x causing the tests to fail when I would run them.


yas375:
From looking at generated images `testHideLeftAxis`, `testHideLeftAxisGridlines`, `testHideRightAxis`, `testHideRightAxisGridlines` I wonder if those tests are actually work as expected? Those images seem to be equivalent to me while I was expecting some changes on them. Or am I missing some details?

petester42:
No they are correct it is just the names that are confusing. `testHideLeftAxis` for example just hides the line that borders the graph. I'll update the tests to make more logical sense.

yas375:
OIC. Thanks!

PR #639, opened 2015-12-31
Remove verbose semicolons

AntiMoron:


Charts/Classes/Utils/ChartViewPortHandler.swift
```diff
@@ -158,9 +158,13 @@ public class ChartViewPortHandler: NSObject
         return CGPoint(x: _contentRect.origin.x + _contentRect.size.width / 2.0, y: _contentRect.origin.y + _contentRect.size.height / 2.0)
     }
     
-    public var chartHeight: CGFloat { return _chartHeight; }
+    public var chartHeight: CGFloat {
```
liuxuan30: {} at same line or follow
{
...
{
maybe?

AntiMoron: {} at same line then semicolons would be needed. That's far from swift style.

I just mistakenly followed my own code style. I've corrected that.

PR #648, opened 2016-01-04
Stop axis labels duplicating. #315  [Pending squash and cleanup]

noais:
With one commit :+1:

### DIFF
--- a/Charts/Classes/Components/ChartYAxis.swift
+++ b/Charts/Classes/Components/ChartYAxis.swift
@@ -103,6 +103,14 @@ public class ChartYAxis: ChartAxisBase
     /// **default**: 0.0 (no maximum specified)
     public var maxWidth = CGFloat(0)
     
+    /// if true, axis values will be repeated
+    public var allowRepeatedValues = false
+    
+    /// the minimum diference value of the axis values
+    ///
+    /// **default**: 1
+    public var granuality = Double(1.0)
+    
     public override init()
     {
         super.init()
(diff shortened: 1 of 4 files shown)


liuxuan30:
I think you need to practice "squash into 1 commit" by creating test branch and simulate the normal cases. Like you have create several commits, and merge upstream master into your test branch, and squash it into one based on the upstream master. It's too much burden if you manually copy paste the code.

If you are using SourceTree, you could right click the latest commit, Rebase, and let it rebase the code on the latest commit.
In order to squash, you need to right click the commit before your first commit, and Rebase Interactively. There you have a squash option :-)

noais:
The problem was my master branch wasn't correctly synchronised with the origin and I thought it was.
Thanks

liuxuan30:
you could make sure you don't touch master, just keep it sync with upstream

petester42:
Are there any cases where you would want values to be duplicated? I get why the code duplicates values but users usually would assume to not have repeating values like a real graph.

liuxuan30:
I guess it's just a switch to not triggering the code?

noais:
@petester42 people was asking to create allowRepeatedValues. So I created, but I can't see one example where I want to see repeated values.

danielgindi:
@noais thank you!
I've adjusted the terminology for the better, and made this the new default behaviour.

@petester42 a case where people will want "repeating values" is for example when you want to show grid lines - while hiding the value labels - but you want to always have at least 5 grid lines on display. No matter how fractioned the value intervals are...

PR #658, opened 2016-01-08
add support for lineCap setting for line chart

liuxuan30:
If user does not draw circles for line chart, the connecting dot seems not very smooth because of default lineCap.Butt
Add support for setting lineCap in this PR. I am not sure if we should put lineCapType into BarLineChartViewBase, currently combined chart and line chart will both have it.

Another thing is when I tried to port the code to drawCubic, I don't see it take effect before/after `CGPathAddCurveToPoint`, and cubic line seems smooth already, so I don't port it at last. Correct me if I am wrong.

### DIFF
--- a/Charts/Classes/Charts/CombinedChartView.swift
+++ b/Charts/Classes/Charts/CombinedChartView.swift
@@ -20,6 +20,8 @@ public class CombinedChartView: BarLineChartViewBase, LineChartDataProvider, Bar
     /// the fill-formatter used for determining the position of the fill-line
     internal var _fillFormatter: ChartFillFormatter!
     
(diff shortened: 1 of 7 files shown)


danielgindi:
It seems like you've introduced duplicate properties...
Also properties of the line I think should be in the data and not in the provider - so you don't have to look around to find line properties scattered across the library :-)

liuxuan30:
For duplicated properties you mean in combined chart and line chart both have `lineCapType`?
putting into data class seems easier and can solve the dup, I did think about it, but at that time what I think is something like line property may not be proper to be included in data class.

liuxuan30:
update to latest interfaces. putting lineCapType into `ILineChartDataSet`

PR #660, opened 2016-01-09
add an option to set line cap of axis grid line

mconintet:
add an option to set line cap of axis grid line


danielgindi:
This is a nice feature - although unfortunately it needs rebasing as there are conflicts...

PR #670, opened 2016-01-12
Added a way to change circle hole line width on Line charts.

terloon:
I didn't see any way to modify the width of the circles in the line chart. This is my initial attempt at it. The current method performs a calculation halving the circle diameter. Right now I hard coded it to be the same value for the default radius, but it will look different for current implementations if they use this. 

Needs invalid value checks and maybe a calculation to match previous look of halving diameter.

### DIFF
--- a/Charts/Classes/Data/LineChartDataSet.swift
+++ b/Charts/Classes/Data/LineChartDataSet.swift
@@ -20,6 +20,7 @@ public class LineChartDataSet: LineRadarChartDataSet
     public var circleColors = [UIColor]()
     public var circleHoleColor = UIColor.whiteColor()
     public var circleRadius = CGFloat(8.0)
+    public var circleLineWidth = CGFloat(4.0)
     
     private var _cubicIntensity = CGFloat(0.2)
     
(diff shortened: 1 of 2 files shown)


liuxuan30:
I don't see the reason here: circleRadius is already in dataSet and can be changed by user;
Adding `circleLineWidth` and a simple subtraction does not make sense to me, because you can simply calculate circleRadius to what you need to to solve the problem.

liuxuan30:
Oops, I wrongly clicked close;

terloon:
That would definitely get rid of the subtraction. I did initially implement it that way, but felt that most people are used to defining the width of a line or border and made it more intuitive as to what it would do to the appearance of the circle.

petester42:
When making a change like this make sure the tests still pass. Other than that good job.

terloon:
Haven't really done tests before. I looked at other peoples changes (specifically pull 660) which are similar to mine and I can't see where they made a change to a "test" swift file. When I try to run the tests I get an error saying "No module FBSnapshotTestCase". Will look into this further.

petester42:
Right now you would have to run "carthage bootstrap" to instal it. I'll look into making the process easier and documenting it.

terloon:
I see the issue, thanks. Now to figure out the best way to resolve this.

danielgindi:
It doesn't look like the commits in this PR do anything actually.

PR #679, opened 2016-01-15
remove old delegate methods in CombinedChartRenderer

liuxuan30:
remove old delegate methods in CombinedChartRenderer, since they are converted to data provider delegate methods, and the delegate is chartView itself rather than renderer

I assume they are forgot to be deleted. Remind me if I missed any


liuxuan30:
close as https://github.com/danielgindi/ios-charts/commit/646d9c3907cc430576e27c533534e7d18e5760f2 did the job.

PR #686, opened 2016-01-20
add an option to hide/show legends

mstfy:
This pull request adds `drawLegends` property to `ChartViewBase` class. When it is set to `false` chart view will not display legends. The default value is `true`.


liuxuan30:
first of all, you should not touch Realm

PR #690, opened 2016-01-21
Add Bubble Chart Demo project

kktam:
I have created a simple Swift 2 demo project, which mirror the original Objective-C project. This can serve as an example of how to use the library in Swift.


petester42:
neat but the original idea of the objective-c demo project was to show how it can be integrated with objective-c and also provides a way to check that we are not using any swift only stuff.

liuxuan30:
Another issue is I suppose only a few is needed for a bubble chart, but it has 59
if this is getting merged, it should be cleaned for unecessary files

danielgindi:
Well, this is a nice effort, and I'd hate to discourage people from contributing...
But there are quite a few problems with this:
1. There's already ChartsDemo - there's no actual need for Swift demos as the framework stays the same, and the presence of a semicolon `;` at the end of a line should not prevent people from reading the code.
2. The demo you have supplied is for some reason centered about "ratings"?
3. The demo crashes when hitting most items in the list.

petester42:
Rather than a whole project why not write some docs about creating a simple chart using objective-c followed by the same code using swift. Kind of like the apple docs.

danielgindi:
I'd love it if someone had the time to write proper and beautiful docs like Apple's. That's the best situation when you can choose in which language you are comfortable to see the sample code...

petester42:
This does pretty much that. https://github.com/realm/jazzy

liuxuan30:
@petester42 that's really cool...

danielgindi:
But will it take code samples in Swift/ObjC and generate the equivalent for the other language?

petester42:
I don't know I have never used it myself but I've only heard good things.

Issue #561, opened 2015-11-20
Recalculate space on label formatting.


lidgardo:
Hi. I found a little issue which i have been struggling a lot, it seems that if i apply a format on the x-labels the API doesn't recalculate the space in the wit the new format. 

I'm adding a formatter to add an elipsis (...) when the label is more than 7 characters length just like you can see in the image.

<img width="328" alt="screen shot 2015-11-20 at 5 46 00 pm" src="https://cloud.githubusercontent.com/assets/5713880/11314790/a29a4f3a-8fae-11e5-93fa-4d4bf2702170.png">

All the gray area is the barChartView, If i don't use the formatter i got this 

<img width="329" alt="screen shot 2015-11-20 at 5 45 28 pm" src="https://cloud.githubusercontent.com/assets/5713880/11314809/ca73952a-8fae-11e5-84cc-69ba79b50693.png">

The code that i'm using to add the formatter is the following: 

Container class:

```objective-c
ChartXAxis *xAxis = barView.xAxis;
LargeValueFormatter *formatter = [LargeValueFormatter new];
xAxis.valueFormatter = formatter;


```

Formatter class:

```swift
    public func stringForXValue(index: Int, original: String, viewPortHandler: ChartViewPortHandler) -> String {
        
        if original.characters.count > 7 {
            return self.addEllipsis(value: original)
        } else {
            return original
        }
    }
```

As you can see, without the formatter the chart  use in a better way the space, i guess that the problem is that API don't recalculate the bottomContent once that the labels are formatted. I'd like to help with this if you guys can give a little explanations if my diagnosis of the issue is ok and where i can find the code that i need to call to recalculate with the new labels. I guess this could be a great feature.

Thanks in advance. 🤓

liuxuan30:
I think this is related to https://github.com/danielgindi/ios-charts/pull/513. This is how the label is rotated. In the PR there is some explainations how it work.

lidgardo:
The alignment is ok, the problem is that i think that the chart needs to be bigger in the first image, this because the labels are smaller than in the second image just to take advantage of the extra space that we gain adding the ellipsis at the labels.

liuxuan30:
I may understand your point, but still It's a little hard to see the difference of sizes. Are you talking about the height? 

Could you please check the chartView's frame size and the viewPortHandler.contentRect? There is some different size for the whole UIView and the area to draw the chart itself, excluding the axis and axis labels.

liuxuan30:
[...]

PR #703, opened 2016-01-23
Fix issue #561 - Labeling formatting 👾🤓

chuynadamas:
I just added a new formatter to the *X-Axis* just like the Y ones to support different kind of custom formatter and not only the predefine ones, also asking always the formatter the values before calculating the space by ViewPortHandler.


liuxuan30:
what happened..

PR #704, opened 2016-01-23
Enhancement - Added the parameter eventParentEnabled to Pie and Radar Charts-  🤓 👻

lidgardo:
Hi @PhilJay and @danielgindi and congrats for all the work, this is my little contribution.

**Disabling the rotate on the Pie and Radar to using that gesture for something different**

In my application I have to use the horizontal scroll to show a menu on a view which has a pie o radar chart, I can disable the gestures on the rest of the charts but in the case of Pie or Radar, I can't. o I added the attribute just to know if the user wants to the chart rotate or use the scroll for anything else.

Somethig like this:

screen shot 2016-01-22 at 6 39 54 pm

If I turn down the even parent handler, I can use the drag inside the chart to move my view and show the menu. 😎

### DIFF
--- a/Charts/Classes/Charts/PieRadarChartViewBase.swift
+++ b/Charts/Classes/Charts/PieRadarChartViewBase.swift
@@ -27,6 +27,9 @@ public class PieRadarChartViewBase: ChartViewBase
     /// flag that indicates if rotation is enabled or not
     public var rotationEnabled = true
     
+    /// flat that indicates if the the events will submit to the parent
+    public var eventParentEnabled =  true;
+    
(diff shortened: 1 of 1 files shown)


danielgindi:
I'm not sure this is necessary. You can avoid conflicts by setting `rotationWithTwoFingers=true`. Did you try that?

danielgindi:
I'm closing this.
This one is specific for rotation - while you can enable two-finger rotation.
For generally disabling events, you can just set `enableUserInteraction=false`.
So I don't really see a point in this

Issue #711, opened 2016-01-26
Cubic line charts don't get filled correctly if data doesn't start at index 0


gunterhager:
If you set up your data entries for the cubic line chart so that they don't start at x-index 0, the fill is still drawn from x-index 0.

Set up data entries like this:
```
[[ChartDataEntry alloc] initWithValue:val xIndex:i + offset]
```

The chart will look like this:
![screen shot 2016-01-26 at 09 21 01](https://cloud.githubusercontent.com/assets/3897424/12575882/60d4f62c-c410-11e5-89b1-3a706e3df9d2.png)

This happens only with cubic line charts. Regular line charts are rendered like this:
![screen shot 2016-01-26 at 09 22 06](https://cloud.githubusercontent.com/assets/3897424/12575911/8796063e-c410-11e5-92dc-397e245870eb.png)

You can easily reproduce the bug by replacing the `setDataCount:range:` method in `CubicLineChartViewController.m` in the demo with:

```
- (void)setDataCount:(int)count range:(double)range
{
    NSMutableArray *xVals = [[NSMutableArray alloc] init];
    
    for (int i = 0; i < count; i++)
    {
        [xVals addObject:[@(i + 1990) stringValue]];
    }
    
    NSMutableArray *yVals1 = [[NSMutableArray alloc] init];
    
    int offset = 10;
    
    for (int i = 0; i < count - offset; i++)
    {
        double mult = (range + 1);
        double val = (double) (arc4random_uniform(mult)) + 20;
        [yVals1 addObject:[[ChartDataEntry alloc] initWithValue:val xIndex:i + offset]];
    }
    
    LineChartDataSet *set1 = [[LineChartDataSet alloc] initWithYVals:yVals1 label:@"DataSet 1"];
    set1.drawCubicEnabled = YES;
    set1.cubicIntensity = 0.2;
    set1.drawCirclesEnabled = NO;
    set1.lineWidth = 1.8;
    set1.circleRadius = 4.0;
    [set1 setCircleColor:UIColor.whiteColor];
    set1.highlightColor = [UIColor colorWithRed:244/255.f green:117/255.f blue:117/255.f alpha:1.f];
    [set1 setColor:UIColor.whiteColor];
    set1.fillColor = UIColor.whiteColor;
    set1.fillAlpha = 0.5f;
    set1.drawFilledEnabled = YES;
    set1.drawHorizontalHighlightIndicatorEnabled = NO;
[...]

PR #712, opened 2016-01-26
Fix for cubic line chart fill when charts that don't start at x-index 0 #711

gunterhager:
Fix for issue #711

Cubic line charts that don’t start at x-index 0 don’t fill correctly
because the fill starts at index 0 (Regular line charts don’t have that
issue). To correct this the index for the fill is taken from the data
entries (similar to regular line chart fill).
Ultimately i think the renderer should be refactored to use a common
code path for filling cubic and regular line charts.


danielgindi:
Thank you!
I'll take a look at it soon.

About the refactoring suggested - you are right, although I'm not sure how easy this would be... There are more complex issues to solve about cubic paths, like clipping them to straight lines when going below minimum axis...

gunterhager:
I understand - yes it might be more complex than i thought. I just felt a little uneasy fixing this issue the way i did. I didn't like the smell of code duplication.

Issue #682, opened 2016-01-18
Center text vertically


hiworld75018:
When using big fonts, I noticed that text is not actually vertically centered.
Adding this in DrawUtils.drawText() makes it work:

if let font = attributes?[NSFontAttributeName] as? UIFont {
  point.y -= font.lineHeight / 2
}

liuxuan30:
could you specify which font size are you using, also is it axis label text or value text? A screenshot is even better.

hiworld75018:
In case there is a single value, I use the chartView.noDataText field to display it, using font size 36 (with chartView.data == nil).

liuxuan30:
a single value, but no data? Can you post a screenshot?

hiworld75018:
You may try this code, to display the "single" value converted into a string that will sit in the middle of the "chart":

UIFontDescriptor* bodyDescriptor = [UIFontDescriptor preferredFontDescriptorWithTextStyle:UIFontTextStyleBody];
UIFont* font = [UIFont fontWithDescriptor:bodyDescriptor size:36.]; // some BIG size

PieChartView* chartView = [[PieChartView alloc] initWithFrame:CGRectMake(0, 0, 300, 200)];

chartView.infoTextColor = [UIColor redColor];
chartView.infoFont = font;

chartView.noDataText = @"99%"; // the "single" value is converted into text here

// Remove the chart legend + label "description"
chartView.legend.enabled = NO;
chartView.descriptionText = @"";

liuxuan30:
if your "single value" means `noDataText`, on my side I have no problem with it:
![pie](https://cloud.githubusercontent.com/assets/4375169/12601175/f01b619e-c4d9-11e5-9410-5552a9c8d074.png)

liuxuan30:
OK I checked the code:

```swift
    public class func drawText(context context: CGContext, text: String, var point: CGPoint, align: NSTextAlignment, attributes: [String : AnyObject]?)
    {
        if (align == .Center)
        {
            point.x -= text.sizeWithAttributes(attributes).width / 2.0
        }
```
It seems it just tried to center x but no y. I will file a PR later.

hiworld75018:
I think I've got a better way to fix it:

In ChartViewBase.drawRect(rect), use lineHeight to center vertically before calling ChartUtils.drawText()

CGContextSaveGState(context)
let lineHeight = infoFont.lineHeight
[...]

PR #715, opened 2016-01-27
[Wrong?] for #682: adjust point.y to count text height, in order to draw it in center

liuxuan30:
for #682: adjust point.y to count text height, in order to draw it in center

Seems we forgot to calculate y or by purpose ? @danielgindi 

I see 
```swift
ChartUtils.drawText(context: context, text: noDataText, point: CGPoint(x: frame.width / 2.0, y: frame.height / 2.0), align: .Center, attributes: [NSFontAttributeName: infoFont, NSForegroundColorAttributeName: infoTextColor])
```
get the center point, but later it just adjust x.

### DIFF
--- a/Charts/Classes/Utils/ChartUtils.swift
+++ b/Charts/Classes/Utils/ChartUtils.swift
@@ -125,6 +125,7 @@ public class ChartUtils
         if (align == .Center)
         {
             point.x -= text.sizeWithAttributes(attributes).width / 2.0
+            point.y -= text.sizeWithAttributes(attributes).height / 2.0
         }
         else if (align == .Right)
         {


liuxuan30:
@petester42 I may need your help here. Not very familiar with the tests. Is this because I changed to noDataText position so the related test cases will fail?

petester42:
Probably yes. If the images don't match exactly then they will fail. For easy comparison I use this https://github.com/orta/Snapshots

liuxuan30:
does this mean I need to update all the related images along with this PR?

petester42:
If you believe that the resulting images are correct then ya. It's pretty easier. Set recordMode to true. Run tests. Set it back to false.

liuxuan30:
Thanks. Will take a look at the test suite

danielgindi:
@liuxuan30 I'm sorry to disappoint you, but that function should not adjust the `y` position :-) 

Notice that there are several text drawing functions. This one specifically is for single-line text, where aligning is done on the X axis only (hence `NSTextAlignment` and nothing more were passed)

There are other text drawing functions, where both X and Y are adjusted, according to alignment and anchoring - suited for much complex situations. Written first to allow rendering the rotated axis labels but keep them aligned according to their "outer" bounding rectangles, and then used for other multiline texts.

Is there a specific issue you were trying to solve?

danielgindi:
@petester42 the tests are great. In other cases I would most probably have not taken a deep look at the commit, assuming it's a one-liner and intuitively seems correct...

liuxuan30:
@danielgindi the issue is for #682, saying the no data text is not centered vertically.

Do you think we could adjust Y here?
```swift
            ChartUtils.drawText(context: context, text: noDataText, point: CGPoint(x: frame.width / 2.0, y: frame.height / 2.0), align: .Center, attributes: [NSFontAttributeName: infoFont, NSForegroundColorAttributeName: infoTextColor])
```
it does not count the text height / 2. 

btw, looks like we need some doc how to update test images?

danielgindi:
Yes but that's not where we need to adjust it. We need to offset the whole drawCenterText, but half the height of both the text and description text

danielgindi:
As for tests - it's documented in the tests code :-) There's a line to be set to `true` for recording instead of testing

PR #740, opened 2016-02-05
Fixed label drawing bug

leoMehlig:
For some reason the Y-Axis labels where disappearing when no data was
displayed in the LineChart.
This could happen, for example, if all currently visible data entries
where nil.

** I am not that familiar with all this CGContext stuff and find it vey
strange, that not calling a method in one renderer (LineChartRenderer)
has an affect on an other renderer (AxisYRenderer), but maybe someone
can explain and/or find a better solution then this one **

### DIFF
--- a/Charts/Classes/Utils/ChartUtils.swift
+++ b/Charts/Classes/Utils/ChartUtils.swift
@@ -132,9 +132,11 @@ public class ChartUtils
         }
         
         UIGraphicsPushContext(context)
-        
+        CGContextSaveGState(context)
+
         (text as NSString).drawAtPoint(point, withAttributes: attributes)
         
+        CGContextRestoreGState(context)
         UIGraphicsPopContext()
     }
     


leoMehlig:
Turned out this issue was caused by the LineChartsRenders drawLiniar method. This method saved the context but didn't release it when returning from the method earlier because of missing entries. 
But I still think this fix is valuable, because bug like this can happen very easily and are hard to detect.

liuxuan30:
Is there an issue id to track this? Or screenshot? I don't understand the problem.

danielgindi:
Are you sure this fixes anything? Seems like a push/pop for nothing, as we change nothing on the context...

leoMehlig:
It did fix it in my special case, but I think the better fix would be to change the LineChartRenderer. I'll submit a PR for this soon. 

The problem is that the LineChartRenderer doesn't restore the context when not displaying any data (only in drawLiniar).

PR #746, opened 2016-02-09
LineChartRenderer context bug

leoMehlig:
In the LineChartRenderers drawLinear method the Context gets saved before checking for existing data.
This leads to never restoring the context, which causes other chart-components to disappear.

This commit fixes this by saving the context after the check for data.

The bug described above can be seen in the two GIFs below:

### Behaviour with bug:
chart_bug

### Behaviour with fix:
chart_fixed

Relates to #740


danielgindi:
*This* is the correct fix :-) (Also note that there could have been use of `defer` which avoids the need to  avoid `return` calls...

leoMehlig:
`defer` would have executed the code after the fill function gets called, which could lead to other problems.

PR #755, opened 2016-02-13
update cocoapods url

stevenedds:


petester42:
Pretty sure the version number is not needed here. `http://cocoadocs.org/docsets/Charts` will point to the most recent version

PR #759, opened 2016-02-16
Port ios-charts to OS X.

pixelspark:
This commit adds a thin UIKit/Cocoa abstraction layer (Util/ChartsPlatform.swift). It also adds an OS X demo application.

With reference to https://github.com/danielgindi/ios-charts/issues/43 

ioscharts

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -1423,7 +1423,7 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
-				SDKROOT = macosx;
+				SDKROOT = iphoneos;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				TARGETED_DEVICE_FAMILY = "1,2";
 				VERSIONING_SYSTEM = "apple-generic";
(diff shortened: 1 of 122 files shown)


pixelspark:
Ah, I might have inadvertently set the target platform for the 'Charts-iOS' target to OSX. Simply changing this back to iOS should fix the test.

petester42:
Great work! Looks good to me. I will merge this into a separate branch so that we can work on it to make sure everything is good before it goes into a release.

petester42:
OS X work can be found [here](https://github.com/danielgindi/ios-charts/tree/osx)

danielgindi:
Great! 
It's amazing that this is actually happening. I love open source :-)

pixelspark:
Glad I could help out & good to see this is being adopted by the project! 

I have been testing the charts in [an OSX app](http://warp.one) and have not seen any major issues. Line, bar, radar, pie and scatter plots seem to work fine. Some things to work on:

* Some of the default fonts are a bit small on OSX. The easy fix is to use [NSUIFont systemFontOfSize:] with a different size for OSX. Even better would be to integrate with Dynamic Type on iOS (don't know if it exists at all on OSX?). 

* The gestures need some work. A pinch gesture recognizer is most notably missing in OSX (but can probably be implemented with a custom gesture recognizer and a bit of work).

* Animation of charts is smooth with my 'fake' timer-based implementation of CADisplayLink. We may want to look into CVDisplayLink on OS X for a 'real' implementation.

danielgindi:
No Dynamic Type in OSX...

danielgindi:
I'm not too much worried about defaults - those could be adjusted later... Let's get the real problems fixed first :-)

danielgindi:
It turns out that there's a `NSMagnificationGestureRecognizer` :-)
I'm creating the bridge now...

danielgindi:
There are issues with "nsuiGestureRecognizerShouldBegin" though - as it is not really mapped to anything.

Issue #763, opened 2016-02-19
Proposal: shall we unify yAxis.entries and yAxis.axisRange to have same max and min value?


liuxuan30:
Below discussion is based on the assumption "we should ensure a value label can be displayed(if `isDrawTopYLabelEntryEnabled` is true) that higher than the chart" if this is not true, this issue can be closed.

So I got a data range [0, 833], and I am making the assumption that the y axis will show 0, 200, 400, 600, 800, 1000 labels for me.

However with ChartsDemo, I cannot see the 1000 label:
![sep](https://cloud.githubusercontent.com/assets/4375169/13170057/5d6d59da-d724-11e5-8b60-44d5a24d6705.png)

After some hours debugging, I found that the reason is _leftAxis.axisMaximum is 916.3, while in `computeAxisValues()`, the entries is being calculated as [0, 200, 400, 600, 800, 1000], which is exptected.

Because of 1000 > 916.3, the '1000' label is out of contentRect, even the chartView bounds, so it is not drawn.

This happens only with subtle data ranges.

Increasing `spaceTop` to 0.2 would have similar visual result, but the problem is, we don't know when will axisMaximum * 1.2 < `computeAxisValues`'s calculated max value.

So with the assumption "we should ensure a value label can be displayed(if `isDrawTopYLabelEntryEnabled` is true) that higher than the chart", here I suggest maybe we should unify them.

PR #764, opened 2016-02-19
For #763 proposal: Unify yAxis.axisRange and yAxis.entries value range

liuxuan30:
Here's the proposal for #763, to unify yAxis.axisRange and yAxis.entries value range

to test, change Line LineChart1ViewController.m:
comment out below in ine 85 and 86
```Objective-C
//    leftAxis.customAxisMax = 220.0;
//    leftAxis.customAxisMin = -50.0;
```
set `leftAxis.startAtZeroEnabled = YES;`
replace with below test code in `setDataCount()`
```Objective-C
    {
        double mult = (range + 1);
        double val = (double) (arc4random_uniform(mult)) + 3;
        if (i == 0)
            val = 833;
        if (i == 1)
            val = 0.0;
        [yVals addObject:[[ChartDataEntry alloc] initWithValue:val xIndex:i]];
    }
```

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -273,9 +273,6 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
         _leftAxis?._defaultValueFormatter = _defaultValueFormatter
         _rightAxis?._defaultValueFormatter = _defaultValueFormatter
         
-        _leftYAxisRenderer?.computeAxis(yMin: _leftAxis.axisMinimum, yMax: _leftAxis.axisMaximum)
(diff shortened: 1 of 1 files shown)


liuxuan30:
Oh no @danielgindi I just saw your changes about them in Feb. 18 commits. I am using the old ones to demonstrate

danielgindi:
The new "range" is not entirely accurate. Because the maximum visible label may not necessarily be at the max value visible on the axis. One may choose not to show some of the labels, first, last modulo etc.

liuxuan30:
OK so the assumption in #763 is not valid - "Because the maximum visible label may not necessarily be at the max value visible on the axis" is against "we should ensure a value label can be displayed(if isDrawTopYLabelEntryEnabled is true) that higher than the chart"

I will close them

PR #765, opened 2016-02-19
`computeAxis()` has been called in `notifyDataSetChanged()`, remove redundant calls

liuxuan30:
`computeAxis()` has been called in `notifyDataSetChanged()`, remove redundant calls

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -158,15 +158,6 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
         // execute all drawing commands
         drawGridBackground(context: context)
         
-        if (_leftAxis.isEnabled)
-        {
-            _leftYAxisRenderer?.computeAxis(yMin: _leftAxis.axisMinimum, yMax: _leftAxis.axisMaximum)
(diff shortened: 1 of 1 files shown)


liuxuan30:
not sure if the fail relates to 
```                       
$ cd Charts && carthage bootstrap
*** Skipped downloading ios-snapshot-test-case.framework binary due to the error:
	"GitHub API request failed: Maximum number of login attempts exceeded. Please try again later."
```
I will try run the test on my machine then. @petester42 is there setup guide? I tried search but seems none.

petester42:
That warning is fine. The images don't match.

liuxuan30:
@petester42 I tried the test on my side, it reports
```
...........

	 Executed 11 tests, with 0 failures (0 unexpected) in 1.211 (1.240) seconds
killall Simulator

```
I am starting a new travis to see what happens

liuxuan30:
I found my branch is behind master after Feb. 14. Syncing to latest master will have failed tests same as Travis. Will check

liuxuan30:
@danielgindi are these two `computeAxis()` duplicated? Not sure why you call it in drawRect, but the tests shows it's not necessary?

liuxuan30:
I was wrong - computeAxisValues are needed in drawRect because it needs handle zoom in.

PR #772, opened 2016-02-23
Make ChartViewBase's _data optional. (Fixes #771)

### LINKED ISSUE #771: Can't set ChartViewBase's data property back to nil to clear out data
`ChartViewBase`'s `_data` property is defined as an implicitly unwrapped optional when really it should be a plain optional. The value starts off as nil, but then once it has been set you can no longer set it back to nil to clear out the data in the graph. By making the `_data` property optional that will allow for clearing the chart data (which will allow the `noDataText` to be shown again) and will make use of `_data` safer because its not implicitly unwrapped (eg could cause a runtime error if the property is nil and is accessed).

Right now this code exists to prevent you from setting the `_data` property back to nil.

     if newValue == nil
     {
          print("Charts: data argument is nil on setData()", terminator: "\n")
          return
    }

Note that if `_data` is changed from `data!` to `data?` that we'll need to handle all of the optional cases properly throughout the code base.

ospr:
Changed ChartViewBase's _data property declaration from _data! to _data? to allow for setting the chart's data back to nil. Updated the code base to property handle _data as an optional value.

### DIFF
--- a/Charts/Classes/Charts/BarChartView.swift
+++ b/Charts/Classes/Charts/BarChartView.swift
@@ -42,18 +42,15 @@ public class BarChartView: BarLineChartViewBase, BarChartDataProvider
     {
         super.calcMinMax()
         
-        if (_data === nil)
-        {
-            return
-        }
+        guard let data = _data else { return }
         
-        let barData = _data as! BarChartData
+        let barData = data as! BarChartData
         
         // increase deltax by 1 because the bars have a width of 1
         _deltaX += 0.5
         
         // extend xDelta to make space for multiple datasets (if ther are one)
-        _deltaX *= CGFloat(_data.dataSetCount)
+        _deltaX *= CGFloat(data.dataSetCount)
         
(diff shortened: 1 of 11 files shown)


danielgindi:
Looks nice. With some code moving from alpha Swift to modern Swift also...

How thoroughly was this tested?

ospr:
When I initially made this pull request I had verified that setting the BarLineChartViewBase object's data property to nil would cause it to behave properly and had verified that there were no other regressions by running through all of the charts in the demo project (along with toggling various options) and verifying that they behaved as expected.

However, I realized I forgot to test setting other chart type's data property to nil. I went through the demo project and added a "Toggle Data" option and then checked each chart type to verify that they behaved properly. I found a few minor bugs, fixed them, and squashed them back into this pull request. I made a separate request at #781 which includes the "Toggle Data" option if you wish to pull that in as well.

danielgindi:
Great! Thanks!

Issue #539, opened 2015-11-07
Stepped line charts


andrelind:
Stepped line charts are not supported...?
If we could get that implemented then we can finally drop the huge bloat that is CorePlot from our projects...!

ezamagni:
I agree, this would be really helpful

danielgindi:
What's "stepped line chart"?

ezamagni:
Take an example input of (0,0), (1, 1), (2, 2), (3, 1) and look at the (terrible) sketch below:
![steppedexample](https://cloud.githubusercontent.com/assets/6697707/13321296/82c4f482-dbce-11e5-906e-610afb8d3bef.png)

The first line is what ios-charts would draw: a linear segment connecting each point; the bottom chart is what is called a "stepped line" and it's very useful for sampled data

danielgindi:
Man you only need about 30 minutes, cloning the `drawLinear` function in the line chart - and making it draw to start and end point of each bar instead of just to the center.

That should just be an option to toggle in the dataset to trigger stepped rendering. I do not have the time to implement this right now but I will be more than happy to review a PR :-)

ezamagni:
Thanks for the suggestion, Daniel!  As soon as i manage to get some time i will try to play with it.

danielgindi:
This is merged in. Thank you @ezamagni !

PR #778, opened 2016-02-26
Feature #539 Stepped line charts

ezamagni:
A new flag was added to ILineChartDataSet and relative implementations. The value of the flag is then used to slightly modify the behavior of drawLinear and generateFilledPath in LineChartRenderer.
The appropriate demo viewControllers were also updated with the new flag.


danielgindi:
Thank you!
I will review this soon :-) But it looks good already!

Issue #90, opened 2015-05-18
Can we show Pie chart Value out of slice instead showing inside the slice ?


BurhanMughal:
Self explanatory question.

Can we show slice values of pie chart out side the slice and not in the slice ?
Thanks.

danielgindi:
I think we'll be implementing that soon

BurhanMughal:
Ok thanks.

krahabb:
I've written an algorithm to spread out labels on pie charts in a rather reasonable layout. It works by pushing out labels from each other but still trying to keep them aligned to their respective slice. I could push the feature in few days if needed (it's actually working on top of the original algorithm so I would prefer to add a property in 'somewhere' (haven't yet think about it) to control the application of the algorithm before pushing to master)

yuercl:
@danielgindi ,  does this feature have supported ? i check document , but not found .

kexoth:
@danielgindi  :+1:  for this feature.

NSri:
@danielgindi :+1:  When can we expect this feature?

wjacker:
@BurhanMughal @yuercl  @kexoth @NSri  maybe you can try this pr:  #869 :)

danielgindi:
This is now implemented in `master`

PR #789, opened 2016-03-03
Feature/improve pie for master

KerryJava:
#90 draw value outside the slice


danielgindi:
This feature is already implemented now :-)

PR #795, opened 2016-03-03
Add trailing newline for preprocessor statement

boourns:
@danielgindi 

I had a build error in xcode Version 7.1 (7B91b) because the preprocessor statement was not recognized without a newline on the line:

image


boourns:
My coworker using XCode 7.3 beta 5 did not have this error, so I'm guessing it's dependent on version

danielgindi:
Weird. But we're at Xcode 7.2 if I recall correctly, you should try that :-)
But yeah, Swift Compiler still has bugs, it's not a mature 44 years old C...

PR #814, opened 2016-03-08
codebeat badge

korzonek:
Is it fine to add codebeat badge to README? 

codebeat is automated code review tool for Swift,Ruby,Go & Python that helps get instant feedback on code quality. 

"Quick wins" suggested by codebeat could be a nice candidate for a pull request and help other developers become contributors.

FYI. To be fully open and honest. I'm co-founder of codebeat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 ![alt tag](https://raw.github.com/danielgindi/ios-charts/master/Assets/feature_graphic.png)
   ![Supported Platforms](https://img.shields.io/cocoapods/p/Charts.svg) [![Releases](https://img.shields.io/github/release/danielgindi/ios-charts.svg)](https://github.com/danielgindi/ios-charts/releases) [![Latest pod release](https://img.shields.io/cocoapods/v/Charts.svg)](http://cocoapods.org/pods/charts) [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage) [![Build Status](https://travis-ci.org/danielgindi/ios-charts.svg?branch=master)](https://travis-ci.org/danielgindi/ios-charts)
+  [![codebeat](https://codebeat.co/badges/71f30bdf-1aa9-4dd8-9872-86b5ed9672a6)](https://codebeat.co/projects/github-com-danielgindi-ios-charts)
 
 * Xcode 7 / Swift 2.0
 * iOS 7.0 (Drag .swift files to your project)


liuxuan30:
I like the idea; but not sure what it looks like; the home page does not show many details

korzonek:
@liuxuan30 this is very good idea and is in our backlog of things to do. Feedback is super important for us. We are small team and we launched product at the beginning of this year. Feedback helps us prioritise things in our backlog :) 

If there are any other things you would like to see in codebeat please let me know.

danielgindi:
Actually this seems like taking the development process of track, for a number of reasons:

1. What we are creating here is a cross-platform solution, not just a framework. Stuff that makes more sense in a structural manner on the Swift side, may not be applicable or even possible on the Android side.
2. Many of the suggestions by codebeat are ones that I cannot agree with. For example:
2. 1. "too many lines in one function". The nature of rendering code is being performance-critical, while by definition you should make as little function calls as you can, and the less memory allocations the better. Adding a single function call to a loop over 1000 points in a `drawRect` code has it's impact - as ridiculous as it sounds.
2. 2. Files that seem too big. Breaking down a renderer to a few files, one with drawing the values, one with drawing the circles etc. - in my opinion will make it much harder to maintain and understand. It's not ideal to have too much code to read if you're lazy and do not want to read code, but it's also not that ideal if you can't even find the code that you need to read.

3. I know I kind of mentioned this already, but this is important: I want people to focus on developing new and important (or even nice-to-have) features, not just creating PRs that modify thousands of lines by do nothing. There are times in which we do some concentrated efforts on refactoring (like when moving between Swift versions), but in the everyday development this could make it hard to maintain, break PRs and just generally annoying...

So I'm sorry but I have to reject this one.

grodowski:
Hi @danielgindi, another codebeat dev here. 🙂I fully agree with your concerns about codebeat's suggestions. Complexity and (especially) duplication analysis is relevant as long as your top priority is to write maintainable and understandable business code. Probably this is not your case as those factors have way lower priority when optimising for speed. 

We've addressed this by allowing users to ignore certain paths and files from analysis by adding a `.codebeatignore` dotfile to a repo. Feel free to [read more in our docs](https://help.codebeat.co/docs/path-exclusions). Moreover, we'd like to allow users to mark specific functions as "quality checked" by adding comments directly next to their declarations in code. This is a feature we have on the roadmap.

Feel free to try it with `.codebeatignore`

liuxuan30:
@grodowski I have been trying codebeat in my OJ code; generally it's great to see many green light and had a high GPA lol. It's nice to remind myself that trying to write clean code
however, some of the criteria seems not practical in real projects: like the lines of code, the block depth. 
I know you are just starting it up, too many things on the plate, so not sure if you already considered making the criteria flexible to choose which to use. So if some criteria is not a factor for some people, they can turn it off and use other great ones.

danielgindi:
I think it would have been more practical if it was an opt-in configuration instead of opt-out. Then it wouldn't take too much time to configure.

Ideally I would set rules for code styling etc. and it will be checked through git hooks when creating PRs.

korzonek:
@danielgindi this is very good idea. If there are any other improvements you would like to see in codebeat please share it with us. It will helps us prioritise our backlog.Thank you!

Issue #786, opened 2016-03-02
Positioning on changing of the screen orientation


oxodpavel:
Hi!

There is a minor drawback, that when screen orientation changes, the vertical scale loses its position and goes to some range around original (previous).

Conditions:
* occurs only when some scaling has been performed
* occurs only on positions which are not aligned to the horizontal and vertical axes (to thier zero values)
* changing screen orientation back restores graph position to the original state (seems like, horiz and vert orientations have independent positioning parameters, am I write)
![test_app_land](https://cloud.githubusercontent.com/assets/17591717/13459119/8073139a-e082-11e5-9de6-1ad5a2b81b02.jpg)
![test_app_port](https://cloud.githubusercontent.com/assets/17591717/13459120/80758a30-e082-11e5-931b-5883b686c613.jpg)

leoMehlig:
I had the same problem and solved it by overriding this method in my Chart-Class:

`public override func observeValueForKeyPath(
keyPath: String?,
ofObject object: AnyObject?, 
change: [String : AnyObject]?, 
context: UnsafeMutablePointer<Void>) {     `

        var oldPoint: CGPoint?
        if keyPath == "frame" || keyPath == "bounds" {
            oldPoint = viewPortHandler.contentRect.origin
            getTransformer(.Left).pixelToValue(&oldPoint!)
        }
        super.observeValueForKeyPath(keyPath, ofObject: object, change: change, context: context)
        if var p = oldPoint {
            getTransformer(.Left).pointValueToPixel(&p)
            viewPortHandler.centerViewPort(pt: p, chart: self)
        }
        
    } `

(Somehow the code function of the github md isn't working properly)

liuxuan30:
I have it fixed @leoMehlig

oxodpavel:
Yes, thank you! This patch works fine

leoMehlig:
@liuxuan30 is this a feature that should be implemented internally?

liuxuan30:
@leoMehlig I'm not sure if everyone has such requirements, some may want keep centered, while some want nothing changed :) If you think it's useful, than add a switch to turn it on and a PR :)

PR #824, opened 2016-03-10
Keep position on rotation

leoMehlig:
I’ve added a flag (keepPositionOnRotation) to BarLineChartViewBase indicating whether the chart
should stay at its current position or not

The flag defaults to false to not change any existing behaviour.

This feature was requested/discussed in #786.

The result can be seen in the two picture pairs below:

## keepPositionOnRotation = false (old behaviour):
simulator screen shot 10 mar 2016 11 03 26
simulator screen shot 10 mar 2016 11 03 31

## keepPositionOnRotation = true:
simulator screen shot 10 mar 2016 11 04 29
simulator screen shot 10 mar 2016 11 04 33

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -52,6 +52,9 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
     /// Sets the minimum offset (padding) around the chart, defaults to 10
     public var minOffset = CGFloat(10.0)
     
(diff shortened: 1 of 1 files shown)


danielgindi:
This is a great feature!

I'll see how we can migrate this to the Android side and then merge. If you are working on this crossplatform and want to issue a PR there as well - please the PR from the other one so we can synchronize the merge.

leoMehlig:
@danielgindi I'm looking into it. Seems like the current behaviour is different. Android completely zooms out, while iOS keeps its scale.

leoMehlig:
Summited [PR on MPAndroidChart](https://github.com/PhilJay/MPAndroidChart/pull/1565/files)

PR #836, opened 2016-03-12
Fixes issue with labels in horizontal bar chart #124

### LINKED ISSUE #124: HorizontalBarChartView drawValueAboveBarEnabled label rendering issue
When i set my HorizontalBarChartView's drawValueAboveBarEnabled to false, I get the following layout issue when the BarChart has a y Value too small to fit the value label

screen shot 2015-06-01 at 3 35 37 pm

In this case, I set the third bar to have a value of 1.00. 

Do you have any ideas of how I can get that label to render on the right side of the bar instead of the left?

alexandreos:
Fixed #124 HorizontalBarChartRenderer to draw label above bar if it doesn't fit the bar when the drawValueAboveBar is set to false.

I hope you find this useful.

Thanks.


jjatie:
Rebase on 4.0.0 and reopen.

Issue #665, opened 2016-01-11
Right Y-Axis values displaying incorrectly in LineChartDemo


adriansoghoian:
Thanks for an awesome library, my app is using it extensively to plot stock market related data. When the app launches, I'll send a note in case you'd like to take a look. 

I currently have a strange bug where the Y-axis values are being plotted incorrectly -- it looks like they're being slightly offset. 

![screen shot 2016-01-11 at 11 02 05 am](https://cloud.githubusercontent.com/assets/3008744/12239546/3c4ae63c-b857-11e5-9911-f34a1c982982.png)

I'm not doing any custom scaling with the data. Here are the relevant code snippets below: 

```
func setChartDefaults() {
    // No data view
    self.lineChartView!.noDataText = ""
    self.lineChartView!.descriptionText = ""
    self.lineChartView!.noDataTextDescription = ""
    
    // Legend
    self.lineChartView!.legend.enabled = false
    

    // Interaction
    self.lineChartView!.setScaleEnabled(true)
    self.lineChartView!.pinchZoomEnabled = true
    self.lineChartView!.doubleTapToZoomEnabled = true

    // Background
    self.lineChartView!.drawGridBackgroundEnabled = true
    self.lineChartView!.gridBackgroundColor = UIColor.clearColor()
    self.lineChartView!.backgroundColor = UIColor.clearColor()

    // Borders
    self.lineChartView!.borderColor = UIColor.clearColor()
    self.lineChartView!.drawBordersEnabled = false

    // Format x axis
    self.lineChartView!.xAxis.enabled = true
    self.lineChartView!.xAxis.labelPosition = .Bottom
    self.lineChartView!.xAxis.drawAxisLineEnabled = false
    self.lineChartView!.xAxis.drawGridLinesEnabled = false
    self.lineChartView!.xAxis.labelFont = UIFont.TriggerFontSemiBold(11)
    self.lineChartView!.xAxis.axisLineWidth = 1.0
    self.lineChartView!.xAxis.spaceBetweenLabels = 9
    self.lineChartView!.xAxis.avoidFirstLastClippingEnabled = true
    
    // Remove left axis
    self.lineChartView!.leftAxis.enabled = false
    self.lineChartView!.leftAxis.drawAxisLineEnabled = false
    self.lineChartView!.leftAxis.spaceTop = 0.5
    self.lineChartView!.leftAxis.startAtZeroEnabled = false
    self.lineChartView!.leftAxis.spaceBottom = 0.4
[...]

PR #843, opened 2016-03-14
Testing AxisDependency before using the data

AndreasIgelCC:
If disabling a YAxis and not setting the datasets AxisDependency correct, there could be strange behaviour (see #665). This change will notify the developer of wrong dataset-settings.

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -288,6 +288,39 @@ public class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChar
         _leftAxisTransformer.prepareMatrixOffset(_leftAxis.isInverted)
     }
     
+    /// The data for the chart
+    public override var data: ChartData?
+    {
+        get
+        {
+            return _data
+        }
+        set
+        {
+            if newValue == nil
+            {
+                print("Charts: data argument is nil on setData()", terminator: "\n")
+                return
+            } else {
+                //Testing datasets for using a disabled axisDependency
+                if let datasets = newValue?.dataSets {
+                    for ds in datasets {
+                        if _leftAxis.enabled == false && ds.axisDependency == ChartYAxis.AxisDependency.Left {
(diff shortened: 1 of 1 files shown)


danielgindi:
I'm actually not sure that we should pollute the console here.
Imagine adding data on the other axis - but hiding that axis... A common use case, but will produce a warning!

AndreasIgelCC:
Do you have an example in this framework how to show that warning without polluting the console?
Just adding the data on the other axis would be a solution, just removing this data would be a solution too. But in my opinion we should give a feedback if the developer makes a mistake in using the framework (like adding data to disabled axis).

danielgindi:
We have no way to know if it's a mistake. Therefore no warning...
The `NSLayoutConstraint`s for example - do issue warnings - when they know that you have absolutely positively made a mistake, due to two colliding constraints.
In our case - what you are describing is completely valid.

danielgindi:
It's not even possible (to my knowledge) to do conditional warning like "if you're in debug mode", because those macros get compiled and removed when building the Framework release.

AndreasIgelCC:
Thank you for your feedback! I agree there is a usecase where developer dont want to show axis but want to show data. So always printing a warning would not be a good solution!

Just one question: Print should only write to console when its in debug mode or do I have a misunderstanding in context of frameworks?

danielgindi:
I don't really know. From my experience console log works in prouction also. In objc projects I use macros to disable logs in release build. It's possible that it's different in Swift- but still it's too much verbosity...

petester42:
Swift's print function works differently. It does not actually print to the device logs like NSLog does. Could `precondition` be what is needed here?

danielgindi:
A `precondition` would mean that the user is absolutely wrong by having the axis invisible... Which is not true
http://stackoverflow.com/questions/29673027/difference-between-precondition-and-assert-in-swift

AndreasIgelCC:
You are right, the developer might have disabled the axis because he just dont want to see it. Thinking about the topic deeply, I find no condition in which we can say: "The chart-config is invalid!". So you are right in closing the pull request.

Issue #822, opened 2016-03-09
BarChartView: Rectangle with borders and without filling color 


AndreasIgelCC:
Hi at all,
as I understood, there is no way of displaying just bar borders without filling. Would it be a desired feature or should I write only an extension for our project? 

Greetings, 
Andreas

liuxuan30:
A feature PR is always welcome :) I guess there are demands have the bars stroked but not filled

PR #844, opened 2016-03-14
Supporting borders on bars, Fixes issue #822

AndreasIgelCC:
Based on my feature request (see #822).
With these changes, the border color of bars could be set independently. Updates on example project helps understanding this feature.


gStanev:
Is it possible to achieve the same behavior but this time in PieChart?

PR #847, opened 2016-03-15
Type bug in PieChartData

leoMehlig:
The property dataSet on PieChartData did expect a PieChartDataSet and
tried to convert the result. If the user was using a Interface to
represent the data, the pie chart would fail, because the yValueSum method would return 0.0


danielgindi:
Great, thank you!

Issue #194, opened 2015-07-08
Making X-Values Properly Scaled (Not Equidistant)


ghost:
Hi, I'm making a line chart with the following example data points:

(0,0)
(1,0)
(5,0)

Is there a way I can get the 3 points to show up so that they're scaled properly? Right now, if I add a Data Set with 2 arrays, all 3 points are the same distance from each other. I know there's a thread about this in MPAndroidCharts, but I was wondering if anyone's figured out how to do this in iOS. Thank you so much!

liuxuan30:
are you saying they are equal distance on xAxis?

ziruixiao:
Hi Xuan,

Yes. The 3 points are the same distance apart on the x axis. Therefore, the graph of 
(0,0) (1,0) (2,0) looks the same as (0,0) (1,0) (100,0) with the exception of the labels. I would like (0,0) and (1,0) to be very close together while (1,0) and (100,0) are far apart. I'm not sure if that makes sense. Thanks for the reply!

Best,
Felix

liuxuan30:
Great, I have been the first (I guess) person to ask for this feature. I have filed the request for both Android and iOS, however no response yet... I hope more and more people can realize this is an important feature for professional users.

@PhilJay and @danielgindi  I think we are aiming at providing the world-class charts. I would love to help on this, however, I cannot see the whole picture for all the features inside the projects. You guys have better understanding on it. From what I have tried, render xAxis like yAxis is not the tough part, what blocks me is the xAxis labels is not in the right place while zooming and scaling. I still don't get how to correctly render the labels, considering the label width and the position.

liuxuan30:
FYI, https://github.com/danielgindi/ios-charts/issues/176 and https://github.com/PhilJay/MPAndroidChart/issues/821 is the request I have been filed.

danielgindi:
This is becoming a popular feature, we know. It is quite complex to implement and we haven't got to designing it yet...
It involves of course changing the whole coordinate system to something that cannot assume anything about the X axis. We haven't got to the details yet but I'm planning on learning this topic soon.

danielgindi:
"Soon" meaning when I get the time to sit a whole night on it. I have a wedding to plan at the moment so you can see I'm a little less active here :-)

liuxuan30:
@danielgindi HUGE congratulations to you from China!

aaga:
Can you give us a rough idea of when this might be implemented? And congratulations!

[...]

PR #856, opened 2016-03-17
Start of work for floating point precision for the x-axis, Issue 194

StephanieB5:
The purpose of this pull request is to open a place for conversation about work I've started to resolve Issue #194 .  This is a work in progress.  My goal is to have the work pulled into mainline when it is ready.  On my fork I've started a Wiki page to give an idea of my current understanding of the problem and roadmap for my work. 

https://github.com/StephanieB5/ios-charts-xTime/wiki/ios-charts-xTime:-support-of-floating-point-and-date-x-axis-values

@RyGuyM, @liuxuan30 have been great to offer support and mentioned interest in a conversation.  I feel code is greatly improved by review and feedback so please let me know what you think.

Thanks!


RyGuyM:
Let's get the fork up to date with the current version. I took a quick stab at Scatter Charts on my local branch yesterday and made good progress on a scaled value rendering but where things start to break down are the labels and grids.

StephanieB5:
I've got the fork up to date.  Would you like to work together to get this feature rolled out?  I'm always open to ideas. :)

liuxuan30:
@danielgindi since it's a big move, shall we use another branch for it?

danielgindi:
Of course- this kind of stuff should be thoroughly tested before going into master... Feel free to create a branch

danielgindi:
Btw, just a hint - it's possible that all you need is actually an extension for the XAxis, not a whole chart-view. Otherwise you would have to create time versions of almost all charts. 
Just try to generify it.

StephanieB5:
Thanks @liuxuan30 and @danielgindi!  A branch would be great.  It's in a branch on my side too. :)  A generalized XAxis should hopefully take care of labelling issues.  However, I believe the changes will spread beyond that though. For example functions like BarLineChartViewBase::lowestVisibleXIndex will need to be changed.  

That being said, in the end I'm hoping that the changes end up being small enough that classes like BarLineChartViewBase can be safely changed.  I've just isolated things into separate classes for now.

RyGuyM:
I'm doing some work on the lowestVisibleXIndex right now which doesn't work in a value scheme. In the index scheme order is based on index so we can always assume index[0] will appear before index[1] but in the value scheme index[0] might have a value of 10 and index[1] might be -5 so we need to filter or renderings based on position instead.

I agree with @danielgindi that an extension for XAxis for time values could useful. I haven't looked at times yet though, just numerical based scaling.

StephanieB5:
If you use time as represented in seconds from a reference date, such as NSTimeInterval, then hopefully getting date representation from values should just be a matter of formatting the labels. The XAxis has space for a ValueFormatter, it just needs to be generalized to support customs formatting.

I suspect the data points will need to be ordered by xNumericVal before rendering.  For example in line chart rendering you to know which point is next to which according to their place on the x-axis.

@RyGuyM it appears that we're working in the exact same area of the code.

liuxuan30:
I have created a branch floatXAxis from master. You can merge it there. Keep the code style :)

StephanieB5:
Thanks!  

Sorry about the scattered commits.  On my end, I like to do fine grain commits to retrace steps in case bugs appear.  I'll squash those together to make it easier to see the changes overall.  I appreciate the feedback. :)  

Once things are squashed I'll open a new pull request (into floatXAxis this time).

StephanieB5:
Closing this pull request in preparation for pull request into floatXAxis branch.

PR #872, opened 2016-03-23
Replace features deprecated in Swift 2.2

colavitam:
I went through the codebase and replaced all increment operators, decrement operators, and C-style for loops with their recommended replacements in Swift 2.2. Since the mechanics of for-in loops are slightly different than those of C-style for loops (as a range a ..< b cannot be created such that a > b), I added conditions for the for loops in which that appeared necessary. A double check on those edge cases would be appreciated.

I refrained from upgrading selectors, as this would break compatibility with previous versions of Swift. I can create a request for those changes too if backward compatibility is not necessary.


danielgindi:
Sorry but there's already a `swift3` branch

dennisweissmann:
@danielgindi Can you point me to that branch? I cannot find it :(

liuxuan30:
@dennisweissmann it has been merged into master for a while, so the branch is deleted. Why asking?

dennisweissmann:
@liuxuan30 Because I'm working with the latest dev snapshot on a project and there are quite a few changes (which I would like to push once I'm finished so that others don't need to spend time on it).

PR #880, opened 2016-03-26
fix line chart renderer crash: can't form range when end < start

liuxuan30:
ChartsDemo will crash all the time for the Line Chart1 Chart view controller,
This PR fix the line chart renderer crash: `can't form range when end < start` while phaseX is 0 (from + 1 ..< from + 0 is crashing); use `stride` instead, seems stride can gracefully fail the loop

@danielgindi I clicked the charts demo one by one, seems others are good, but I'm not sure if any of the rest for loops will have similar issue. Let's fix what we met first.

### DIFF
--- a/Charts/Classes/Renderers/LineChartRenderer.swift
+++ b/Charts/Classes/Renderers/LineChartRenderer.swift
@@ -440,7 +440,7 @@ public class LineChartRenderer: LineRadarChartRenderer
         }
         
         // create a new path
-        for x in from + 1 ..< Int(ceil(CGFloat(to - from) * phaseX + CGFloat(from)))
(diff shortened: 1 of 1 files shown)


liuxuan30:
@petester42 I see builds keeps failing for swift 3? How can we fix it, or we wait Travis CI to fix? Seems the test machine's compiler is not 2.2 yet.

petester42:
Travis has to update for it to work.

liuxuan30:
@petester42 what a quick reply! Thanks

danielgindi:
Yeah I've just tried to update the travis configuration file, as they've added xcode 7.3 to their supported version. Let's see if it builds :-)

liuxuan30:
reran today, still complier error :(

PR #887, opened 2016-03-28
add a switch whether to draw limit line's labels. default is true

liuxuan30:
add a switch whether to draw limit line's labels. default is true


danielgindi:
There's a typo there :-)

Anyway, by default the label is not rendered, as it is empty. So I'm guessing if the user has set a label - it means he wants it to draw. Am I missing something here?

liuxuan30:
typo fixed and squashed.
Normally what you said is true. But I did this because our UE has a sick demand... the label is there, and user decides whether to display it. I have to store the label somewhere if I don't have a switch, so I guess it may help others

Issue #888, opened 2016-03-28
Two line in X-Axis


hardikdevios:
Hello,
My Current data is something like this 

`["Sun\n12:00 AM", "Sun\n03:11 AM", "Sun\n03:12 AM", "Sun\n03:27 AM", "Sun\n03:46 AM", "Today\n12:00 AM", "Today\n12:05 PM"]
`
```
 func chartSetup(){ 

        lineChartView.descriptionText = ""
        lineChartView.extraRightOffset = 10
        lineChartView.rightAxis.enabled = false
        lineChartView.highlightPerTapEnabled = false
        lineChartView.drawGridBackgroundEnabled = false
        
        lineChartView.borderLineWidth = 1.0
        lineChartView.leftAxis.axisLineWidth = 0.5
        lineChartView.xAxis.axisLineWidth = 0.5
        lineChartView.legend.enabled = false
        lineChartView.pinchZoomEnabled = false
        lineChartView.dragEnabled = false
        lineChartView.doubleTapToZoomEnabled = false
        
        lineChartView.leftAxis.startAtZeroEnabled = false
        lineChartView.leftAxis.spaceBottom = 0
        lineChartView.leftAxis.gridLineDashLengths = [(3.0)]
        lineChartView.leftAxis.drawGridLinesEnabled = true
        lineChartView.xAxis.labelPosition = .Bottom
        lineChartView.xAxis.drawGridLinesEnabled = true
        lineChartView.leftAxis.xOffset = 7
        lineChartView.leftAxis.valueFormatter = NSNumberFormatter()
        lineChartView.leftAxis.valueFormatter?.alwaysShowsDecimalSeparator = false
        
        lineChartView.xAxis.avoidFirstLastClippingEnabled = true
        
        lineChartView.extraLeftOffset = 5
        lineChartView.extraRightOffset = 25
        lineChartView.infoTextColor = APP_COLOR_COMPUTED

        lineChartView.xAxis.wordWrapEnabled = true
        lineChartView.xAxis.labelHeight = 200
        lineChartView.animate(xAxisDuration: 1.0, yAxisDuration: 1.0, easingOption: .EaseInBounce)
        


    }
```
Here is the correct View 
![screen shot 2016-03-28 at 2 36 12 pm](https://cloud.githubusercontent.com/assets/6657591/14074940/7dc687e2-f4f2-11e5-8787-91362fe8076c.png)

I want to display

**Today
12:00AM**

Possible ? if yes please let me know,
[...]

PR #894, opened 2016-03-29
remove duplicated statement

liuxuan30:
remove duplicated statement according to #888


liuxuan30:
I merged it since it's clear and simple.

PR #934, opened 2016-04-09
Property circleHoleRadius added to ILineChartDataSet protocol. 

olbartek:
Property circleHoleRadius added to ILineChartDataSet protocol. The new circleHoleRadius value is set only if it is less than circleRadius. Otherwise a default value (4.0) is used.

### DIFF
--- a/ChartsRealm/Classes/Data/RealmLineDataSet.swift
+++ b/ChartsRealm/Classes/Data/RealmLineDataSet.swift
@@ -30,6 +30,7 @@ public class RealmLineDataSet: RealmLineRadarDataSet, ILineChartDataSet
     // MARK: - Styling functions and accessors
     
     private var _cubicIntensity = CGFloat(0.2)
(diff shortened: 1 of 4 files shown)


ChartsRealm/Classes/Data/RealmLineDataSet.swift
```diff
@@ -69,6 +70,18 @@ public class RealmLineDataSet: RealmLineRadarDataSet, ILineChartDataSet
     /// The radius of the drawn circles.
     public var circleRadius = CGFloat(8.0)
     
+    /// The radius of the drawn circles.
+    public var circleHoleRadius : CGFloat {
```
danielgindi: code styling...

olbartek: What exactly should be changed?

danielgindi:
As the label says - "code styling" :-)
It should match the project

PR #937, opened 2016-04-11
Fix a crash when using markers with a PieChart

rofreg:
fa5d01f introduced a crash when using a MarkerView with a PieChartView – because PieChartView no longer has an `_xAxis`, any attempt to call `drawMarkers` will crash [here](https://github.com/danielgindi/Charts/blob/b1dcdbd25f4b3bff3fba9c9593fcede1bb148617/Charts/Classes/Charts/ChartViewBase.swift#L533). I'm not sure if this is the ideal fix with how you've been refactoring the ChartView classes, but it does restore the ability to use markers with a pie chart.

### DIFF
--- a/Charts/Classes/Charts/ChartViewBase.swift
+++ b/Charts/Classes/Charts/ChartViewBase.swift
@@ -530,7 +530,7 @@ public class ChartViewBase: NSUIView, ChartDataProvider, ChartAnimatorDelegate
             let highlight = _indicesToHighlight[i]
             let xIndex = highlight.xIndex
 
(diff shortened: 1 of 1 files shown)


danielgindi:
@PhilJay I'm thinking "Oops!" :-) In the latest axis changes, we didn't notice that a null xAxis might cause trouble. This will happen on Android too.

danielgindi:
Thanks @rofreg

PhilJay:
Yep, it did happen, someone mentioned this already and provided a PR!

PR #941, opened 2016-04-12
fix #940. another loop bounds crash

### LINKED ISSUE #940: highestVisibleXIndex is smaller than lowestVisibleXIndex
I just downloaded the demo app, adding one line ` _chartView.autoScaleMinMaxEnabled = YES;` to the BarChartViewController.m at line:56.

And I run the demo app, and open the bar chart, zoom in. It crashes.
The screenshot:
[screenshot](https://dl.dropboxusercontent.com/s/dqtwiw9ix4qmyql/Screenshot%202016-04-12%2017.23.11.png)
Back trace:
[back trace](https://dl.dropboxusercontent.com/1/view/uchzyycy27rgm96/Screenshots/Screenshot%202016-04-12%2018.03.11.png
)

liuxuan30:
fix #940. another start ... endValue crash
use stride instead.

I searched there are many ... and ..<, I'm not sure how many loop is in danger... right now we can only fix what we have met.


danielgindi:
Thanks! I hate those

PR #944, opened 2016-04-12
Allow PieChartView to hide labels for tiny slices

rofreg:
Right now, if you have a labelled pie chart with a large number of tiny sections, the labels are very likely to overlap in a visually unappealing way. This commit adds a basic property to auto-hide labels for any pie chart section below a certain size – e.g. only show pie chart labels for pie sections that span at least 20º.

I also updated the color of the labels in the PieChart demo from white to black for legibility.

Before:
screen shot 2016-04-12 at 11 58 33 am

After:
screen shot 2016-04-12 at 11 58 29 am

### DIFF
--- a/Charts/Classes/Charts/PieChartView.swift
+++ b/Charts/Classes/Charts/PieChartView.swift
@@ -25,6 +25,7 @@ public class PieChartView: PieRadarChartViewBase
     private var _circleBox = CGRect()
     
     private var _drawXLabelsEnabled = true
+    private var _drawXLabelsMinimumAngle: Double = 0
     
     /// array that holds the width of each pie-slice in degrees
     private var _drawAngles = [CGFloat]()
@@ -510,6 +511,19 @@ public class PieChartView: PieRadarChartViewBase
         set
         {
             _drawXLabelsEnabled = newValue
+            setNeedsDisplay()
+        }
+    }
+    
+    public var drawSliceTextMinimumAngle: Double
+    {
+        get
+        {
+            return _drawXLabelsMinimumAngle
+        }
+        set
+        {
+            _drawXLabelsMinimumAngle = newValue
             setNeedsDisplay()
         }
     }
--- a/Charts/Classes/Charts/PieChartView.swift
+++ b/Charts/Classes/Charts/PieChartView.swift
@@ -25,7 +25,7 @@ public class PieChartView: PieRadarChartViewBase
     private var _circleBox = CGRect()
     
     private var _drawXLabelsEnabled = true
-    private var _drawXLabelsMinimumAngle: Double = 0
+    private var _drawXLabelsMinimumAngle: CGFloat = 0
     
     /// array that holds the width of each pie-slice in degrees
     private var _drawAngles = [CGFloat]()
@@ -515,7 +515,7 @@ public class PieChartView: PieRadarChartViewBase
         }
     }
     
-    public var drawSliceTextMinimumAngle: Double
+    public var drawSliceTextMinimumAngle: CGFloat
     {
         get
         {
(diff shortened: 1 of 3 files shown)


Charts/Classes/Charts/PieChartView.swift
```diff
@@ -25,6 +25,7 @@ public class PieChartView: PieRadarChartViewBase
     private var _circleBox = CGRect()
     
     private var _drawXLabelsEnabled = true
+    private var _drawXLabelsMinimumAngle: Double = 0
```
DingSoung: Use CGFloat

DingSoung:
This is not a good solution, as some big slices have short x value string(should display), and some small slice have long x value (should hide), a specific value of angle could not determine this.

danielgindi:
@DingSoung I would say "it's not the perfect solution", not necessarily not a "good" one.
@rofreg Some more magic needs to be put in there.

Charts/Classes/Charts/PieChartView.swift
rofreg: Good point! I could have sworn that I used a `Double` here for a reason, but I can't remember that reason now, so I've switched it to a `CGFloat`.

rofreg:
@danielgindi Do you have any recommendations on how to make it more magic? I did try a few solutions that were more intricate, but in the end I thought it was best to keep this simple. (For my purposes, the most important thing was to hide the labels for a series of extremely small slices, each around 1% of the pie or less – they were all showing up on top of each other in a garbled mess. Not sure how common that is for other people.)

A few things I considered:
- Should X labels be allowed to extend beyond the bounds of the pie slice? Right now they can, and I figured that was more of a feature than a bug. Filtering by slice size seemed like an elegant way to hide less important labels (assuming that smaller pie segments are less important), while still allowing labels of any size.
- Alternately, should X labels only be shown if they fit entirely inside the pie slice? If so, then code would need to account for the size of the text, the size of the pie slice, **and** the specific rotation angle of the pie slice. (e.g. A pie slice of 10º might be able to display a long label if it's horizontal, but not if it's vertical.) That seemed like overkill when I was mainly trying to hide labels for ultra-small slices, but I could potentially revisit that.
- If an X label is too large to be displayed, should it be truncated instead of hidden? That seemed like a separate setting to me, so I left it out.

Let me know if you have thoughts on any of those possibilities!

DingSoung:
@danielgindi @rofreg 
I have tried, here is effect
screen shot 2016-05-04 at 2 22 28 pm

way:
1 add a Bool property for PieChartView
```
    private var _limitXLableSizeInSlice: Bool = false
    public var limitXLableSizeInSlice: Bool {
        get {
            return _limitXLableSizeInSlice
        }
        set {
            _limitXLableSizeInSlice = newValue
            setNeedsDisplay()
        }
    }
```
2 in drawValues of PieChartRenderer, check before draw, example
```
                    else if drawXInside
                    {
                        let attributes = [NSFontAttributeName: valueFont, NSForegroundColorAttributeName: dataSet.valueTextColorAt(j)]
                        let size = data.xVals[j]!.sizeWithAttributes(attributes)
                        
                        if (chart as? BDPPieChartView)?.limitXLableSizeInSlice == true
                            && size.width > labelRadius * sliceAngle * ChartUtils.Math.FDEG2RAD {
                            
                        } else {
                            ChartUtils.drawText(
                                context: context,
                                text: data.xVals[j]!,
                                point: CGPoint(x: x, y: y + lineHeight / 2.0),
                                align: .Center,
                                attributes: attributes
                            )
                        }
                    }
```

I do this in subClass and used particular configuration, some more need consider such (with hole or not, show y value or not etc.)
if you think this way ok, I will pull request later

rofreg:
@DingSoung
Thanks for the code sample! :) I definitely see the appeal of that approach, though there are limitations on basing it on `labelRadius * sliceAngle * ChartUtils.Math.FDEG2RAD`:

- The label will still appear in some situations where it is too large
too-big
- The label will not appear in some situations where it would actually fit
too-small

We could potentially do more advanced math to figure out the actual maximum size (calculating the largest box that could fit in the center of the pie slice, then hiding the label if it doesn't fit in that box), but that seemed like overkill to me, at least for the application where I used this code. The main goal I wanted to achieve was to hide labels for insignificant pie slices (e.g. sections that only make up 1% of the pie), in order to filter out "less important" parts of the data set and highlight the larger pie slices.

rofreg:
I'm closing this in favor of #2062, which appears to reimplement the same improvements in Swift 3.

PR #969, opened 2016-04-22
Proposal: simplify horizontal bar chart Y axis Label offset calculation

liuxuan30:
Context:
I notice in `renderAxisLabels`, there is a hard coded `baseYOffset = 2.5`, and when turnning the legend off, the horizontal bar chart's y axis labels will be partially cut off. This is because `yOffset` is set to 0 by default, and we hard code 2.5 offset. This is not discovered when legend enabled because we have enough space for the label.

I found in recent changes, we have below change for `getRequiredHeightSpace()` that impacted the label height:

```
     public func getRequiredHeightSpace() -> CGFloat
     {
-        return requiredSize().height + 2.5 * 2.0 + yOffset
+        return requiredSize().height
     }
```

So to align the idea, while fixing bug:
1. replace horizontal bar chart `renderAxisLabels()` duplicated `baseYOffset = 2.5` with `yAxis.yOffset`. We can use `yOffset` to adjust, so `drawYLabels()` does not need offset as well.
2. we could also reduce requiredSize `yOffset` to 1x in `requiredSize()`, we can use `extraOffset` to adjust the rest half

I am not sure if I missed anything especially on Android side, so please review carefully.

PS: if we want to go this way, I will record new tests later.

### DIFF
--- a/Charts/Classes/Charts/HorizontalBarChartView.swift
+++ b/Charts/Classes/Charts/HorizontalBarChartView.swift
@@ -34,6 +34,10 @@ public class HorizontalBarChartView: BarChartView
         _xAxisRenderer = ChartXAxisRendererHorizontalBarChart(viewPortHandler: _viewPortHandler, xAxis: _xAxis, transformer: _leftAxisTransformer, chart: self)
         
         self.highlighter = HorizontalBarChartHighlighter(chart: self)
+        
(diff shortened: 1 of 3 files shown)


danielgindi:
I think that this too is irrelevant anymore :-)

PR #978, opened 2016-04-25
Added fun getSnapshot -> UIImage

128keaton:
Useful for if you want to do a share sheet


danielgindi:
There's already a `getChartImage` function.
Also - next time, please keep the code styling matching the project.

PR #982, opened 2016-04-27
Bug fix: Move fill before stroke for drawLinear()

liuxuan30:
move fill before stroke for drawLinear(). if we fill after stroke, half of the stroke, line width will be covered by the fill path.

so if the fill color has some transparency, half of the line will not be solid, like:
fill

PS: I have checked drawCubicBezier and drawHorizontalBezier is filling before stroking.


liuxuan30:
@danielgindi time for some bug fix reviews. I am going to review some bug fixes before v3 release. Not sure the impact to v3 branch.

PR #995, opened 2016-04-29
Fixed limit lines on xAxis position for bar charts with several data sets.

KiArthur:
Limit lines on xAxis position are calculated wrong when bar chart has more than one data set.
beforev3
Same for horizontal bar charts.
beforeh3
I looked how grid lines position are calculated, and did similar calculations for limit lines.
afterv3
afterh3


danielgindi:
This PR is irrelevant as of Charts 3.0, thank you :-)

PR #1010, opened 2016-05-06
Draw one more cubic point at left side to keep the cubic shape while scrolling

liuxuan30:
For cubie bezier lines:

Because it involves prevPrev, prev, cur, next (two different control points's X), when we zoom in and scroll, sometimes the cubic shape will change like below image:

before:
before

after:
after

So I propse we draw one more cubic point at left side to keep the cubic shape while scrolling.

I didn't see the symptom on drawHorizontalCubic (because same control point.x?), so I didn't touch it.

### DIFF
--- a/Charts/Classes/Renderers/LineChartRenderer.swift
+++ b/Charts/Classes/Renderers/LineChartRenderer.swift
@@ -103,7 +103,7 @@ public class LineChartRenderer: LineRadarChartRenderer
             else { return }
         
         let diff = (entryFrom == entryTo) ? 1 : 0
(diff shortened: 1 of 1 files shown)


danielgindi:
Yeah we had this - but it caused some crashes. After the safeguards put in in the latest updates - I hope this will become safer.
But I recall this causing another problem, by having the curve and the circles not in sync when animating. Did you test that?

liuxuan30:
could you point out "After the safeguards put in in the latest updates"?

I will check the circles later.

PR #1040, opened 2016-05-16
Added callback for pan gesture has ended or canceled

aastein:
Added the method in ChartViewBase protocol. Method is called within BarLineChartViewBase.swift in the panGestureRecognized function.

This is a required functionality if something is done outside of Charts while the pan gesture is occurring, which uses data retrieved during the pan gesture. 

Example: During the pan gesture I want to display the currently highlighted value in a chart on a label in my view. This can be done with the current callbacks. If I want that value to be removed when the pan gesture is ended, because nothing is actively being selected by the user, there is currently no callback for that.

### DIFF
--- a/Charts/Classes/Charts/BarLineChartViewBase.swift
+++ b/Charts/Classes/Charts/BarLineChartViewBase.swift
@@ -799,6 +799,11 @@ public class BarLineChartViewBase: ChartViewBase, UIGestureRecognizerDelegate
                 _outerScrollView?.scrollEnabled = true
                 _outerScrollView = nil
             }
+            
+            if (delegate !== nil)
+            {
+                delegate?.chartPanEnded?(self)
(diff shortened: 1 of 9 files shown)


danielgindi:
Please have a PR contain only code related to that PR.

Regarding this PR: You don't need it. You just add another PanGestureRecognizer to the chart view from your code. It's a public mechanism in iOS/OSX

sinnott-balin:
@danielgindi Sorry to reference an old issue, but I am running into a similar problem in my project.  If I try to add my own PanGestureRecognizer to the LineChart, the current PanGestureRecognizer provided by BarLineChartViewBase gets overwritten.  Is it possible to add my own PanGestureRecognizer to the LineChart to detect state == ended while keeping the current functionality of BarLineChartViewBase?

PR #1049, opened 2016-05-18
- Fix bubble overlap selection 👾

### LINKED ISSUE #654: Selection on bubble chart
Hi All ios-charts users 😃. 

First, congrats on all the work @PhilJay and @danielgindi put into these libraries. High quality products, both.

Then, I found the little bug in the bubble chart, the specific case is when you have more than one bubble which belongs at the same set and both has the same x position **e.g. check the image below.**

screen shot 2016-01-06 at 12 13 20 pm

In the example, we have tree bubbles in the same x-position, I mean 121.2K. Now when I try to select one of them it seems that the API always select the same and I guess is the one in the middle. 

So I suppose that the API only search at the first element of every set wich is near to the tap, but in my case, I need to search to the elements of the set. Do you have any idea how I can do this? or where I can start reading the code just to add the fix?

lidgardo:
- Fix #654 issue

### DIFF
--- a/Charts/Classes/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Charts/Classes/Data/Implementations/Standard/ChartDataSet.swift
@@ -201,52 +201,56 @@ public class ChartDataSet: ChartBaseDataSet
     {
         var entries = [ChartDataEntry]()
         
-        var low = 0
-        var high = _yVals.count - 1
-        
-        while (low <= high)
-        {
-            var m = (high + low) / 2
-            var entry = _yVals[m]
-            
-            if (x == entry.xIndex)
-            {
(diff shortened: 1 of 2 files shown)


danielgindi:
Man you're breaking everything here, an not even doing so in a clean manner.

If the highlight is not working for you - then your dataset is not correct.
You may have an unordered dataset, where in Charts all datasets should be ordered.

We may change that requirement in the future, but currently that helps avoiding calculating and rendering values which are out of viewport.

lidgardo:
Sorry for the dirty way - I just wanted to show that the problem was that Charts only works with ordered data sets, I didn't notice about this constraint.

Thanks in advance.

danielgindi:
It's really ok! Thanks for your efforts!

PR #1094, opened 2016-05-31
fix ChartHighlight init data in function of highlightValue

hanjt:
If dataIndex and dataSetIndex set -1 when initialize ChartHighlight(), it will be crash because of  out of bound of array in function of drawMarkers() and class of  ChartViewBase.


liuxuan30:
I guess we should first check if the fix is valid, as it's a new feature of isHighlightFullBarEnabled
Just change to 0 not solving all problems I guess.

liuxuan30:
Well this issue is more complicated than we thought, so this fix doesn't solve all the issues. Thanks for you commit, however I have to close it, sorry

PR #1099, opened 2016-06-01
The interval sometimes end up having a value of `0`, which makes the app crash

JonathanBouaziz:
The interval sometimes end up having a value of `0`, which makes the app crash


JonathanBouaziz:
ignore it

PR #1126, opened 2016-06-11
Fixed building with latest Carthage due to accidental use of UIColor instead of NSUIColor

einsteinx2:
…


danielgindi:
It seems that you have not pulled everything... As this was fixed before release :-)

einsteinx2:
Whoops you're right, we forgot to pull from upstream.

Btw, was there a technical reason for reverting the 10.10 support pull request? We're using it on 10.10, so if there are any issues that may affect us, I'm interested in hearing them. Thanks.

PR #1145, opened 2016-06-15
Charts changes for Swift 2.3

opswhisperer:
If using cocoapods and Xcode 8 add the following to ensure the framework is set for Swift 2.3. Optionally modify this to only change the Charts pod
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings['SWIFT_VERSION'] = '2.3'
    end
  end
End
```
If its not a swift pod it still gets added but ignored, if it its a Swift pod then it sets the `Use Legacy Swift Language Version` build setting to `Yes`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# OS X
+.DS_Store
+
 # Xcode
 #
 build/
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -953,29 +953,20 @@
 			attributes = {
 				LastSwiftMigration = 0700;
 				LastSwiftUpdateCheck = 0710;
-				LastUpgradeCheck = 0800;
+				LastUpgradeCheck = 0730;
 				ORGANIZATIONNAME = dcg;
 				TargetAttributes = {
 					06AEE7A11BDC3F8B009875AE = {
 						CreatedOnToolsVersion = 7.1;
-						DevelopmentTeam = 4ZBD65B36X;
-						DevelopmentTeamName = "Robert Bell";
-						LastSwiftMigration = 0800;
 					};
 					5BA8EC3F1A9D14DC00CE82E1 = {
 						CreatedOnToolsVersion = 6.1.1;
-						DevelopmentTeam = 4ZBD65B36X;
-						DevelopmentTeamName = "Robert Bell";
-						LastSwiftMigration = 0800;
 					};
 					65B3F5BE1C7370BE000983D0 = {
 						CreatedOnToolsVersion = 7.2.1;
-						ProvisioningStyle = Automatic;
 					};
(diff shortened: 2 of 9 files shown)


liuxuan30:
@danielgindi shall we consider 2.3 or we jump to 3.0 directly? I did see more work for 3.0 while converting. 2.3 is much easier to convert automatically. 2.3 is the version to get started with Xcode 8 very quickly

opswhisperer:
I’d love to go to 3 directly but I have a lot more dependancies that need to move too. 
Theoretically I think you could have a Swift 3 framework in a 2.3 project since they each have their own build settings.

jerryhjones:
There are a lot of project related changes in there that you problem shouldn't include in the pull request. Thinks like...
```
DevelopmentTeam = 4ZBD65B36X;
DevelopmentTeamName = "Robert Bell";
LastSwiftMigration = 0800;
```

opswhisperer:
True, I was being lazy and didn't want to break my project. He's going to have to replace those anyway as its a part of the new code signing system. Normally I'd .gitignore the project file but thats where you have to set the swift version.

liuxuan30:
I compare this branch with my swift 2.3 branch, and actually you are having some info like development team, xcscheme that I don't have in diff. I think it's your .gitignore file issue? Because the one from the library seems filted xcscheme.

opswhisperer:
I would take your 2.3 branch one mine, the Team info in the diff was me trying to get Cocoapods to respect those new settings in Xcode 8 so I pushed them to my Charts repo.

liuxuan30:
I see, I don't use pod for my repo, just the same clone of the upstream branch.

liuxuan30:
@opswhisperer would you fix the irrelevant diffs or I create a new 2.3 merge?

opswhisperer:
I can fix it, might be later today.

opswhisperer:
Updated, left only the swift changes and the build settings recommended by Xcode

liuxuan30:
I am closing this and use #1163 instead.

liuxuan30:
oops, my bad, I just saw you updated your PR.

However, an interesting thing is for ChartPlatform.swift, you and I have different results.. Do you use Xcode 8 converter to do it? Or you manually did that?

liuxuan30:
BTW, please choose Swift-2.3 branch for this PR. 

I am also seething some minor differences, like whole module optimization vs single file optimization.

@danielgindi what you think

opswhisperer:
All of those changes are from the migration tool. I didn’t do any manually

liuxuan30:
closing; we had a swift 2.3 branch for this. I am not sure how long will swift 2.3 will exists, but currently 2.3 and 3.0 branch is for testing iOS 10, so once it is released, I think convert all in once at that time based on master is a better choice instead of continuous merging

PR #1179, opened 2016-06-27
add BarLineChartViewBase.resetZoom()

DivineDominion:
Matches the existing zoomIn()/zoomOut() convenience methods.


liuxuan30:
scale 0? why not 1?

DivineDominion:
Uh ... Didn't try 1 :) 0 Seems to work, too, probably because of a division by 0 guard clause? Will have a look at this later.

PR #1195, opened 2016-06-29
Fix position of value label in bar chart in case of on of the value is 0

### LINKED ISSUE #1191: [Bug] Double BarChart second value text mispaced
I would like to create chart as in screenshot^
screen shot 2016-06-29 at 11 07 39

But can't place label in right position in case of `0`
Please, have a look at screenshot in comment for details

skywinder:
This will fix #1191 🎉 

screen shot 2016-06-29 at 19 03 43

### DIFF
--- a/Charts/Classes/Renderers/BarChartRenderer.swift
+++ b/Charts/Classes/Renderers/BarChartRenderer.swift
@@ -416,7 +416,11 @@ public class BarChartRenderer: ChartDataRendererBase
                                 let value = vals[k]
                                 var y: Double
                                 
-                                if value >= 0.0
+                                if value == 0.0
+                                {
+                                    y = value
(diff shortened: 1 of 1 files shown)


skywinder:
@petester42 please All tests are passed. Travis get error not related to PR: 

>The command "cd Charts && carthage bootstrap" failed and exited with 1 during .

liuxuan30:
would you please rebase with master (and squash if needed), I just fixed travis last week I think

skywinder:
@msanders I just merged latest changes from `master`. 
`rebase` is destructing command, so I prefer to not do it without specaial need.

danielgindi:
This actually introduces new bugs. Fix coming soon

Issue #44, opened 2015-04-23
Use of undeclared type 'CGFloat'


msanders:
Has anyone had any luck getting this to compile for an iOS 7 project? I tried the suggested approach (copying all .swift files to your project), but I'm getting lots of compiler errors from what look to be missing UIKit imports:

```
ios-charts/Charts/Classes/Data/ScatterChartData.swift:19:43: Use of undeclared type 'CGFloat'; did you mean to use 'CGFloat'?
ios-charts/Charts/Classes/Data/ScatterChartData.swift:21:19: Use of unresolved identifier 'CGFloat'
ios-charts/Charts/Classes/Data/BarChartData.swift:18:31: Use of unresolved identifier 'CGFloat'
ios-charts/Charts/Classes/Data/BarChartData.swift:21:28: Use of undeclared type 'CGFloat'; did you mean to use 'CGFloat'?
```

Any idea what's going wrong? I'm using Xcode 6.3.1.

danielgindi:
There are several people who already did... But try to add an import to
UIKit and let me know

vishw3:
Import UIKit will work i tried it few days back

danielgindi:
In do not experience that- so if one of you wants to make a pull request
for iOS 7 compilations with the proper UIKit imports that would be great!

‏בתאריך יום חמישי, 23 באפריל 2015, Vishwas NG <notifications@github.com>
כתב:

vishw3:
ya i wanted to support for iOS 7 so i just copied swift file in to my project and worked around every thing works fine if UIKit is imported to all shift headers

vishw3:
hi i have created example project to run on iOS 7 tested on simulator and working fine look into it
https://github.com/vishw3/IOSChart-IOS-7.0-Support/tree/master

danielgindi:
Merged :-)

zhijunsheng:
I cloned yesterday. Don't understand why I had to manually add "import UIKit" for swift files dragged in. I'm using Xcode 7.3.1 and my own project is an ObjC one. What did I miss?

kaleevans:
I've had the same issue as zhijunsheng. Objective-c project still having this issue.

I am trying the importing swift files method as I had issues just importing the framework with the bridging header.

balkarov:
I have this problem too

PR #1205, opened 2016-07-01
Add missing imports for iOS 7 support

VincentSit:
Same like  #45. This can cause compile failure in iOS 7 like #44. I just ran into this problem.


liuxuan30:
why not import CoreGraphics?

VincentSit:
Updated.

The reason is for consistency with the code style in #45.

I didn't read the source code before, didn't notice the style has been changed.

The 7eb5810 is useless for now, you can colse this PR if you want to keep the commit history clean. I don't think it's necessary to open a new PR.

PR #1245, opened 2016-07-13
improve comment to warn users how to use setVisibleRange APIs

liuxuan30:
Seeing some users don't know how to use setVisibleRange APIs (can only be called after we have data, otherwise no effect). For now, add the API description for this


dxclancy:
Hi!  This is a good idea but i find your change confusing. You description of your change in the parenthesis makes more sense to me. I would say something more like that. "Chart must have data or these have no effect".

liuxuan30:
@dxclancy thanks a lot, not a native speaker, so.. thanks

PR #1258, opened 2016-07-20
Fix Swift 3 compilation errors for iOS

puthirith:

PR #1230, opened 2016-07-06
Supports Xcode 8-beta 2

acegreen:

PR #1266, opened 2016-07-23
RunLoopMood changes for Swift 3.0 & Xcode 8.0 Beta 3

atetlaw:
- RunLoop.main is now a class var instead of a method
- Thread.isMainThread is now a class var instead of a method
- display link method now take the RunLoopMode struct instead of a string

It should be this way:
`_displayLink.remove(from: RunLoop.main, forMode: RunLoopMode.commonModes)`

This way is just redundant:
`_displayLink.remove(from: RunLoop.main, forMode: RunLoopMode(rawValue: RunLoopMode.commonModes.rawValue))`

### DIFF
--- a/Charts/Classes/Animation/ChartAnimator.swift
+++ b/Charts/Classes/Animation/ChartAnimator.swift
@@ -71,7 +71,7 @@ public class ChartAnimator: NSObject
     {
         if (_displayLink != nil)
         {
-            _displayLink.remove(from: RunLoop.main(), forMode: RunLoopMode.commonModes.rawValue)
+            _displayLink.remove(from: RunLoop.main, forMode: RunLoopMode.commonModes)
             _displayLink = nil
             
             _enabledX = false
@@ -196,7 +196,7 @@ public class ChartAnimator: NSObject
         if (_enabledX || _enabledY)
         {
             _displayLink = NSUIDisplayLink(target: self, selector: #selector(ChartAnimator.animationLoop))
-            _displayLink.add(to: RunLoop.main(), forMode: RunLoopMode.commonModes.rawValue)
+            _displayLink.add(to: RunLoop.main, forMode: RunLoopMode.commonModes)
         }
     }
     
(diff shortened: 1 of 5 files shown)


liuxuan30:
There are two PRs about swift 3.0 - #1230, #1258 pending merge.
One question: why you have changes in ChartPlatform.swift, while those two  don't?

atetlaw:
Hi @liuxuan30, It contains the `NSUIDisplayLink` class. That class had an API the same as `CADisplayLink`. The old API took a `String` for the `runLoop` arg. I changed it to take the `RunLoop`.

liuxuan30:
@atetlaw what's the purpose of changing to `RunLoopMode`?

atetlaw:
A couple of reasons,

- matches the `CADisplayLink API
- avoids the "stringiness"

On the stringliness point: the old version took a string, which meant that to call it you'd need to pass `RunLoopMode.commonModes.rawValue`, and then in the `NSUIDisplayLink` class it took the string and created a `RunLoopMode` via `RunLoopMode(rawValue: stringArg)`.

It seemed redundant to do that, when passing `RunLoopMode.commonModes` provides addition compile-time protection that you're passing the right type to the method.

liuxuan30:
@atetlaw ok, I have to let @danielgindi take a look at this, not sure why he uses string at first place

atetlaw:
👍 no problem

danielgindi:
@liuxuan30 Could you please take all of those issues & PRs regarding Swift 3 and leave only one? It will be finalized after the Swift 3 release, as Swift changes all the time... And will probably sneak in Charts v3 release...

liuxuan30:
@danielgindi sure, but I need you to take a look at this PR regarding `RunLoopMode`, is it right to use `RunLoopMode` instead of 'string'?

Current swift 2.3 and swift 3.0 branchs are suited for test purpose only, once Swift 3.0 is released, it is much easier for us to convert directly at that time than merging these branches.

However I observed people using the same Xcode beta can generate slightly different PRs.. So I am careful with these differences

liuxuan30:
I'm not sure if this is a mistake while merging Swift 3.0 code.. It seems intended to use `string` for NSUIDisplayLink and later convert to RunLoopMode for UIDisplayLink.
I will look into this this weekend.

liuxuan30:
I looked into this, it's valid. Since it has conflicts, I have manually merged it in https://github.com/danielgindi/Charts/pull/1296

PR #1281, opened 2016-08-02
Few more changes needed to build with Swift 2.3

EpicDraws:
Xcode 8 beta 3 and 4 needed this to build.


liuxuan30:
Thanks.
Please notice, swift beta branch is used for test only. After Swift 2.3(maybe 3) released, we will delete the branch and convert directly at that time, which is much easier

PR #1284, opened 2016-08-02
ChartMarker subclassing in Objective C

dyeazel:
I'd like to propose changing ChartMarker to using a protocol so that it can be subclassed in Objective-C. I've added the ChartMarkerProtocol and modified ChartMarker to implement this protocol.

### DIFF
--- a/Charts/Charts.xcodeproj/project.pbxproj
+++ b/Charts/Charts.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


danielgindi:
In v3 it is already solved :-) But v3 is in progress so please hold on!

fundon:
@danielgindi BTW, Does it support for Swift 3.0 in v3?

danielgindi:
Not yet. v3 is the official Charts 3.0 branch until a Charts 3 release and will be merge to master. When Swift 3 will be a final release - it will be merge to master too.

PR #1335, opened 2016-08-18
Fixes for latest Swift 3.0 Changes as of Xcode 8 beta 6

druidsbane:
Hi,

I've made changes to get the project to compile under the latest Xcode beta.  The only thing I haven't updated is the binary for the RealmSwift project so the RealmSwift project won't link properly, but I wasn't sure where to get the latest binary and as compatibility will continue changing I didn't think it was worth adding in for this request.

Do you mind taking a look to see if it is worth merging?  Hopefully that should reduce the need to keep making these changes every time a new release comes out.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-osx_image: xcode7.3
+osx_image: xcode8
 language: objective-c
 
 env:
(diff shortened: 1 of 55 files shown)


liuxuan30:
thanks @druidsbane! sorry that I will use https://github.com/danielgindi/Charts/pull/1342 for xcode 6 beta.. there are several PRs for swift 3.0.

PR #1336, opened 2016-08-19
IBDesignable crash because of Animator

lunalugo25:
I create a subclass of LineChartView and this subclass is loaded in a IBDesignable UIView (CustomView), so when I add a UIView on a StoryBoard and subclass from my CustomView it crash because it's not capable to create '_displayLink'.

### DIFF
--- a/Charts/Classes/Animation/ChartAnimator.swift
+++ b/Charts/Classes/Animation/ChartAnimator.swift
@@ -195,7 +195,7 @@ public class ChartAnimator: NSObject
         
         if (_enabledX || _enabledY)
         {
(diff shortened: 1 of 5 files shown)


liuxuan30:
well, are you saying it's about swift 2.3 issue? I saw you want to merge into 2.3, but I'm not sure which part the issue resides on

danielgindi:
This PR does not really describe what the changes are, and it contains various commits - not by the author of the PR.
If you want to create a PR, please create one against the latest `master`, with clean and clear code...

danielgindi:
Oh I think I get it

lunalugo25:
Yes, danielgindib that was the issue

PR #1342, opened 2016-08-20
Conforms to Swift 3 - Xcode 8 beta 6

acegreen:


acegreen:
@liuxuan30 there could be a few errors in this. I would go over with a fine-tooth comb. I noticed my Pie chart hole is no longer transparent. Looking into that

liuxuan30:
are you saying there are issues with the library or xcode beta?

acegreen:
@liuxuan30 No the changes introduced because of beta 6 might have caused some small errors. I noticed my pie chart hole is no longer transparent for example. Not sure if this is related to the changes or from before. All I'm saying is we should go through this before merging it.

liuxuan30:
well.. I assume the conversion tool will take care of all of them. Why you said copy/paste mistakes? Downloading beta 6...

acegreen:
I do most of my changes manually to make sure its the intended change and for example
context.move(to:) now takes a CGPoint, so I had to change it to one and in some cases MIGHT have not gotten right. Again I'm being overly cautious with my statements. Could also be that everything is 100%. So again to summarize, I didn't catch any mistakes in my review but there could be.

Charts/Classes/Renderers/CandleStickChartRenderer.swift
```diff
@@ -43,7 +43,7 @@ public class CandleStickChartRenderer: LineScatterCandleRadarChartRenderer
         }
     }
     
-    private var _shadowPoints = [CGPoint](repeating: CGPoint(), count: 4)
+    private var _shadowPoints = [CGPoint](repeating: CGPoint(), count: 2)
```
acegreen: @liuxuan30 this is a mistake. Surprising since no change needed to be made on that line

Charts/Classes/Renderers/LineChartRenderer.swift
```diff
@@ -709,10 +689,10 @@ public class LineChartRenderer: LineRadarChartRenderer
                 {
                     // Begin path for circle with hole
                     context.beginPath()
-                    context.addEllipse(inRect: rect)
+                    context.addEllipse(in: rect)
                     
                     // Cut hole in path
-                    context.addArc(centerX: pt.x, y: pt.y, radius: circleHoleRadius, startAngle: 0.0, endAngle: CGFloat(M_PI_2), clockwise: 1)
+                    context.addArc(center: CGPoint(x: pt.x, y: pt.y), radius: circleHoleRadius, startAngle: 0.0, endAngle: CGFloat(M_PI_2), clockwise: true)
```
acegreen: Assumed clockwise: true is the same as clockwise: 1

Charts/Classes/Renderers/CandleStickChartRenderer.swift
liuxuan30: such changes should be included I guess, if not converted by Xcode 8. I would only use the conversion tool to do the job, since it's all about swift 3.0, not about bugs or issues. If you manually changed something, with so many diff in one PR, it's hard to find out.

Anyway, swift 3.0 branch is just for testing

Charts/Classes/Renderers/LineChartRenderer.swift
```diff
@@ -771,11 +750,11 @@ public class LineChartRenderer: LineRadarChartRenderer
                 context.setLineWidth(set.highlightLineWidth)
                 if (set.highlightLineDashLengths != nil)
                 {
-                    context.setLineDash(phase: set.highlightLineDashPhase, lengths: set.highlightLineDashLengths!, count: set.highlightLineDashLengths!.count)
+                    context.setLineDash(phase: set.highlightLineDashPhase, lengths: set.highlightLineDashLengths!)
                 }
                 else
                 {
-                    context.setLineDash(phase: 0.0, lengths: nil, count: 0)
+                    context.setLineDash(phase: 0.0, lengths: [])
```
acegreen: Assumed [] would suffice. setLineDash expects a value for length now.

Charts/Classes/Renderers/CandleStickChartRenderer.swift
acegreen: Not sure what you are saying. That change is simply a type, I removed count: by accident and put it back and instead of 4, I put it back as 2. For other changes, I don't use the tool because with Swift 3 it causes of syntax issue. Also many in this beta were not suggested.

liuxuan30: hmm, it seems get complicated... I mean the conversion tool will not change count 4 to 2 right? It seems like a manual change. I am asking why we change from 4 to 2.
Also, why you said "many changes in this beta were not suggested" ? I am still downloading, but I need to work then.

acegreen: Yea give the conversion tool a try then you will understand. Anyways in this instance its simply a mistake. I can submit another commit with the fix once we go over the rest, to make sure there aren't any other mistakes.

Charts/Classes/Components/ChartYAxis.swift
```diff
@@ -254,7 +254,7 @@ public class ChartYAxis: ChartAxisBase
             return ""
         }
         
-        return (valueFormatter ?? _defaultValueFormatter).string(from: entries[index])!
+        return (valueFormatter ?? _defaultValueFormatter).string(for: entries[index])!
```
liuxuan30: I am not sure if we need to use for (default imp will crash) or use NSNumber. Have you test it yet?

Charts/Classes/Renderers/CandleStickChartRenderer.swift
liuxuan30: @acegreen I now understand. However, I think `_shadowPoints` count should be 4. not 2. Code below uses 4 of them.

Charts/Classes/Renderers/PieChartRenderer.swift
```diff
@@ -554,17 +541,17 @@ public class PieChartRenderer: ChartDataRendererBase
...
                         width: secondHoleRadius * 2.0,
                         height: secondHoleRadius * 2.0))
-                    context.addEllipse(inRect: CGRect(
+                    context.addEllipse(in: CGRect(
                         x: center.x - holeRadius,
                         y: center.y - holeRadius,
                         width: holeRadius * 2.0,
                         height: holeRadius * 2.0))
-                    context.eoFillPath()
+                    context.fillPath()
```
liuxuan30: according to CGContextEOFillPath, the rule is even-odd fill rule. However fillPath() seems taking default rule which is winding. I suspect we need check this.

Charts/Classes/Data/Implementations/Standard/ChartDataEntry.swift
```diff
@@ -38,30 +38,32 @@ public class ChartDataEntry: NSObject
         self.xIndex = xIndex
     }
     
-    public init(value: Double, xIndex: Int, data: AnyObject?)
+    public init(value: Double, xIndex: Int, data: Any?)
```
liuxuan30: I am not sure if this is the correct way to use Any. I am confused what Swift is trying to do here about Any and AnyObject. I know `isEqual` method in beta 6 uses Any type, however simply use `self.data = data as AnyObject` seems lame to me..

liuxuan30:
@acegreen hey I just spend several hours on this. Thanks for your work, I know it's painful and requires patience.

There are several issues we need to address (also in line note):

1. public init(value: Double, xIndex: Int, data: Any?) in ChartDataEntry
2. eoFillPath(), also not sure if this is related to your pie chart issue
3. string(for:)
4. private var _shadowPoints = [CGPoint](repeating: CGPoint(), count: 2)

You could continue commit in this PR

pixelspark:
Great work @acegreen! Unfortunately some build issues when building for OS X as well (many of which are simple fix its, it appears). Might have time to submit a PR later this week (if you haven't applied the fix-its already yourself by then :))

Charts/Classes/Renderers/CandleStickChartRenderer.swift
acegreen: haha ITS MY MISTAKE :P, I removed the count and when I added it, I added it back as 2. IT SHOULD BE 4. I'll update it in the next commit

acegreen:
@liuxuan30 Yea but we should be able to sort it out. 
1. Can you elaborate on this?
2. I wrongly assumed fillPath() is equivalent to eoFillPath() but it should be
fillPath(using: .evenOdd). I changed 4 instances back to that. 
3. Reverted back to string(from: ) but casted values as NSNumber
4. Corrected back to count: 4

Charts/Classes/Data/Implementations/Standard/ChartDataEntry.swift
acegreen: I think it is. Its proposal [SE-0116](https://github.com/apple/swift-evolution/blob/master/proposals/0116-id-as-any.md)

Charts/Classes/Components/ChartYAxis.swift
acegreen: I changed it back to from: as NSNumber in the new commit

Charts/Classes/Renderers/PieChartRenderer.swift
acegreen: Corrected in new commit

liuxuan30:
@acegreen I am going to merge this PR soon, but a few more comments:

for 1, I am just saying using `Any` and later `self.data = data as AnyObject` seems weird.. but you give me SE0116, so let's move forward.

for 2, have you investigated .evenOdd or winding, and relates to your pie chart issue?
for 3, my beta 6 told me to use `formatter.string(from: NSNumber(val))` instead of `as`, though they look equivelant, so I am fine.

acegreen:
2. I investigated that and it turns out there is an .evenOdd option in fillPath. Which is what i should have used in the first place for all eoFillPath() (eo = evenOdd) but instead I used just fillPath(). Second commit reverts back to using evenOdd with the Swift 3 syntax
3. Both are the same, though Apple recommends casting "as" for dealing with Swift to objc equivalent. I think it's SE-0072 https://github.com/apple/swift-evolution/blob/master/proposals/0072-eliminate-implicit-bridging-conversions.md
Using .evenOdd addresses the pie chart issue and is back to the way it was (which is good)

liuxuan30:
great, merged. Thanks!

PR #1355, opened 2016-08-25
Fix two transcription errors related to horizontal cubic drawing.

MrMage:
These just make the code in-line with MPAndroidChart (see https://github.com/PhilJay/MPAndroidChart/blob/2b04973813b7012d108c4da819b7c1b689b77557/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/LineChartRenderer.java).

Screenshot of the previous rendering:

simulator screen shot 25 aug 2016 14 09 44

Fixed rendering:
simulator screen shot 25 aug 2016 14 10 52

### DIFF
--- a/Charts/Classes/Renderers/LineChartRenderer.swift
+++ b/Charts/Classes/Renderers/LineChartRenderer.swift
@@ -195,14 +195,14 @@ public class LineChartRenderer: LineRadarRenderer
             if cur == nil { return }
             
             // let the spline start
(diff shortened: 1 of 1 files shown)


danielgindi:
Actually these are errors that were introduced while getting rid of the x-indexes and moving to Charts 3.0. This piece of code was originally written for iOS, but a semicolon for the `CGFloat` cast  messed it up :-) 
Anyway thanks for noticing, but I've just fixed this moments ago

PR #1376, opened 2016-08-30
Swift 3.0

JonathanBouaziz:
Hello,
I've fixed a couple warnings but more importantly updated the ChartMarket and its properties from `public` to the new swift type `open`. That allows subclassing outside the module.


liuxuan30:
Thanks. But I need to hold on this a little while, maybe we will replace public with open for all available classes.

liuxuan30:
I manually merged this PR, thanks! Now all available classes func and vars are now open

PR #1387, opened 2016-09-01
migrate Chart v3 code in master to swift 2.3 syntax

liuxuan30:
migrate Chart v3 code in master to swift 2.3 syntax


liuxuan30:
use github 'squash and merge' option gives me more trouble. use #1389 as a clean shot

PR #1388, opened 2016-09-01
migrate Chart v3 code in master to swift 2.3 syntax

liuxuan30:
migrate Chart v3 code in master to swift 2.3 syntax


liuxuan30:
this mess up #1387, I reseted swift 2.3 branch and use #1389 for a clean shot

Issue #1326, opened 2016-08-16
Error with carthage update ChartsRealm fails to build


lewissk:
When running a carthage update --platform iOS ChartsRealm fails to build.  I have tested this on 3 of our machines with the same result.  We have cleaned the project deleted the Carthage directory tried downgrading the version.  All with the same result.

Here is the output of the failure:

The following build commands failed:
	Ld /Users/user/Library/Developer/Xcode/DerivedData/ChartsRealm-admcvrfsqdqvoxfzhmelrhhyzqfx/Build/Intermediates/Charts.build/Release-iphoneos/Charts-OSX.build/Objects-normal/armv7/Charts normal armv7
	Ld /Users/user/Library/Developer/Xcode/DerivedData/ChartsRealm-admcvrfsqdqvoxfzhmelrhhyzqfx/Build/Intermediates/Charts.build/Release-iphoneos/Charts-OSX.build/Objects-normal/arm64/Charts normal arm64
(2 failures)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
A shell task (/usr/bin/xcrun xcodebuild -project .../Carthage/Checkouts/Charts/ChartsRealm/ChartsRealm.xcodeproj -scheme ChartsRealm-iOS -configuration Release -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build) failed with exit code 65:
** BUILD FAILED **

The following build commands failed:
	Ld /Users/user/Library/Developer/Xcode/DerivedData/ChartsRealm-admcvrfsqdqvoxfzhmelrhhyzqfx/Build/Intermediates/Charts.build/Release-iphoneos/Charts-OSX.build/Objects-normal/armv7/Charts normal armv7
	Ld /Users/user/Library/Developer/Xcode/DerivedData/ChartsRealm-admcvrfsqdqvoxfzhmelrhhyzqfx/Build/Intermediates/Charts.build/Release-iphoneos/Charts-OSX.build/Objects-normal/arm64/Charts normal arm64
(2 failures)

Cunqi:
same here, get the same error

calvingit:
[...]

Issue #1119, opened 2016-06-09
Extracting ChartsRealm to a separate project


eshurakov:
I think it would be nice to separate Charts and ChartsRealm into two projects.

There are several issues with having them in the same project:
1. Building with carthage takes a lot of time as it builds both Charts and ChartsRealm. For me it also randomly fails when compiling ChartsRealm.
2. Big repository size as Realm framework is checked in

If this is split into two projects, ChartsRealm could have a dependency on Charts (via carthage or as a submodule) and on Realm.

liuxuan30:
they are two projects on master, could check if I am wrong?

eshurakov:
Sorry, my bad, I meant separate repository. 
The thing with carthage is that it builds every shared scheme in the repository, which now includes both Charts and ChartsRealm frameworks. Of course, this sounds like a problem that carthage should solve, I still feel like ChartsRealm deserves its own repository. But I can also see a problem that questions about charts will most likely be asked both here and there.

Regarding my initial issue with carthage, as a workaround I forked Charts repository and removed Realm so it doesn't take time to compile code that I don't use.

ahknight:
Agreed.  This is the best solution all around.  One repo, one project. (Having similar issues with Carthage, myself.)

liuxuan30:
Putting ChartsRealm as a separate repo seems a overkill? It is just a small part of the Data series

ahknight:
Perhaps, but including it confuses a major package manager and a good segment of the user base.

liuxuan30:
alright, guess we will need @danielgindi review this

liuxuan30:
also seems related to #1118 ?

danielgindi:
Truth is I really want to do this - so I won't have to include binaries of Realm in the project.
I don't mind keeping it in the project or moving to a separate repository - as long as regular users can compile the project (and run the demos too) right away without errors, 
And Realm users can use it easily without a hassle.

(Keeping Carthage/Pods compatibility is nice too)

I'm open to suggestions here on how to do this!

@petester42 do you have any idea how to do this?

petester42:
I have ideas on how to make this better so binaries are not included. It won't address the first point in the issue though.

danielgindi:
Well let's hear it :-)

petester42:
Basically what I would do is what I did for (this project)[https://github.com/Moya/Moya]. One project with all the targets. Carthage to download the dependencies for the project. Cocoapods shouldn't really change because of this.

[...]

PR #1406, opened 2016-09-07
Remove Realm binaries from project and fix carthage building

petester42:
This change address two issues having to do with Carthage (#1326 and #1119). 

This first change is removing the prebuilt binaries inside of the ChartsRealm project and replacing them with ones that come from Carthage when running "carthage bootstrap" or "carthage update". There has been a lot talk in #1119 about the possible solutions and to me this seems to be the best for now. This means that contributors wanting to update the ChartsRealm project will have to have Carthage install and run "carthage bootstrap" to get the Realm dependencies. A note in the CONTRIBUTING.md should be added for newcomers. 

The second change sort of goes hand in hand with the first. It changes how ChartsRealm requires Charts.framework in order to work with Carthage. For whatever reason when building using Carthage most of the time it will pick the macOS version of Charts.framework (but not all of the time, its weird). So instead of having the ChartsRealm project depend on Charts.framework from the Charts project it now depends on Charts.framework built by Carthage.

I believe this addresses most of the Carthage issues. It seems to work consistently now.


petester42:
This should be good for review now @danielgindi or @liuxuan30. Let me know what you think.

liuxuan30:
seems still have dependency for Charts.framework to require Realm?

petester42:
I'm not sure I understand what you mean. But I think I'm gonna close this. I found a better way.

liuxuan30:
never mind :)

petester42:
close in favour of #1422

PR #1422, opened 2016-09-09
Update project structure for simplicity and fixing carthage

petester42:
A while ago a came across a [post](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/) that greatly simplifies how many targets we need for multiple platforms. So basically the removes the need for a different target for each platform so now there is only 2, Charts and ChartsRealm.

This change address two issues having to do with Carthage (#1326 and #1119).

This first change is removing the prebuilt binaries inside of the ChartsRealm project and replacing them with ones that come from Carthage when running "carthage bootstrap" or "carthage update". There has been a lot talk in #1119 about the possible solutions and to me this seems to be the best for now. This means that contributors wanting to update the ChartsRealm project will have to have Carthage install and run "carthage bootstrap" to get the Realm dependencies. A note in the CONTRIBUTING.md should be added for newcomers.

The second change sort of goes hand in hand with the first. It changes how ChartsRealm requires Charts.framework in order to work with Carthage. Having the two targets inside the same project and having a Carthage.xcconfig telling carthage where to look for dependencies seems to address the Carthage issues that have been seen. 

I believe this addresses most of the Carthage issues. It seems to work consistently now. And it is now more streamlined, 1 target for all platforms!


petester42:
@danielgindi, @liuxuan30 what do you guys think? I like this a lot and it makes contributing much easier. 1 project, 1 target, simple.

liuxuan30:
to confirm, I just check out the branch, and ChartsDemo needs to run "carthage bootstrap" to get Realm and continue compile right?
I simply tried to build Realm and warns me no Realm framework

petester42:
ya anything realm based will need a carthage bootstrap. everything else should work as before

liuxuan30:
guess let @danielgindi review the structures. 
BTW, seeing old directories like Charts/classes/Charts/ is empty, actually they are in source/. Do we need to delete the empty directories?

petester42:
run `git clean -xfd` they will be gone. You just have old stuff

liuxuan30:
git clean work on my side, thanks

Cramsden:
Is this pull request going to be accepted? Would love to be able to use Charts without Realm dependancies.

danielgindi:
@petester42 
I apologize, being busy with other stuff I didn't look into this branch.
Any chance that you rebase it on `master`?

Also - will people be able to build the project without Carthage?
Many developers seem to be unable to read simple instructions, or open an Xcode project and hitting "Run", so I want to avoid all those "Issues" from being created in the first place ;-)

petester42:
Ya will do no problem. 

As far as just hitting run, for Charts its fine. For ChartsRealm there is a carthage dependency and as discussed in #1119 there is no real good solution. So contributors will have to run `carthage bootstrap` to contribute to ChartsRealm. Seeing as ChartsRealm is just a data layer implementation I think its ok for now. 

I might create scripts in the future to simplify the process so the contributor will just run an init script and have the project setup automatically done for them.

petester42:
I'm gonna merge this to make the swift 3.0 version ready.

danielgindi:
👍 
Do you want to update the README to tell people how to handle this?

danielgindi:
By the way, Realm should point to `master` branch, not to a release, as only `master` support Swift 3.0

liuxuan30:
Realm 1.1.0 should be Swift 3.0 now

PR #1437, opened 2016-09-14
Chart2.2.5 swift2.3

ScottishFold:
I put a barChartView in a scrollView,then I move up barChartView,the position of touchPoint to highlight the barChartView is not right,it's above the barChartView,is there any solution to solve this problem?Thank you.


liuxuan30:
have no idea what are you saying and branch mess up. closing.

Issue #1462, opened 2016-09-18
(v3.0)There is no Chart-iOS framework in Product


Huang-Libo:
I need to import chart's iOS framework to my project, but I can only found chart's mac framework. In previous version, there has a Charts-iOS in Charts.xcodeproj, but can not be found in v3.0
<img width="823" alt="2016-09-18 1 31 32" src="https://cloud.githubusercontent.com/assets/12227130/18613325/9f9bb702-7daa-11e6-94aa-8de19f97532b.png">

MonkeyRing:
I also found this situation. @liuxuan30

SvenMuc:
You can change it in the project setting → Architectures → BaseSDK = Latest iOS (iOS 10.0.1)

But I'm struggling with a lot of other warnings and errors. It seems that in Charts v3 a lot has been changed. The migration path is not always clear. Is it planed to prepare some migration guidelines?

liuxuan30:
@SvenMuc in master branch, it's already swift 3.0 and chart 3.0. main changes are for how x axis is rendered. Daniel do have a plan to write something, but too many things recently..

SSSakura:
same problem..

SvenMuc:
@liuxuan30 thanks! Swift 3.0 ist not a big issue. I already migrated my project successfully. As you mentioned I have mainly problems with the x axis and the value formatter (around 180 errors).
Has the demo already been migrated to Charts v3 and Swift 3.0? Maybe I can use this two analyse the changes. Xcode trys to auto migrate the code.

After my migration I can provide here some notes, what changes are required.

danielgindi:
`master` branch **is already converted to Swift 3.0**. Please pull master, instead of trying to convert it yourself.

petester42:
There is only one framework target for all platforms now. Just include the macOS one and it will also work for iOS and tvOS.

PR #1467, opened 2016-09-19
change Charts baseSDK to iOS 10

liuxuan30:
https://github.com/danielgindi/Charts/issues/1462 change Charts baseSDK to iOS 10


petester42:
Did Xcode tell you to do this? You shouldn't override the sdkroot it has been set in the global project settings on purpose.

liuxuan30:
not really, it's just #1462 has asked this. I did a quick test, changing baseSDK not impacting OS X demos.

PR #1484, opened 2016-09-19
Don't override project settings in targets

petester42:
Project settings don't need to be set to iOS. Due to having a single target they need to be macOS which has been set in the project's main settings.

Please read the article [here](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/) for more info.


liuxuan30:
great to know, thanks!

liuxuan30:
but one more thing, how about https://github.com/danielgindi/Charts/issues/1462 question? Not seeing the framework for iOS

petester42:
The demo works. If you include it like that it works. If you use Carthage then it will build all the types.

PR #1485, opened 2016-09-20
Chart2.2.5 swift2.3

pockyzhang:


liuxuan30:
master is with Chart 3.0 + Swift 3.0 already. invalid PR

PR #1509, opened 2016-09-21
Fix Cocoapods setup being broken

petester42:
While working of the v2.3.0 release I noticed that there were problems with our podspec because we used a prepare_command. This is not really necessary and the same can be achieved with a compiler flag in the xcode project.

Changes:
- Updates to cocoapods 1.1.0.rc.2 which supports xcode 8
- Sets the swift version to 3.0 in ".swift-version" so cocoapods can use the right version
- Uses a "NEEDS_CHARTS" compiler flag instead of a prepare_command (xcode project will have the flag set so that it uses `import Charts` while cocoapods will not so it will be ignored)
- Bumps the version of the podspec to 3.0.0

If you have problems linting or pushing the pod to trunk you need to set an environment variable in order to get around some problem's with Realm's podspec due to cocoapod issues. To fix this run `export FORK_XCODE_WRITING=true` and everything should work.


danielgindi:
For me - you're the cocoapods-wiz, I know nothing about it except that it messes with my project settings too much ;-)
So if you're waiting for me to review, go ahead and merge...

petester42:
It was more to inform everyone about what was going on.

PR #1510, opened 2016-09-21
Attempt to make CI more stable

### LINKED ISSUE #1508: Fix Intermittent CI Failures
CI seems to fail every second time. This is probably due to simulators not being loaded in time. Will look into a solution.

petester42:
Fixes #1508

From others around it seems that starting the simulator before building and giving it time to load fixes most of these errors we are seeing on CI.

The UUIDs of the simulators were taken from the travis build log in case anyone is wondering.


danielgindi:
And "All checks have passed" :-)
I would rebase this a few times, just change something in the commit name, see if travis consistently passes

petester42:
Ya we should leave this open for a while and test it more

petester42:
5 passes in a row I think this is good!

PR #1512, opened 2016-09-21
Chart2.2.5 swift3.0

kansalmohit19:


petester42:
Swift 3.0 will only be for charts v3. See charts 2.3.0 for swift 2.3 support.

PR #1541, opened 2016-09-23
PieChart highlighting animation

marcin-zbijowski:
I added animation of highlighting all values of Pie Chart.
I also added couple options to that:
- drawing shadows underneath each slice in the same color as the slice itself but with alpha, width configurable blur and offset
- separate inner radius value for highlight
- separate slice space value for highlight.

`Animator` class is refactor to allow adding new animation dimensions easily.


danielgindi:
Hi! 
You have created a PR against the 2.3 branch, which is not being developed but is there for easying the transition for Swift 2.3 users.
Please rebase on `master` :-)

jjatie:
Closing. If you still want this feature merged, please rebase to 4.0.0

kalmurzayev:
I have a solution (not the best one, but very fast one for sure):
you can actually have 2 pieChartViews (exact copies), one underneath another. Call them A-chart (in front) and B-chart (backstage)
So when you will need to highlight a slice N with animation with shift X, then you do following:
- edit backstage pieChartView's dataset, so that all its entry colors  except for one of slice N become `UIColor.clear`
- you will end up with a B-chart with N-1 slices invisible and slice #N of its color
- start animating B-chart's size, so that its radius becomes 10px greater

This is a very brief explanation, but you get the point.
I did that ONLY because my PieChartView is not spinnable and not highlightable on tap.
I do not recommend this solution for fully functioning pieChart

Issue #1540, opened 2016-09-23
Crash while using Combined Chart


vaibhav-varshaaweblabs:
Charts crashes while populating the dataset 

`let set: BarChartDataSet = BarChartDataSet.init(yVals: userTimesData, label: "User Times")`

Whole function - 

```
func populateBarChartData(userTimes: [Double?]) -> BarChartData {
        let barChartData: BarChartData = BarChartData()
        
        var userTimesData : [ChartDataEntry] = [ChartDataEntry]()
        
        for i in 0 ..< userTimes.count {
            userTimesData.append(ChartDataEntry(value: userTimes[i]!, xIndex: i))
        }
        
        let set: BarChartDataSet = BarChartDataSet.init(yVals: userTimesData, label: "User Times")
        set.axisDependency = .Left
        set.setColor(UIColor.greenColor().colorWithAlphaComponent(0.5))
        set.highlightColor = UIColor.whiteColor()
        
        var dataSets : [BarChartDataSet] = [BarChartDataSet]()
        dataSets.append(set)
        
        barChartData.dataSets = dataSets
        
        return barChartData
    }
```

liuxuan30:
you say it crashes but why don't you paste the crash log?

petester42:
This seems like it could be the same issue as #1550 which has been addressed by #1558. Is userTimes an empty array when it crashes?

Issue #1491, opened 2016-09-20
(v3.0 + Swift3.0)LineChart crash when dataSets is a empty array


Huang-Libo:
in v2.x,  when LineChart's dataSets  is a empty array it will show a charts with no line, but v3.0 will crash.

 error info：
fatal error: Double value cannot be converted to Int because it is either infinite or NaN

danielgindi:
Try with latest master, there's a commit from a few hours ago that is supposed to fix that

Huang-Libo:
@danielgindi 
I noticed the commit that you mentioned, and my version is just updated, but still crash.

MonkeyRing:
<img width="668" alt="2016-09-21 5 30 56" src="https://cloud.githubusercontent.com/assets/8415820/18705646/9753dccc-8021-11e6-931b-bf938c1b0650.png">
master  -> Charts 3.0

danielgindi:
This is fixed now due to another issue that has pointed out the source of the issue (#1511)

apurva:
I can confirm that the fix does NOT work. Stepping through the code, I can see that the Number value 
starts at 1.7976931348623157E+308 and goes over the protection.  However, in this case,

`let i = roundToNextSignificant(number: Double(number))`

results in making i = infinity and then 

`Int(ceil(-log10(i))) + 2`

goes on to crash the app.

Why the initial number value is the max double value is beyond me though.

liuxuan30:
@danielgindi reopened

petester42:
Same issue as #1550. Should be fixed by #1558.

quicklywilliam:
This issue is still at large. You can see the illegal conversion on this line (it tries to convert it to an Int inside the call to roundToNextSignificant):

https://github.com/danielgindi/Charts/blob/master/Source/Charts/Renderers/AxisRendererBase.swift#L125

quicklywilliam:
If anyone is looking for a workaround for this without patching, just set custom axis min and max before setting the data on the chart, i.e.:

```
if (dataArray.count == 0) {
                chartView.xAxis.axisMinimum = 0
               chartView.xAxis.axisMaximum = 10
                
                chartView.leftAxis.axisMinimum = 0
                chartView.leftAxis.axisMaximum = 10
                chartView.rightAxis.axisMinimum = 0
                chartView.rightAxis.axisMaximum = 10
}
```

semireg:
We're seeing this only when the chart is zoomed and we empty the data.

PR #1558, opened 2016-09-25
Fix decimals crash in ChartsUtil

### LINKED ISSUE #1550: fatal error: Double value cannot be converted to Int because it is either infinite or NaN
When I use LineChart APIs, I am getting fatal error at ChartUtils.swift @ line 53.

`fatal error: Double value cannot be converted to Int because it is either infinite or NaN`

```
    internal class func decimals(_ number: Double) -> Int
    {
        if number.isNaN || number.isInfinite || number == 0.0
        {
            return 0
        }
        
        let i = roundToNextSignificant(number: Double(number))
        **return Int(ceil(-log10(i))) + 2**
    }
```

petester42:
Closes #1550, #1540, #1491

I added some tests based on the crash mentioned in #1550 and there was a crash when using DBL_MAX as the input of the `decimals` function. 

There are still some abnormalities in the `decimals` function I believe. In MPAndroidCharts I ran the same tests and it returns 2 not zero for all these cases except for the ones that crash and return negative Int max.

This pull request fixes the crash but I don't think the `decimals` function does what we want it to do.

@danielgindi, @PhilJay what is this function actually supposed to do?

### DIFF
--- a/Charts.xcodeproj/project.pbxproj
+++ b/Charts.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		06165F2F1D8110E600722320 /* Charts.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 06165F241D8110E600722320 /* Charts.framework */; };
+		06761B5E1D98176000D5CE09 /* ChartUtilsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06761B5D1D98176000D5CE09 /* ChartUtilsTests.swift */; };
 		06B11FAF1D811B7000D14B02 /* Charts.h in Headers */ = {isa = PBXBuildFile; fileRef = 06B11FAE1D811B7000D14B02 /* Charts.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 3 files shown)


danielgindi:
It's supposed to calculate a default number of decimals to show after the decimal point

danielgindi:
It's not really important what the value is on edge cases. It is used as a default for formatters that were not configured by the user, it just shouldn't crash :-)

petester42:
The main reason I was asking what it does is that I believe it does not do what it's supposed to. See the `testDecimalWithNormalValue` test case. Shouldn't it return 6? If not then why are we doing some complex math if it can only be 0 or 1.

umairnow:
This issue is still there in AxisRendererBase.swift on line 125
fatal error: Double value cannot be converted to Int because it is either infinite or NaN

acegreen:
@petester42 @danielgindi this isn't resolved yet.. 

```
        let yRangeBuffer: Double = 50000
        lineChartView.xAxis.axisMinimum = set1.yMin - yRangeBuffer
        lineChartView.xAxis.axisMaximum = set1.yMax + yRangeBuffer
```

danielgindi:
Guys, the reason that this is not fixed, is that we do not see this error.
So you would have to:

1. Make sure you are doing everything correctly
2. Create steps for us to reproduce
3. Maybe do a little debugging. Why does a `NaN` or `Infinity` get passed to that function?

Issue #1564, opened 2016-09-27
[Chart 3.0 BUG] xAxis labels disappear after a deep zoom in


liuxuan30:
xAxis labels disappear after a deep zoom in
Step to reproduce:
1. Open ChartsDemo Time Line Chart
2. zoom in a lot, until the xAxis label disappears

After triage, I found there are two issues:
1: `centerAxisLabelsEnabled` will be false after entryCount = 1.
Change to below, xAxis.entries will have only 1 entry after zoom in.
```swift
    open var centerAxisLabelsEnabled: Bool
    {
        get { return _centerAxisLabelsEnabled && entryCount > 0 }
    }
```
2: I found in 
```swift
open func computeAxisValues(min: Double, max: Double)
    ...
    let last = interval == 0.0 ? 0.0 : ChartUtils.nextUp(floor(yMax / interval) * interval
    if interval != 0.0 && last != first // condition is false due to last equals first now
    {
        for _ in stride(from: first, through: last, by: interval)
        {
            n += 1
        }
    }
    ...
```
After zoom in a lot `first` and `last` are the same value, causing `n` to be default value, which is 1 (centerAxisLabelsEnabled is 1 now instead of 0 because above code change in # 1)

danielgindi:
That is fixed. However, you should limit the maximum zoom in to a reasonable number.
If the zoom is in to a position where the difference between two axis labels is "0.000000000000001", it's not a zoom that is useful to the user.

danielgindi:
I guess that by "deep zoom in" you are referring to a zoom so deep the the distance between two labels is zero or close to zero. That's not a usable zoom anyway. Please limit the max zoom in available to the user.

liuxuan30:
yeah 'deep zoom in' means zoom in until you can't see any data entry in a line chart.

Crossarchon:
import UIKit
import Foundation
import Charts

@objc(BarChartFormatter)
public class BarChartFormatter: NSObject, IAxisValueFormatter{
    var months: [String]! = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"]
    
    
    public func stringForValue(_ value: Double, axis: AxisBase?) -> String {
        
        return months[Int(value)]
}
    
    
class BarGraphViewController: UIViewController {

    @IBOutlet weak var barChartView: BarChartView!
   
    var months: [String]!
    
[...]

PR #1565, opened 2016-09-27
Bugfix for fix #1488, #1564

### LINKED ISSUE #1488: (v3.0 + Swift3.0) Time Line Chart Demo crash
## Time Line Chart Demo crash when xAxis direction Zoom in to largest status：
img_3297

## crash info:
```
fatal error: Index out of range
```

liuxuan30:
Attempt to fix #1488, #1564
@danielgindi I need you carefully review this and give details about existing implementation about centerAxisLabels feature. My changes may lack of information and cause more issues, though I have tesed it in time line chart, bar chart, combined chart, multiple bar chart, line chart.

1. I removed the dependency of getting initial `n` value, because `first` and `last` will be equal after zooming a lot
2. I change `centerAxisLabelsEnabled` condition, if entry count > 0, we could turn on centerAxisLabel. I need context why it's `> 1`.
3. I removed using local variable `centeringEnabled`, because `axis.entries` would be changed but `centeringEnabled` keeps reading old entryCount, which seems a dirty read issue
4. I use `offset: Double = interval / 2.0` instead of `offset = (axis.entries[1] - axis.entries[0]) / 2.0`, because `axis.entries[1] - axis.entries[0]` always equals to interval, and when there is only one entry, we still need a valid offset instead of 0

### DIFF
--- a/Source/Charts/Components/AxisBase.swift
+++ b/Source/Charts/Components/AxisBase.swift
@@ -50,7 +50,7 @@ open class AxisBase: ComponentBase
     /// This is useful especially for grouped BarChart.
     open var centerAxisLabelsEnabled: Bool
     {
-        get { return _centerAxisLabelsEnabled && entryCount > 1 }
+        get { return _centerAxisLabelsEnabled && entryCount > 0 }
         set { _centerAxisLabelsEnabled = newValue }
     }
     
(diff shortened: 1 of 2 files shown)


liuxuan30:
@petester42 
seeing
```
*** Building scheme "Realm" in Realm.xcworkspace

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself
```
But no more logs.. Have you met this?

petester42:
Ya realm takes too long to build on ci. Usually it's not a problem because the download limit is not reached. Just restart the build.

danielgindi:
@PhilJay Looks good to me, right?

danielgindi:
Okay I've done some adjustments and merged

Issue #1323, opened 2016-08-16
text value on the last entry is missing from line chart


kennycoding:
repro: run the ios demo, show the first line chart, notice that the value (text) on the last entry is missing, although the value point is drawn.

kennycoding:
my current work around is to add a dummy entry with same value as the last

danielgindi:
On what version of Charts?

kennycoding:
version 2.2.5, build 21

danielgindi:
Did you try on 3.0?

spinosa:
@danielgindi this is still happening on 3.0 as of https://github.com/danielgindi/Charts/commit/55405033684cda362ba2a1b839335dd8b3ba5395 (latest code as of right now; September 28th).

Run the demo.  See the last data point drawn without a label:

<img width="321" alt="screen shot 2016-09-28 at 12 41 07 pm" src="https://cloud.githubusercontent.com/assets/68957/18923064/faecfadc-8578-11e6-8778-e9ea4674c0c9.png">

It seems `drawValues(context:)` may not be calculating `stride` correctly (`LineChartRenderer.swift:552`).

liuxuan30:
I tried to fix it by changing `to` to `through` in 
```swift
for j in stride(from: _xBounds.min, to: Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min))), by: 1)
```
This did fix the missing label, however, I found a new issue that when zoomed in, and the animation on x axis is not in sync: the value  text will jump out first and then come with the line. Changing back to `to` the animation will be good. Weird...

ryanschneider:
Sorry, didn't see this issue before opening #1585 and the #1586 PR.  I'll keep my comments here, but will leave it up to you guys on if you want to close those.

Anyways, what about adjusting the stride based on `phaseX`?

In pseudo-code:

```swift
open override func drawValues(context: CGContext) {
    ///snip
    let isAnimatingX = phaseX != 1.0
    let through = isAnimatingX ?
        Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min))) - 1 :
        Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min)))

    for j in stride(from: _xBounds.min, through: through, by: 1) {

    }
    ///snip
}
```

[...]

PR #1586, opened 2016-09-30
Make sure we render the last y Label on LineCharts (Fixes #1585)

### LINKED ISSUE #1585: Last LineChart y-label not being rendered
In 3.0 (using master from commit 3a2da1b) I'm seeing that the last y-label in my charts is not being rendered.

I've tracked down was I think is the cause to this line:

https://github.com/danielgindi/Charts/blob/a34d55a5c9d16c1f4f6f48f2148b4e88902c55d0/Source/Charts/Renderers/LineChartRenderer.swift#L552

I think the `stride(from:to:by)` should be a `stride(from:through:by)` (use `through`, not `to`), like so:

```swift

    open override func drawValues(context: CGContext)
    {
                //....
                for j in stride(from: _xBounds.min, through: Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min))), by: 1)
                {
                    guard let e = dataSet.entryForIndex(j) else { break }
```

This edit works locally for me, I'll send a PR for the change as well here in a minute, but haven't done an extensive audit for the other chart types to see if a similar issue exists there.

ryanschneider:
The wrong stride termination was causing us to not render the last text value.
Verified the issue in the `ChartsDemo` example, but have not looked closely
at other chart types.

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -549,7 +549,7 @@ open class LineChartRenderer: LineRadarRenderer
                 
                 _xBounds.set(chart: dataProvider, dataSet: dataSet, animator: animator)
                 
-                for j in stride(from: _xBounds.min, to: Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min))), by: 1)
+                for j in stride(from: _xBounds.min, through: Int(ceil(Double(_xBounds.max - _xBounds.min) * phaseX + Double(_xBounds.min))), by: 1)
                 {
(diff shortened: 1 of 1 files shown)


liuxuan30:
This did fix the missing label, however, I found a new issue that when zoomed in, the animation on x axis is not in sync: the value text will jump out first and then come with the line. Changing back to to the animation will be good. awkward...

ryanschneider:
See my comment on https://github.com/danielgindi/Charts/issues/1323#issuecomment-250808675, I agree this PR alone is not enough, but it looks like there's more to the issue that originally thought, as even without this change the current master "pops" in the last data point when animating.

I'm closing this PR and will keep the discussion on #1323.

PR #1621, opened 2016-10-08
fix #1603 and API comment

### LINKED ISSUE #1603: Setting AxisBase.axisMaximum results in an axisRange of zero
When I set axisMaximum for the ordinate on a graph to a value, the range of the values plotted on the ordinate appeared was collapsed to zero. The code for AxisBase.axisMaximum is:

    open var axisMaximum: Double
    {
        get
        {
            return _axisMaximum
        }
        set
        {
            _customAxisMax = true
            _axisMaximum = newValue
            axisRange = abs(_axisMaximum - newValue) // Always results in a range of zero.
        }
    }

This results in axisRange is always being set to zero, because newValue has the same value as _axisMaximum. Making the following change fixed the problem:

    open var axisMaximum: Double
    {
        get
        {
            return _axisMaximum
        }
        set
        {
            _customAxisMax = true
            _axisMaximum = newValue
	   axisRange = abs(_axisMaximum - _axisMinimum)
        }
    }

liuxuan30:
fix #1603 and API comment. Looks like typo, but I changed the order to highlight the difference

### DIFF
--- a/Source/Charts/Components/AxisBase.swift
+++ b/Source/Charts/Components/AxisBase.swift
@@ -328,7 +328,7 @@ open class AxisBase: ComponentBase
     
     /// The maximum value for this axis.
     /// If set, this value will not be calculated automatically depending on the provided data.
-    /// Use `resetCustomAxisMin()` to undo this.
+    /// Use `resetCustomAxisMax()` to undo this.
     open var axisMaximum: Double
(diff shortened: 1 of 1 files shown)


danielgindi:
Dammit, I just fixed that, except the docs

PR #1637, opened 2016-10-10
Fix chart crash when dataSets is an empty array.

raudabaugh:
Swift 2.3 version of #1558

### DIFF
--- a/Charts/Classes/Renderers/AxisRendererBase.swift
+++ b/Charts/Classes/Renderers/AxisRendererBase.swift
@@ -103,18 +103,14 @@ public class AxisRendererBase: Renderer
         let labelCount = axis.labelCount
         let range = abs(yMax - yMin)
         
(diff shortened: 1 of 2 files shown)


petester42:
Thanks. I'll check it out when I have time.

petester42:
Oh wait! This is for charts 3.0? We are not going to support 2.3 for that.

raudabaugh:
Oh, that's unfortunate news. I guess I'll have to migrate my project to Swift 3 sooner. Assuming there are other folks like me who made the move to Charts 3.0 before moving to Swift 3, maybe it would be a good idea to keep a semi-active (albeit unsupported) Chart3.0-Swift2.3 branch on this repo for a little while.

petester42:
Charts 3.0 is not even released yet so it makes no sense to support an older version. You can use charts 2.3 if you need swift 2.3 support.

PR #1638, opened 2016-10-10
Add option for PieChart to disable/enable highlight by tapping inside hole.

sdkdimon:


samova:
Hi,sdkdimon.
Would you like to tell me how to add a tap event of the hole?
I want to show a list detail data by tapping the hole.
Thank you in advance.

PR #1711, opened 2016-10-25
fix #1691 drawTransparentCircleHole bug

### LINKED ISSUE #1691: Setting circleHoleColor to clear causes artifact
If you set the circle hole colour to be `[UIColor clearColor]` then the circle is drawn with an artifact as shown. 
screen shot 2016-10-20 at 20 28 39

liuxuan30:
fix #1691
1. `M_PI_2` is Pi / 2, not Pi \* 2
   
   > public var M_PI_2: Double { get } /\* pi/2           */
2. `clockwise` in `UIKit` and `CoreGraphics` is flipped(compared to UIBezierPath), though 0 to M_PI \* 2 does not get affected by clockwise.
   
   > The clockwise parameter determines the direction in which the arc is created; the actual direction of the final path is dependent on the current transformation matrix of the graphics context. For example, on iOS, a UIView flips the Y-coordinate by scaling the Y values by -1. In a flipped coordinate system, specifying a clockwise arc results in a counterclockwise arc after the transformation is applied.
3. using even odd rule to fill between two arcs, not default.

BTW, I hate the new Apple API doc:( 
The mentioned APIs above do not have any words, while the old doc has rich information. Have to use CGContextAddArc etc. to get the discussion

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -665,10 +665,10 @@ open class LineChartRenderer: LineRadarRenderer
                     context.addEllipse(in: rect)
                     
                     // Cut hole in path
-                    context.addArc(center: pt, radius: circleHoleRadius, startAngle: 0.0, endAngle: CGFloat(M_PI_2), clockwise: true)
+                    context.addArc(center: pt, radius: circleHoleRadius, startAngle: 0.0, endAngle: CGFloat(M_PI * 2), clockwise: false)
                     
                     // Fill in-between
-                    context.fillPath()
+                    context.fillPath(using: .evenOdd)
                 }
                 else
                 {


liuxuan30:
test code (replace `setDataCount` in `LineChart1ViewController`:
```objective-C
{
    NSMutableArray *values = [[NSMutableArray alloc] init];
    
    {
        double val = arc4random_uniform(range) + 3;
        [values addObject:[[ChartDataEntry alloc] initWithX:i y:val]];
    }
    
    LineChartDataSet *set1 = nil;
    if (_chartView.data.dataSetCount > 0)
    {
        set1 = (LineChartDataSet *)_chartView.data.dataSets[0];
        set1.values = values;
        [_chartView.data notifyDataChanged];
        [_chartView notifyDataSetChanged];
    }
    else
    {
        set1 = [[LineChartDataSet alloc] initWithValues:values label:@"DataSet 1"];

        [set1 setHighlightColor:[UIColor clearColor]];
        set1.drawValuesEnabled = NO;
        set1.drawCirclesEnabled = YES;
        set1.circleHoleColor = [UIColor clearColor];
        set1.circleColors = [NSArray arrayWithObject:[UIColor orangeColor]];
        set1.circleHoleRadius = 7.5f;
        set1.circleRadius = 16.0f;
        set1.colors = [NSArray arrayWithObject:[UIColor blueColor]];

        NSMutableArray *dataSets = [[NSMutableArray alloc] init];
        [dataSets addObject:set1];
        
        LineChartData *data = [[LineChartData alloc] initWithDataSets:dataSets];
        
        _chartView.data = data;
    }
}
```

liuxuan30:
@petester42 did you changed any config? Last week it still failed, but I restarted it today and it passes!

petester42:
I havent fixed it yet but I know why it fails. For pull requests the github token doesn't get used in the build. Because of this we hit the limit sometimes.

liuxuan30:
but it somehow passed.. randomly

danielgindi:
This does not fix the hole thing for me...

liuxuan30:
really? What part?

PR #1733, opened 2016-10-27
Added a check against NaN

Selficide:
There needs to be a check against NaN after ChartUtils::roundToNextSignificant() returns.


petester42:
Could you add tests for this change?

liuxuan30:
under what situation it is NaN?

liuxuan30:
root cased in #1737 comments. Due to 32 bit double min is smaller. merging.

PR #1738, opened 2016-10-28
ChartUtils Crash (Fixes #1737)

### LINKED ISSUE #1737: ChartUtils Crash on 32 Bit Devices
On 32 Bit Devices the function `decimals` in `ChartUtils` causes a bad memory access when the default formatter for a Chart is created.

``` swift
    internal class func decimals(_ number: Double) -> Int
    {
        if number.isNaN || number.isInfinite || number == 0.0
        {
            return 0
        }

        let i = roundToNextSignificant(number: Double(number))

        if i.isInfinite
        {
            return 0
        }

        return Int(ceil(-log10(i))) + 2
    }
```

That is because `i` is NaN so the line `return Int(ceil(-log10(i))) + 2` crashes.

ObjColumnist:
Check that i is valid after rounding to prevent a crash.

### DIFF
--- a/Source/Charts/Utils/ChartUtils.swift
+++ b/Source/Charts/Utils/ChartUtils.swift
@@ -51,7 +51,7 @@ open class ChartUtils
         
         let i = roundToNextSignificant(number: Double(number))
         
-        if i.isInfinite
+        if i.isNaN || i.isInfinite
         {
             return 0
         }


liuxuan30:
seems a dup to https://github.com/danielgindi/Charts/pull/1733

PR #1744, opened 2016-10-29
Negative bar now represented by negative range (Fixes #1726)

### LINKED ISSUE #1726: Weird highlight rect in Stacked Bar Chart Negative
Just seeing a weird highlight for bars in Stacked Bar Chart Negative
sbn

SumoSimo:
Try to fix #1726
Positive ranges were previously created for the negative side of the bar. Therefore, tapping on the positive side of a bar would initiate a search for a range that contains the specified point, which in turn returns the first range at index 0 and represents the negative side of the bar. This range is then used to highlight the positive portion of the bar. As an example, if the positive side has a range of 0 to 13, and the negative side has a range of 0 to 12 (because again, the current implementation treats negative ranges as positive), the range of 0 to 12 will be used to highlight the positive side, because that range comes first in the array [Range(0, 12), Range(0, 13)]. This is why we see that weird highlight, which is missing one unit of highlight.

Notice that this issue goes away when you tap the very edge of the bar that has this weird highlight. Using the aforementioned example, if you tap the very edge of the positive bar, such that the y-value is greater than 12, then you get the full highlight because Range(0, 12) does not contain that y-value, but Range(0, 13) does.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
@@ -150,8 +150,8 @@ open class BarChartDataEntry: ChartDataEntry
             
             if value < 0
             {
-                _ranges?.append(Range(from: negRemain, to: negRemain + abs(value)))
-                negRemain += abs(value)
+                _ranges?.append(Range(from: negRemain, to: negRemain + value))
+                negRemain += value
             }
             else
             {


liuxuan30:
I guess highlighting bars when negative values involoved has more bugs than we first thought.
Your code fixes the original issue I reported, however,
in stacked bar chart (ChartsDemo), if you change code like this:
```objective-c
{
    NSMutableArray *yVals = [[NSMutableArray alloc] init];
    
    {
        double mult = (range + 1);
        double val1 = (double) (arc4random_uniform(mult) + mult / 3);
        double val2 = (double) (arc4random_uniform(mult) + mult / 3);
        double val3 = (double) (arc4random_uniform(mult) + mult / 3);
        if (i % 2 == 0)
            [yVals addObject:[[BarChartDataEntry alloc] initWithX:i yValues:@[@(val1), @(val2), @(val3)]]];
        else
            [yVals addObject:[[BarChartDataEntry alloc] initWithX:i yValues:@[@(-val1), @(-val2), @(val3)]]];
    }
```
and then try highlight negative bar segments, it breaks again:
neg

SumoSimo:
I see. I'll take a look later today.

By the way, I changed the code like you had mentioned, but I couldn't get the negative y-values to appear. Was there a setting you modified to get those to show?

liuxuan30:
I just modified Stack bar chart demo and post all changes I made. Not sure what you can't reproduce.

danielgindi:
@liuxuan30 It seems to me that the PR fixes the issue. Do you see any side-effects to this PR? Does this completely fix what you are experiencing?

danielgindi:
Oh right... I tested your sample. It is messed up still

danielgindi:
I think that the PR does not fix, but make the rendering go wrong.
The fix should be in the highlighter's code maybe.

danielgindi:
I did found that the line of code you changed is different from Android implementation. I'm commiting a fix.
But that still does not fix it.

danielgindi:
@PhilJay we do have a problem here. Try to do what @liuxuan30  suggested in MPAndroidChart demo:
Replace in `StackedBarActivity.java`
```
            yVals1.add(new BarEntry(i, new float[]{val1, val2, val3}));
```

by 
```
            if (i % 2 == 0)
                yVals1.add(new BarEntry(i, new float[]{val1, val2, val3}));
            else
                yVals1.add(new BarEntry(i, new float[]{-val1, -val2, val3}));
```

And comment out the `axisMinimum` line.
Now try to highlight the negative bars.

So when there are mixed positives/negatives - we cannot highlight.

PR #1745, opened 2016-10-30
Fix: extra offsets added in PieRadarChartViewBase

Negativchik:
calculateOffsets() function adds extra offsets: 

```
    legendTop += self.extraTopOffset
    legendRight += self.extraRightOffset
    legendBottom += self.extraBottomOffset
    legendLeft += self.extraLeftOffset
    ...
    let offsetLeft = max(minOffset, legendLeft)
    let offsetTop = max(minOffset, legendTop)
    let offsetRight = max(minOffset, legendRight)
    let offsetBottom = max(minOffset, max(self.requiredBaseOffset, legendBottom))
```

Then, diameter reduces by extra offsets. 

```
open var diameter: CGFloat
    {
        var content = _viewPortHandler.contentRect
        content.origin.x += extraLeftOffset
        content.origin.y += extraTopOffset
        content.size.width -= extraLeftOffset + extraRightOffset
        content.size.height -= extraTopOffset + extraBottomOffset
        return min(content.width, content.height)
    }
```

So pie chart becomes smaller then it was expected.

### DIFF
--- a/Source/Charts/Charts/PieRadarChartViewBase.swift
+++ b/Source/Charts/Charts/PieRadarChartViewBase.swift
@@ -217,12 +217,12 @@ open class PieRadarChartViewBase: ChartViewBase
             legendRight += self.requiredBaseOffset
             legendTop += self.requiredBaseOffset
             legendBottom += self.requiredBaseOffset
+            
+            legendTop += self.extraTopOffset
+            legendRight += self.extraRightOffset
+            legendBottom += self.extraBottomOffset
+            legendLeft += self.extraLeftOffset
         }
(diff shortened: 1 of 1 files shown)


danielgindi:
To me it seems like the diameter being affected by the extra offsets is pretty much expected...
As it is meant to move or shrink that charts, and leave more space for the legend.

So what is the point in this?

danielgindi:
The base offsets are dependent on the legend being present. "extra" offsets are supplied by the developer as he requires more spacing, without any relation to the legend.
I'm closing this.

PR #1771, opened 2016-11-03
Updated Width Constraints - Fixes #1770

### LINKED ISSUE #1770: [BUG] multiple bar chart in ChartsDemo not display full for last x value
multiple bar chart in ChartsDemo not display full for last x value.
When dragging the x slider, it grows and shrinks, not always full.
@danielgindi is this expected behavior? I would assume it should display fully.

mb

SumoSimo:
Removed width constraint on slider-text-x and slider-text-y, so width  can adjust automatically based on set text. Also added equal-width
constraint on slider-text-x in relation to slider-text-y, so the width
of both will always be the same.


danielgindi:
Thank you!

PR #1785, opened 2016-11-04
Fix: issue with image representation cases for iOS in v2.3.1

laxmankhanal:


laxmankhanal:
Cannot build successfully v2.3.1 similar to v2.3.2 using carthage.

petester42:
The code you are changing only affects macOS anyways. The code is also wrong. `NSPNGFileType` is not an enum case so it can't be used as `. NSPNGFileType`. Please update your cartfile and run `cathage update`. I have fixed the bug you mentioned in your previous pull request so I will close this. If you have troubles installing and such please open an issues.

PR #1867, opened 2016-11-21
Percentage problem

ybeapps:
When setting NumberFormatter numberStyle to .percent, it expects to get a number from 0 to 1.
But here we give from 0 to 100.
So fix it =]

### DIFF
--- a/Source/Charts/Renderers/PieChartRenderer.swift
+++ b/Source/Charts/Renderers/PieChartRenderer.swift
@@ -346,7 +346,7 @@ open class PieChartRenderer: DataRenderer
                 
                 let transformedAngle = rotationAngle + angle * CGFloat(phaseY)
                 
-                let value = usePercentValuesEnabled ? e.y / yValueSum * 100.0 : e.y
+                let value = usePercentValuesEnabled ? e.y / yValueSum : e.y
                 let valueText = formatter.stringForValue(
                     value,
                     entry: e,


liuxuan30:
Why this? As you see:
```swift
let value = usePercentValuesEnabled ? e.y / yValueSum * 100.0 : e.y
```
It only `* 100` when `usePercentValuesEnabled = true`
If you turn on usePercentValuesEnabled, why you need a percent style formatter? Am I missing anything?

ybeapps:
> usePercentValuesEnabled gives the numberFormatter a number from 0 to 100 - instead of the value (which is not limited) it's the calculation that changes.
> 
I use the formatter for another thing, but the formatter by defaults return a number without the % char. So I use the formatter with .percent numberStyle, and I get at last the correct number with %.

The Charts framework as you can see (by multiplying in 100) gives the formatter a value between 0 to 100.
But the formatter expects to get a number from 0 to 1.
So if Charts gives 94.6667, in the screen it will look 9,466% which is a very bad result...
By eliminating the 100 multiplication that fixes the error (the other option is to subclass the number formatter and divide by 100 the input. Not a nice way)

Good day


‫ב-22 בנוב׳ 2016, בשעה 4:00, ‏‏Xuan ‏<notifications@github.com> כתב/ה:‬

liuxuan30:
@ybeapps yeah I see it. My question would be, why don't you just set `usePercentValuesEnabled` and not set the formatter to be percent style again? It looks like you enabled two percent formatting, however I think you just need set only one of them.
I am guessing `usePercentValuesEnabled` is just a convenient switch to format the pie chart, then you don't need a percent style formatter.

ybeapps:
@liuxuan30 I use the `usePercentValuesEnabled` to _calculate_ the **values** of the data as percents... and I use the formatter to _format_ the number as percent (by **adding '%' char**)... That's why I think both are needed.
1) calculation
2) formatting

If you think otherwise I can do many workarounds so it work as I expect for me.. but I just wanted to contribute because I thought it's supposed to be like this...

good day

liuxuan30:
Yes I understand, but I am asking because Pie chart demo seems fine so far.

Can you show how you set up your formatter?

Here's what it looks like in ChartsDemo pie chart:
```objective-c
    NSNumberFormatter *pFormatter = [[NSNumberFormatter alloc] init];
    pFormatter.numberStyle = NSNumberFormatterPercentStyle;
    pFormatter.maximumFractionDigits = 1;
    pFormatter.multiplier = @1.f;
    pFormatter.percentSymbol = @" %";
    [data setValueFormatter:[[ChartDefaultValueFormatter alloc] initWithFormatter:pFormatter]];
```
I am thinking `pFormatter.multiplier = @1.f;` is all you needed for correctly showing percent numbers. I tested its default is 100.0, setting to 1 will eliminate the side effect while formatting again.

On the other side, this is not an obvious feature, I am not sure why it's implemented this way, so we can have @danielgindi provide some details.

ybeapps:
yeah I agree that we should ask @danielgindi 
I'm not setting multiplier and percentSymbol.. I use the default values...

jjatie:
This is an issue of ambiguity. The API should be more obvious on how to do percent formatting, however we have a demo showing how we expect it to be done. My concern with merging this is that it would break existing formatting. When dg get's a chance, hopefully he can make a decision on it : )

PR #1948, opened 2016-12-12
ChartDataSet index o.o.b. check

### LINKED ISSUE #1945: notifyDataChanged() will crash for dataset without entries
More specifically
Setting CandleChartData with a dataset containing 0 entries and then calling data?.notifyDataChanged() will cause a crash.

I think this happens because it will try to reference entries using subscript index, so probably tries to get entries[0] (because count = 0) and that obviously won't work because that entries doesn't exist.

A workaround is to check if your dataset contains any entries and set it data to nil otherwise.

However, I think Charts could improve here by either using safe subscript referencing (so you can check entries[0] safely without crashing) or just checking if there are actual entries before trying to subscript access them.

whvandervelde:
Return nil for index out of bounds to prevent crash on empty datasets (Fixes #1945)

Signed-off-by: Wytse van der Velde <w.h.vandervelde@gmail.com>

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -184,6 +184,10 @@ open class ChartDataSet: ChartBaseDataSet
     /// if `i` is out of bounds, it may throw an out-of-bounds exception
     open override func entryForIndex(_ i: Int) -> ChartDataEntry?
     {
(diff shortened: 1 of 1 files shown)


liuxuan30:
just saw https://github.com/danielgindi/Charts/pull/2167 get merged, and there are two similar PRs. I am closing this.
Thank you for your contribution!

PR #2026, opened 2016-12-31
Center align multiple lines of no data text

edwardmarks:
I believe this is the expected behavior of no data text.

The current behavior, if you write multiple lines (i.e. `chartView.noDataText = "No Data Yet\n(check back in a bit)"` is for them to left-align to each other. Given that a single line is centered within the chart, the expected behavior is that multiple lines would be centered within the chart.


jjatie:
@liuxuan30 can you please verify if this is desired?
@edwardmarks Can you please pull master.

jjatie:
@liuxuan30 This seems like something we want, what do you think? I'll make a new PR for 3.1 if we do.

jjatie:
#3199 provides custom alignment

Issue #2039, opened 2017-01-06
xAxis.valueFormatter = self  create retain cycle


hotuanlinh89:
When you create CombinedChart. Config ChartXAxis.
ChartXAxis *xAxis = _chartView.xAxis;
    xAxis.labelPosition = XAxisLabelPositionBottom;
    xAxis.axisMinimum = 0.0;
    xAxis.granularity = 1.0;
    xAxis.valueFormatter = self;

It's create memory leak. Please help.

liuxuan30:
refer this: https://github.com/danielgindi/Charts/issues/1854
You can either change to weak or not use self

Issue #1873, opened 2016-11-22
Should valueFormatter property use weak modifier?


Huang-Libo:
Tihs property is found in `Charts-Swift.h` file, it's a delegate, should it use **weak** modifier to avoid retain cycle?
```
@property (nonatomic, strong) id <IChartAxisValueFormatter> _Nullable valueFormatter;
```

petester42:
Would probably make sense.

Huang-Libo:
@petester42  
I found a same issue:  https://github.com/danielgindi/Charts/issues/1854

There are three ways to solve this problem: 
1. Use weak to modify valueFormatter (I think it's better because valueFormatter is a delegate)
2. Use weakSelf to avoid retain cycle
3. Manually set valueFormatter to nil when leave the controller.

liuxuan30:
Where do you see it? In which class?

kientux:
@liuxuan30 It's in `ChartAxisBase` class. I had this problem too and if `valueFormatter` doesn't need to be `strong` property, it should be `weak` to prevent retain circle when assign it to `self.` It took me 2 hours to find out why my view controller didn't deallocated :(

liuxuan30:
I filed a PR for related issues.

liuxuan30:
I tested it today and found serious issues. If we apply weak, ChartsDemo won't display axis labels at all. Still investigating.

liuxuan30:
Some findings:
1. if I change it to weak, `_axisValueFormatter` can't be correctly initialized. The getter of `valueFormatter` will even return nil, though we force to create a default one, however it does not set `_axisValueFormatter`. I post the details in http://stackoverflow.com/questions/41582846/weak-property-cannot-be-initialized-in-swift-3-0 to track. I don't think I have enough knowledge for this, so help wanted.

2. changing to weak in swift code, will not impact the objective-C header file, it's still strong:
>@property (nonatomic, strong) id _Nullable valueFormatter;

You might want to check on your side to see if it's true. I use latest Xcode 8.2.1.

Sitting down and think of this, I don't think it's a good practice to use weak regardless the problems I had above, but if we really do so, people need to retain the formatter somewhere. So using `weakSelf` might be a better solution here.

Huang-Libo:
@liuxuan30 
Interesting...

931743010:
use weakSelf   ；Still unable to release
 __weak typeof(self) weakSelf=self;
    ChartXAxis *xl = _chartView.xAxis;
[...]

Issue #1854, opened 2016-11-18
Retain cycle setting delegate to valueFormatter


edalford11:
Just migrated to 3.0. Good to go except I can't get rid of a retain cycle that seems to be coming from IChartAxisValueFormatter

Chart set up below

```Objective-C
CGRect *frame = CGRectMake(0, 0, self.chartContainer.frame.size.width, self.chartContainer.frame.size.height);
LineChartView *chart = [[LineChartView alloc] initWithFrame:frame];
chart.legend.enabled = NO;
chart.delegate = delegate;
chart.descriptionText = @"";
chart.backgroundColor = [UIColor whiteColor];
chart.rightAxis.enabled = NO;
chart.dragEnabled = NO;
[chart setScaleEnabled:NO];
chart.pinchZoomEnabled = NO;
chart.drawGridBackgroundEnabled = NO;
chart.userInteractionEnabled = NO;
    
ChartYAxis *leftAxis = chart.leftAxis;
leftAxis.labelPosition = YAxisLabelPositionInsideChart;
leftAxis.labelFont = axisFont;
leftAxis.labelTextColor = [UIColor lightGrayColor];
leftAxis.gridColor = kDisabledGray;
    
ChartXAxis *xAxis = chart.xAxis;
xAxis.labelPosition = XAxisLabelPositionBottomInside;
xAxis.labelFont = axisFont;
xAxis.labelTextColor = [UIColor lightGrayColor];
xAxis.gridColor = kDisabledGray;        

self.mainChart.userInteractionEnabled = YES;
[self.mainChart setAccessibilityLabel:@"Landscape Price Chart"];
[self.mainChart setViewPortOffsetsWithLeft:0.0 top:5.0 right:0.0 bottom:0.0];
        
ChartYAxis *leftAxis = self.mainChart.leftAxis;
NSNumberFormatter *numberFormatter = [[NSNumberFormatter alloc] init];
[numberFormatter setNumberStyle:NSNumberFormatterDecimalStyle];
[numberFormatter setUsesGroupingSeparator:NO];
[numberFormatter setMaximumFractionDigits:2];
[numberFormatter setMinimumFractionDigits:2];
ChartDefaultAxisValueFormatter *valueFormatter = [[ChartDefaultAxisValueFormatter alloc] initWithFormatter:numberFormatter];
leftAxis.valueFormatter = valueFormatter;
        
ChartXAxis *xAxis = self.mainChart.xAxis;
[...]

PR #2054, opened 2017-01-10
Use weak for axisValueFormatter, addresses #1873, #1854, #2039

liuxuan30:
Use `weak` for axisValueFormatter, addresses #1873, #1854, #2039


liuxuan30:
I tested it today and found serious issues. If we apply weak, ChartsDemo won't display axis labels at all. Still investigating.

liuxuan30:
Closing as I don't think using weak here is a good practice here, not even mentioning the problems it introduces. After we have a clean vision, we will redo it.
Discussions in #1873

PR #2056, opened 2017-01-10
Fix NSCopying implementation in CandleChartDataEntry

leo150:
Fix few typos in `copyWithZone` method.


liuxuan30:
looks good to me.

PR #2064, opened 2017-01-11
Update Podspec RealmSwift Dependency to 2.1.1 to be in line with Cartfile

anttyc:
Bumped the dependency version from 1.1 to 2.1.1 because it was preventing me from using the latest RealmSwift(2.1.2) with Charts, but thought it would be a good idea to keep it in line with what's in the Cartfile


petester42:
Thanks

PR #2096, opened 2017-01-19
Swift 3: Don't confuse ChartRange with Range

x128:
In Swift 3, [`Range`](https://developer.apple.com/reference/swift/range) is a generic structure. Probably it's connected to [this issue](https://github.com/apple/swift-evolution/blob/master/proposals/0086-drop-foundation-ns.md) (NSRange has been a part of Foundation, it's still present there, but Apple has already started to migrate from NSRange to Range).

When I tried to used it (for example, `Range<Int>` in a function argument type like [here](http://stackoverflow.com/a/26775912/1289683)), the compiler shows two errors:
**Cannot specialize non-generic type 'Range'
'Range' is ambiguous for type lookup in this context**

Unfortunately, I can't paste my project here. The two "Ranges" used to successfully co-exist in it, but it was broken after a major refactoring (I didn't touch any code with Charts or [NS]Ranges, I promise! maybe it was something with bridging headers)...

Anyway, we should not confuse Charts Range and the new Range structure.

### DIFF
--- a/Charts.xcodeproj/project.pbxproj
+++ b/Charts.xcodeproj/project.pbxproj
@@ -90,7 +90,7 @@
 		06B120901D811B8900D14B02 /* PieHighlighter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06B120071D811B8900D14B02 /* PieHighlighter.swift */; };
 		06B120911D811B8900D14B02 /* PieRadarHighlighter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06B120081D811B8900D14B02 /* PieRadarHighlighter.swift */; };
 		06B120921D811B8900D14B02 /* RadarHighlighter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06B120091D811B8900D14B02 /* RadarHighlighter.swift */; };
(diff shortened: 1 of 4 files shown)


petester42:
You can use `Charts.Range` to fix that

x128:
The compile problem was with Swift's built-in `Range`.

petester42:
Oh ok but in our code or yours? There is also `Swift.Range` to fix things like this.

petester42:
Also I'm reluctant to change the `Range` name since it is matches the swift conventions where `ChartsRange` matches the objc conventions. Since swift has modules and name spacing we don't need to include the prefix `Charts` in the type declaration.

x128:
Swift has some kind of namespacing but lacks any common approach to handle collisions. Who on Earth could invent such a weird solution?

The thing is: `Swift.Range` works perfectly in my modules, great, thanks for the hint, maybe it will help someone. But this solution incompatible with [AKMaskField](https://github.com/artemkrachulov/AKMaskField) which is also used in my project. The compiler gives the same errors as above. Do you think the authors of AKMaskField need to specify `Swift.Range`, too?

petester42:
Without having source code that reproduces this problem it's hard for me to say. Moving clashing modules to other files sometimes helps. If you import 2 modules with the same type in one file it may clash. Try moving one of the two to another file and put your code there. It should fix your problem.

PR #2101, opened 2017-01-20
Adds NSPhotoLibraryUsageDescription to plist of ChartsDemo

valeriyvan:
Adds NSPhotoLibraryUsageDescription to plist of ChartsDemo to prevent crash on iOS 10 on try to store shot of char to photo library.


petester42:
Thanks!

Issue #1880, opened 2016-11-24
Chart x offset changes when finished zooming in


BenjaJimenez:
I'm seeing a weird behavior when I zoom in my charts.

After I end zooming (which means releasing both fingers) the xAxis offset changes as it moves like 20 chart points to the right of where the zoom ended.

The chart is inside another scrollview, but has only vertical scroll. Also this only happens when the zoom level is really close.. not sure exactly when. 
My code is pretty much the same as the demo.. only a few changes here and there.

Here's a video with the behaviour I'm seeing https://youtu.be/6ZIJmTyuie4
Any idea on why this is happening or how can I fix it?

liuxuan30:
Thanks for the video. But I don't see whats the problem.. Maybe a time range so I can pay attension.
BTW, `drawLabels()` in x axis renderer is where it display the labels, like calcuating if it's out of bound, etc.

BenjaJimenez:
Thanks I'll try to clarify the video

[0:01-0:14] Pinch gesture to zoom into the graph
[0:15] Pinch gesture ends (released both fingers, no other gesture performed) Instead of staying at the zoomed spot, the graph automatically "scrolls" a few points to the right. 
[0:16-0:17] Me scrolling back left to where the graph should be when the gesture ended..
[0:17+] repeating the same process

 This only happens at high level zooms, my guess is that when the pinch gesture ends, the graph is drawing itself again but at a wrong offset or chart place.

liuxuan30:
I kind of getting it, but I can't answer.. need @danielgindi 's help

ClausEspersen:
I have the same issue most significant at high zoom levels, one out of like 4 zooms results in a wrong x-axis position.
It can jump a tiny bit at low zoom too. 
It also happens when the chart is set up to only zoom in one direction. In my case it jumps several screens widths to the right which is around 2% of the total chart view width.
viewportHandler scaleX is around 13500
Maybe it could be a rounding bug.

hanneshallen:
Same problem here. Seems to be related to the width of the y axis. The error appears clearly when the y axis is scaled and there is another decimal added or removed from the labels on the y axis. Setting a fixed min and max width of the y axis solved it for me.

mikeCalvir:
I am seeing this too, and actually in both directions. When zoom in and get close/at to maximum zoom, it jumps to the end of the data to the right. When zooming out, it jumps back a week or so. It doesn't always happen, but is frequent.
[...]

PR #2114, opened 2017-01-24
#1880 Fix pinch zoom handling at scale limits

markdanyo:
# **Summary**
Fix for issue [#1880](https://github.com/danielgindi/Charts/issues/1880) - Chart x offset changes when finished zooming in

Revised handling of pinch gesture recognizer in `BarLineChartViewBase` to avoid misaligned scale/translation values in transform applied to `viewPortHandler.touchMatrix`.

# **Cause of Issue**
The observed behavior when pinching to zoom a bar/line chart was that, when the min/max scale were reached on a particular axis, the offset on that axis would instantaneously shift by some amount. This behavior was caused by `ViewPortHandler.limitTransAndScale()`, which correctly clipped the scale to the appropriate min/max values, but did not account for the difference in translation that that clipping should have introduced.

For example, if a user attempts to zoom the chart out past its minimum X-axis scale, the `tx` value of `touchMatrix` will initially result in a `minimumVisibleX` value that is too low (causing the unwanted change in X offset).

# **Description of Fix**
Prior to concatenating `touchMatrix` with the incremental transform calculated in the chart's pinch gesture handler, potentially resulting in a transform that must be clipped to the chart's scale limits, we now concatenate an already-clipped incremental transform. This calculation is performed in `ViewPortHandler.clippedZoom()` and generates a transform that, when concatenated with `touchMatrix`, will result in a transform that does not need to be clipped to the scale limits. At the min/max scales, both the scale and translation of `touchMatrix` will reflect the expected value.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -629,9 +629,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                     let scaleX = canZoomMoreX ? recognizer.nsuiScale : 1.0
                     let scaleY = canZoomMoreY ? recognizer.nsuiScale : 1.0
                     
-                    var matrix = CGAffineTransform(translationX: location.x, y: location.y)
-                    matrix = matrix.scaledBy(x: scaleX, y: scaleY)
-                    matrix = matrix.translatedBy(x: -location.x, y: -location.y)
+                    var matrix = _viewPortHandler.clippedZoom(scaleX: scaleX, scaleY: scaleY, x: location.x, y: location.y)
                     
                     matrix = _viewPortHandler.touchMatrix.concatenating(matrix)
                     
--- a/Source/Charts/Utils/ViewPortHandler.swift
+++ b/Source/Charts/Utils/ViewPortHandler.swift
@@ -226,6 +226,41 @@ open class ViewPortHandler: NSObject
         return matrix
     }
     
+    /// Calculates an incremental transform to the viewport based on a scale amount and anchor location
+    /// Min/max scale are taken into account to clip the resulting scale and translation values accordingly
+    /// `limitTransAndScale` will eventually clip these values independent of one another,
+    /// but we need to make sure they are aligned to avoid unwanted translation at min/max zoom levels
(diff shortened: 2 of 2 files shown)


jjatie:
@liuxuan30 can this be considered for 3.1?

liuxuan30:
@jjatie if we target at Jan 31st, I guess not. But it's good to take a look and get merged, if legit.

jjatie:
@markdintrepid does this only affect bar and line charts? How did we avoid the issue in the other chart types?

mikeCalvir:
Is this going to make it into 3.1 do we know? And is v3.1 due at the end of Feb?

liuxuan30:
@mikeCalvir I hope so, but if we are adding more PRs on our plate, it will be delayed.

liuxuan30:
please pull master to include pie chart tests

mikeCalvir:
Hi guys, any news on this one for the 3.1 release? 
Sorry to hassle, but it's the only bug I have now in this app I have made, unfortunately it is an annoying bug - it kind of ruins the data that whenever we zoom in to look at it, we get jumped to the end.
I'll try and see if I can help some way this afternoon...

jjatie:
@mikeCalvir 3.1 should be released very soon. Two minor fixes in existing PRs and we're good to go. Unfortunately it looks like @markdintrepid has abandoned the PR, so this will not be included. If you would like to resubmit the changes and fix the existing PR, you're welcome to do so and we can release this fix.

liuxuan30:
I don't think this will be addressed in 3.1, as it's a big change behind the scene. We need to figure out the math and the issue together, which takes lots of time and we don't have much right now. I mared it 2 stars to make it serious for us to consider first.

philfi:
Hi is there anything I can do to help get this PR back on track? Doesn't look like there has been any activity in a long while.

philfi:
@liuxuan30 Aren't we all! Still interested in contributing, I think there's lots of potential in this project and would love to assist in whatever way I can.

liuxuan30:
@philfi I agree, but lack of time 😂  you could help review, send pull requests to fix bugs or add enhancements.

ericschaal:
Hey! any update on this? Still happening in v3.4.0

liuxuan30:
the OP didn't merge master

jjatie:
OP has abandoned PR. Feel free to submit again.

PR #2116, opened 2017-01-25
Fixes Xcode 8.3 beta build issue

davidarve:
This fixes the following build error in Xcode 8.3 beta
>/.../Charts/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift:61:40: Binary operator '!==' cannot be applied to operands of type 'AnyObject?!' and 'AnyObject?'

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
@@ -58,7 +58,7 @@ open class ChartDataEntryBase: NSObject
             return false
         }
         
-        if (object! as AnyObject).data !== data && !((object! as AnyObject).data??.isEqual(self.data))!
+        if (object! as AnyObject).data as AnyObject? !== data && !((object! as AnyObject).data??.isEqual(self.data))!
         {
             return false
         }


petester42:
I created a new Xcode 6.3 branch. If you could target your changes to that branch it would be great!

davidarve:
The failed build doesn't seem to have anything to do with the patch, could someone take a look at it?

petester42:
It actually is related. The fix you are proposing doesn't work in Xcode 8.2 that is why it is failing.

davidarve:
Hmm. That sucks. Does anyone have another solution, otherwise I may have time to look at it closer later today.

davidarve:
I've updated the pull request. I don't understand and can't reproduce the current build failure with Xcode 8.2.

petester42:
The current error is fine. It happens from time to time.

davidarve:
Could someone trigger the build again to see if it can pass? @petester42? It would of course also be nice if we could run a Travis build using Xcode 8.3 beta to verify that the fix works.

petester42:
Xcode 8.3 beta is not available yet. We don't need to restart the build. It's fine 2 out of 3 platforms passed. Should be good.

PR #2162, opened 2017-02-14
remove deprecation warnings in XCode 8.3 

LeviDahl:
Remove deprecations related to M_PI, M_PI_2, DBL_MAX, and DBL_EPSILON, which are deprecated in Swift 3.0.


liuxuan30:
.greatestFiniteMagnitude this name is killing me.

danielgindi:
```
"This value corresponds to type-specific C macros such as FLT_MAX and DBL_MAX. The naming of those macros is slightly misleading, because infinity is greater than this value."
```

So the old naming is slightly misleading....

PR #2167, opened 2017-02-15
Fixes index out of range crash.

kzaher:
This crashed in prod for us.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -184,6 +184,9 @@ open class ChartDataSet: ChartBaseDataSet
     /// if `i` is out of bounds, it may throw an out-of-bounds exception
     open override func entryForIndex(_ i: Int) -> ChartDataEntry?
     {
+        guard i >= 0 && i < _values.count else {
+            return nil
+        }
         return _values[i]
     }
     


petester42:
Thanks. The failing ci is because realm is taking too long to build. I'll merge this.

danielgindi:
Actually, the docs of this function says very clearly that it throws an exception if out of bounds. Why did you run into this in the first place? The caller code should check the bounds

kzaher:
@danielgindi Swift throwing mechanism is different then Java's.

If this was intended behavior, then the prototype should have been:

```
open override func entryForIndex(_ i: Int) throws -> ChartDataEntry
```

or if you wanted fatal error:

```
open override func entryForIndex(_ i: Int) -> ChartDataEntry
```

screen shot 2017-02-22 at 21 06 00

danielgindi:
Yes, I agree that it should be declared as `throws`.

But I'm trying here to understand the situation, not the syntax...
How did you end up getting a fatal error? Did you manually call the function with an invalid argument, updated the chart with invalid data, or did the chart fail and did something wrong?

nikwest:
I ran into the same problem. I created an empty LineChartDataSet. This will crash once the view is displayed.
It might be the problem that 
stride(from:through:by:)
should be replaced by 
stride(from:to:by:)
in LineChartRenderer -> drawLinear(..) Line 332
(just a guess)

Issue #2053, opened 2017-01-10
Handling critical Y values with autoScaleMinMaxEnabled


Fahrenheit-sp:
Hi. I have a LineChartView with **autoScaleMinMaxEnabled** option enabled. I retrieve some data from internet and dynamically add entries to my chartView, based on the data retrieved. 
Here is some sample code:

```
let receivedEntry = ChartDataEntry(x: json["time"].doubleValue, y:json["value"].doubleValue)
set.addEntry(receivedEntry)
chartView.notifyDataSetChanged()
```

everything works just well, but if _receivedEntry.y_ is higher than my current yAxisMaximum, or lower than minimum, the new value gets drawn beyond the bounds (it means that some part of chart goes beyond the top of chartView, or beyond the bottom, depending on receivedEntry.y value), and gets autoscaled only if I translate or scale it manually.

I tried to set axisMaximum/axisMinimum manually, and then reset it by calling _resetAxisMinMax_, when I start scaling/translating chartView
```
func handleCriticalValues() {
        if receivedEntry.y >= chartView.rightAxis.axisMaximum - (chartView.rightAxis.axisRange * Double(chartView.rightAxis.spaceTop)) {
            chartView.rightAxis.axisMaximum = receivedEntry.y + (chartView.rightAxis.axisRange * Double(chartView.rightAxis.spaceTop))
            chartView.notifyDataSetChanged()
        }
        
        if receivedEntry.y <= chartView.rightAxis.axisMinimum + (chartView.rightAxis.axisRange * Double(chartView.rightAxis.spaceTop)) {
            chartView.rightAxis.axisMinimum = receivedEntry.y - (chartView.rightAxis.axisRange * Double(chartView.rightAxis.spaceTop))
            chartView.notifyDataSetChanged()
        }
    }
```
(this gets called right before _chartView.notifyDataSetChanged()_ from previous code sample). This solution works almost perfect, but when I translate chart to the current position, and critical value arrives, chart goes beyond the bounds anyway.

If any additional information or logs needed, please let me know. Thanks for any help and thanks again for such awesome library.

liuxuan30:
Your code sample seems fine, the chart will recalculate and trigger a new draw(), but I am not sure what you mean 
>"but when I translate chart to the current position, and critical value arrives, chart goes beyond the bounds anyway."

You have to debug a little to see when you translate and add new entries, what's the order of the processing.

Fahrenheit-sp:
[...]

PR #2177, opened 2017-02-20
Fixed the inconsistency of AxisMax and AxisMin

aelam:
When enabled autoScale.
`data.calcMinMaxY(fromX: self.lowestVisibleX, toX: self.highestVisibleX)` is called after 
` _leftYAxisRenderer?.computeAxis(min: _leftAxis._axisMinimum, max: _leftAxis._axisMaximum, inverted: _leftAxis.isInverted)`
`_rightYAxisRenderer?.computeAxis(min: _rightAxis._axisMinimum, max: _rightAxis._axisMaximum, inverted: _rightAxis.isInverted)`
`_xAxisRenderer?.computeAxis(min: _xAxis._axisMinimum, max: _xAxis._axisMaximum, inverted: false)`

that means _leftAxisRenderer and _rightAxisRenderer will render AxisLines and Values based on old values of _axisMinimum, _axisMaximum

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -174,6 +174,12 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         // execute all drawing commands
         drawGridBackground(context: context)
         
+
+        if _autoScaleMinMaxEnabled
+        {
+            autoScale()
+        }
+
         if _leftAxis.isEnabled
         {
             _leftYAxisRenderer?.computeAxis(min: _leftAxis._axisMinimum, max: _leftAxis._axisMaximum, inverted: _leftAxis.isInverted)
@@ -191,11 +197,6 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         _leftYAxisRenderer?.renderAxisLine(context: context)
         _rightYAxisRenderer?.renderAxisLine(context: context)
 
-        if _autoScaleMinMaxEnabled
-        {
-            autoScale()
-        }
-        
         // The renderers are responsible for clipping, to account for line-width center etc.
         _xAxisRenderer?.renderGridLines(context: context)
         _leftYAxisRenderer?.renderGridLines(context: context)


liuxuan30:
We need to understand the context. Any issues to track?
I saw a few reports regards autoScaleMinMaxEnabled, I'm not sure if they all are related.

aelam:
Yeah, exactly.
the main point is when is the `data.calcMinMaxY(fromX: self.lowestVisibleX, toX: self.highestVisibleX) called
`

aelam:
Do you have a minute,  we talk on gitter

aelam:
I've found maybe it's the reason of  #2053

danielgindi:
This IS an issue.
@liuxuan30 can you find affected issues and close them? We have too many I can't get a handle on them...

aelam:
@dannielgindi , since the android version has the same problem and this PR is accepted. Do you need to make sure android version use the same solution?

danielgindi:
Nope, I trust that the behavior is the same on both :-)

aelam:
I mean Android version has the same bug. Do you need to tell PhilJay to sync the code

danielgindi:
I've already committed the change to the android version. Please look 4 lines above...

aelam:
Sorry, I was on mobile phone. Anyway that's great !

aelam:
I found your PR to android version is a bit different from this 
the `autoScale()` is after 
`mAxisRendererLeft.computeAxis(mAxisLeft.mAxisMinimum, mAxisLeft.mAxisMaximum, mAxisLeft.isInverted());
`
`mAxisLeft.mAxisMinimum`  and `mAxisLeft.mAxisMaximum` could be changed in `autoScale()`?
If I'm right, it will still have some render problems

danielgindi:
I think that the `autoScale` should be after computer axis, and before the renderers.
Why did you move it before `computeAxis`?

aelam:
Let's go through this again 
if `autoScale` is after compute axis
1. `_leftYAxisRenderer?.computeAxis(min: _leftAxis._axisMinimum, max: _leftAxis._axisMaximum, inverted: _leftAxis.isInverted)`
the entries of YAxis are determined and drawn
We can know there is a range for leftYAxis <-- these values are old

2 .`autoScale() ` is called 
look into the steps of `autoScale()`
In `data.calcMinMaxY(fromX: self.lowestVisibleX, toX: self.highestVisibleX)`,  `_leftAxisMax/Min` the new range of LeftYAxis  are recalculated.   <-- these values are new
3. `renderer?.drawData(context: context)` bases on the new Range **while AxisRenderer render AxisData based on the old range** 

That still leads the inconsistency

Correct me if I am wrong

danielgindi:
Okay, looking more closely at the situation - you're right. I'll apply the right fix

liuxuan30:
I will try to find related issues to close. Just have baby to take care, so I may have a long delay time

danielgindi:
@liuxuan30 Congratulations :-) That's a good reason for not having free time, and that's my main reason!

liuxuan30:
Just linked the open issues that mentioned autoScaleMinMaxEnabled, however some of them seem different issue.
I used to imagine I could finally do some work in the midnight, but now I just can't help going to sleep right after babysitting, lol

PR #2229, opened 2017-03-07
Fixed, If the last value is the max or min, the range will be wrong

aelam:
The last entry is ignored.
If the last value is the max or min, the range will be wrong

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -109,8 +109,7 @@ open class ChartDataSet: ChartBaseDataSet
         }
     }
     
-    open override 
-    (fromX: Double, toX: Double)
+    open override calcMinMaxY(fromX: Double, toX: Double)
     {
         if _values.count == 0
         {
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -124,7 +124,7 @@ open class ChartDataSet: ChartBaseDataSet
         
         if indexTo <= indexFrom { return }
         
-        for i in indexFrom..<indexTo
+        for i in indexFrom...indexTo
         {
             // only recalculate y
             calcMinMaxY(entry: _values[i])
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -109,7 +109,8 @@ open class ChartDataSet: ChartBaseDataSet
         }
     }
     
-    open override func calcMinMaxY(fromX: Double, toX: Double)
+    open override 
+    (fromX: Double, toX: Double)
     {
         if _values.count == 0
         {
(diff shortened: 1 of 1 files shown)


aelam:
Another problem.
How about just one entry? 
The max/min are the same, if the autoScale is enabled, the charts doesn't know the range because it is 0, right?

liuxuan30:
Is there any issue to track this PR? we have no context for this change.

aelam:
I didn't notice other issues about this. 
Let me tell you my case 
1. XAxis can scale , YAxis can autoScale
2. ChartView just has 2 CandleStick entries(other type Data is the same, I think), and the second one 
is the larger one 

Go through this case 
My visible XValues is 1 ,2, my XAxis is from 0.5 to 2.5
When goes into  `calcMinMaxY`
indexFrom = 0, indexTo = 1
and the` indexFrom..<indexTo` is an open range 
that means it just calculate one value 
if it were `indexFrom...indexTo`, it would work fine

Case 2: Just one entry
There is only one entry. then `indexTo = 0`, `indexFrom = 0`
when goes into ` if indexTo <= indexFrom { return }`  DONE 
it will never run into the for-loop to calculate. 
Then I can't get the correct Y range anyway 

Case 3: Based on the case 2, Just one entry 
and the YMax and YMin is the same!,
What can I do? 



So the correct code to fix the first two problems  I think it may be like this 
```
open override func calcMinMaxY(fromX: Double, toX: Double)
         let indexFrom = entryIndex(x: fromX, closestToY: Double.nan, rounding: .down)
         let indexTo = entryIndex(x: toX, closestToY: Double.nan, rounding: .up)
          
          if indexTo < indexFrom { return }
          
       for i in indexFrom...indexTo {
            ...
       }


```

aelam:
I just checked the android version .
We are different here  

@Override
    public void calcMinMaxY(float fromX, float toX) {

        if (mValues == null || mValues.isEmpty())
            return;

        mYMax = -Float.MAX_VALUE;
        mYMin = Float.MAX_VALUE;

        int indexFrom = getEntryIndex(fromX, Float.NaN, Rounding.DOWN);
        int indexTo = getEntryIndex(toX, Float.NaN, Rounding.UP);

        for (int i = indexFrom; i <= indexTo; i++) {

            // only recalculate y
            calcMinMaxY(mValues.get(i));
        }
    }

    /**

aelam:
@xuan  @danielgindi  could you  please verify this

liuxuan30:
@danielgindi should be still busy.. Haven't seen him recently. As this is a foundermental change, we need to have @danielgindi to review.

aelam:
😒
I think it's easy to reproduce this bug, you can verify it. And it's very reasonable, also android code can support it. 


发自我的 iPhone

liuxuan30:
@danielgindi please take a look at this PR

liuxuan30:
@aelam I am taking this right now, but some thoughts:
>        let indexTo = entryIndex(x: toX, closestToY: Double.nan, rounding: .up)

This is rounding '.up', which means `indexTo` may larger by 1, which would cause oob issue.

I took a look at `open override func entryIndex(x xValue: Double, closestToY yValue: Double, rounding: ChartDataSetRounding) -> Int`, actually quite complicated than I expect.

If we just want to make sure it's ok, we have to check the boundary again. 
I guess that's why it uses `..<` in the first place. However, in order to make sure `...` works without crash, we need to verify that

liuxuan30:
should be safe to merge then with the analysis above and CI tests.

emmaviolet:
@danielgindi @petester42 are there any plans to release a new version with the latest changes? This is a useful bugfix 🙏

liuxuan30:
It's already merged, so you can easily build one, or take the change and make a patch

PR #2247, opened 2017-03-15
Repair the warning：Use of undeclared type 'CGFloat'; did you mean to …

dsencheng:
'import Foundation' replace by 'import UIKit'

### DIFF
--- a/Source/Charts/Animation/Animator.swift
+++ b/Source/Charts/Animation/Animator.swift
@@ -9,7 +9,7 @@
 //  https://github.com/danielgindi/Charts
 //
 
-import Foundation
+import UIKit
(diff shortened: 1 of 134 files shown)


thierryH91200:
@dsencheng 

Besides not compatibility with osx, this is not very useful

```
import Foundation
import CoreGraphics

#if !os(OSX)
    import UIKit
#endif
```

Becomes

```
import UIKit
import CoreGraphics

#if !os(OSX)
    import UIKit
#endif
```

dsencheng:
i get it

Yangchengfeng:
i run it for iOS，and i deal with 【'import UIKit' replace by 'import Foundation'】

PR #2275, opened 2017-03-24
Adds a property to check whether to draw grid lines or not

FrankKair:
This simple pull request aims to make possible to choose whether to show the grid lines behind the Bar chart or not, since sometimes (for design purposes) it's desired a "cleaner" chart.


liuxuan30:
already have
```
    open override func renderGridLines(context: CGContext)
    {
        guard
            let xAxis = self.axis as? XAxis,
            let transformer = self.transformer
            else { return }
        
        if !xAxis.isDrawGridLinesEnabled || !xAxis.isEnabled
        {
            return
        }
```

PR #2315, opened 2017-04-02
Update Cartfile

sohayb:
Update to latest Realm version, as it was causing `cathage update` to hang if a project is using the current version.
**[No breaking changes for Realm]**

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "realm/realm-cocoa" ~> 2.4
+github "realm/realm-cocoa" ~> 2.5.0


liuxuan30:
@petester42 how about we use `>= 2.4` to avoid upgrading issues? Do we have special requirement for Realm versions?

petester42:
I believe the version is already loose enough. Realm use semantic versioning so a bump to 2.5 means there are new features which we should test if our code still works.

liuxuan30:
OK, so we can close this for now. Thanks @sohayb but petester42 explained

liuxuan30:
Oh wait, I just came across RealmCocoa and found Realm 2.5 is for Swfit 3.1, 2.4 seems not work with 3.1. Since this repo has removed Realm dependency and move to https://github.com/danielgindi/ChartsRealm, we should update Realm over there instead.

petester42:
yes

liuxuan30:
@petester42 I just saw ChartsRealm has updated to 2.5 already. Shall we delete Realm here?

liuxuan30:
No idea why the diff shows 2.4 but the real cartfile is 0 bytes now. So it should be fine now, no more action to take

sohayb:
Thanks guys!

PR #2335, opened 2017-04-06
Compatibility with swift playgrounds

macteo:
- Added missing `import CoreGraphics` where needed.
- Added missing `import UIKit` where needed.

### DIFF
--- a/Source/Charts/Data/Interfaces/IChartDataSet.swift
+++ b/Source/Charts/Data/Interfaces/IChartDataSet.swift
@@ -10,6 +10,7 @@
 //
 
 import Foundation
+import CoreGraphics
 
 @objc
 public protocol IChartDataSet
--- a/Source/Charts/Components/Description.swift
+++ b/Source/Charts/Components/Description.swift
@@ -12,6 +12,10 @@
 import Foundation
 import CoreGraphics
 
+#if !os(OSX)
+    import UIKit
+#endif
+
 @objc(ChartDescription)
 open class Description: ComponentBase
 {
--- a/Source/Charts/Filters/DataApproximator.swift
+++ b/Source/Charts/Filters/DataApproximator.swift
@@ -10,6 +10,7 @@
 //
 
 import Foundation
+import CoreGraphics
 
 @objc(ChartDataApproximator)
 open class DataApproximator: NSObject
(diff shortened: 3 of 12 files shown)


petester42:
Neat idea. Maybe you could add a playground the repo and we could use that instead of the demo project. Good work 👌

liuxuan30:
@petester42 we cool to merge some PRs like this? I think @danielgindi is too busy to review all the PRs :)

BTW, Seeing Travis new error type:
>xcrun instruments -w '22FA2149-1241-469C-BF6D-462D3837DB72' || sleep 15
Instruments Usage Error: Unknown device specified: "22FA2149-1241-469C-BF6D-462D3837DB72"

Is this travis issue?

macteo:
I tried to solve the travis configuration error, unfortunately without success. Many PR are failing due to the same issue.

liuxuan30:
@macteo did you just try to replace the device id with any one in the list, for example,
>iPhone 7 (10.2) [34FA4749-C467-4D45-9F8E-E31AEDDC39C2] (Simulator)

petester42:
Ya that error is probably because some simulators got removed with Xcode 8.3. I'll look at this pull request soon and merge.

liuxuan30:
@petester42 I update the device id in #2378

Source/Charts/Data/Interfaces/IChartDataSet.swift
```diff
@@ -10,6 +10,7 @@
 //
 
 import Foundation
+import CoreGraphics
```
liuxuan30: It's interesting in my Xcode, it never warns or complains missing CoreGraphics, though IChartDataSet refer CGPoint class

liuxuan30: It's interesting in my Xcode, it never warns or complains missing CoreGraphics, though IChartDataSet refer CGPoint class

petester42: Since UIKit and AppKit imports CoreGraphics it is not needed.

macteo: Until you try to use in a Playground

liuxuan30: @macteo shall we consider it's playground bug?

macteo: I don't think so as it seems a strict requirement.

PR #2353, opened 2017-04-12
Small fix for calculating minimum and maximum ChartDataSet values not…

piotr-sekara:
Without this fix, ChartDataSet is calculating minimum and maximum values not for whole value array. The result is that the Bar (in my example) is outside the chart, because maximum value of the last object in array was skipped. 

Before this change : 
before

After this change :
after


liuxuan30:
dup to https://github.com/danielgindi/Charts/pull/2229

PR #2355, opened 2017-04-13
Bug fixing with one line, updating ChartViewBase.swift

Eric0625:
fixing a bug: when clearing chart, the 'lastHighlighted' var isn't set to nil. This will cause user to tap twice to highlight the new value, if the new value's index is the same as the old lastHighlighted's.

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -155,7 +155,10 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     
     /// flag that indicates if offsets calculation has already been done or not
     fileprivate var _offsetsCalculated = false
-    
+	
+    /// if `true` , when call clear(), also set lastHighlighted to nil
+    open var clearLastHighlightedEnabled = false
+	
     /// array of Highlight objects that reference the highlighted slices in the chart
     internal var _indicesToHighlight = [Highlight]()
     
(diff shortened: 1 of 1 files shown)


liuxuan30:
do you mean you called `clear()` first, and then highlight? What's the context for clearing? Have you tried just set the new chart data directly?
First time I just `clear()` today :)

Eric0625:
I tried to highlight the first bar in my barChart, then I press a button to remove old data and fill in new ones, but I find out you need to tap twice to highlight the first bar of new data.

And I don't know how to  "set the new chart data directly" without calling `clear()`.

```
open func clear()
    {
        _data = nil
        _offsetsCalculated = false
        _indicesToHighlight.removeAll()
        setNeedsDisplay()
    }
```

Function `clear()` first sets _data to nil, IMO that sort of means "set the new chart data directly". Then it called `_indicesToHighlight.removeAll()`, which deletes all highlighted slices in the chart, so what's the meaning to keep the `lastHighlighted` variable? It's now representing a non-exist highlighted object.

liuxuan30:
What I mean is just call `chartView.data = newData` to update your entire chart data without calling clear()

Eric0625:
That's exactly what I did when using the chart for the first time: only set `chartView.data` with new data. Then the app behaved like this:

This was the initial status:  I clicked the first bar


Then I hit the button on top of it to fill the chart with new data, but you can see the first bar of new data is automatically selected, which is definitely not what I want:


I've tried several times but it looks like that only to both call `chartView.lastHighlighted=nil` and `_indicesToHighlight.removeAll()` can you get the fully cleaned chart.

liuxuan30:
I see. The `lastHighlighted` is reserved by default, which seems by design. resetting chartData and clear it seems not get reported by others in my memory, but I am fine to put it in `clear()`

Eric0625:
Ok, I'll try it when I have time

liuxuan30:
looks good to me. we will try fix travis first, and later handle some simple PRs, thanks for you patience.

PR #2357, opened 2017-04-13
adding some protection for _matrixValueToPx property avoid some special case

wjacker:
Hi @danielgindi @liuxuan30 

I meet an strange crash issue in our project, I investigate this issue and find that may case by follow code in **prepareMatrixValuePx** method:

`_matrixValueToPx = _matrixValueToPx.translatedBy(x: CGFloat(-chartXMin), y: CGFloat(-chartYMin))`

The chartXMin and chartYMin may is ±inf in some case, cause the  _matrixValueToPx tx,ty to nan, so in next convert the xaxis label value to position to nan point, and then crashed on draw line. please see the screenshot:
alt tag
alt tag

So I just add and simple protection check before set _matrixValueToPx value, avoid _matrixValueToPx have incorrect value.  

Please notes, this issue only appear on my project, I can not reproduce in demo app with same code, very weird 😰 maybe I used autolayout case that issue.

Thanks,
-Jack


liuxuan30:
In the first place, why call `_matrixValueToPx = _matrixValueToPx.translatedBy(x: CGFloat(-chartXMin), y: CGFloat(-chartYMin))` and the value is inf? It's not supposed to do so.

If it's kind of customization of your modified code, then it's fine, but shall we support this case for the moment?

wjacker:
Hi @liuxuan30, the code not  call `_matrixValueToPx = _matrixValueToPx.translatedBy(x: CGFloat(-chartXMin), y: CGFloat(-chartYMin))` first, you can check my commit change. original code only check scaleX and scaleY inf case. but tx/ty maybe alos can inf the code don't check. if tx/ty is inf, will case the _matrixValueToPx value incorrect.

liuxuan30:
my bad :(

Source/Charts/Utils/Transformer.swift
```diff
@@ -35,19 +35,30 @@ open class Transformer: NSObject
...
+        if CGFloat.infinity == abs(scaleX)
         {
             scaleX = 0.0
         }
-        if CGFloat.infinity == scaleY
+        if CGFloat.infinity == abs(scaleY)
         {
             scaleY = 0.0
         }
+        if CGFloat.infinity == abs(tx)
```
jjatie: can you make all of these checks us `.isInfinite` instead. i.e.
```
if tx.isInfinite
{
    tx = 0.0
}
```

PR #2364, opened 2017-04-15
add example playground

thierryH91200:
In addition to #2335 with markup

It's very good as tutorial

capture d ecran 2017-04-15 a 20 45 48

capture d ecran 2017-04-15 a 20 46 44


liuxuan30:
wow, thanks so much for your various help and contribution

petester42:
This looks very cool! This should remove people asking `why isn't there a swift demo`.

liuxuan30:
@petester42 swift demo support involves several PRs, do you think we could manage to merge them to reduce @danielgindi overhead?

thierryH91200:
@petester42 
II think this is not a good idea
this is swift 3
see pieChart
it is better to remove it
and I return a PR in swift 4

thierryH91200:
and other problem

Playgrounds

Known Issues
macOS playground execution may hang in the Running state with no results produced. This is more prevalent when changing pages. (32429211)
Workaround: Stop playground execution and rerun. If issue persists, close and reopen Xcode.

The liveTouchBar property of PlaygroundSupport is not supported. (31073754)

petester42:
Works for me. Swift 3 or 4 doesn't matter the interface of library is the same.

petester42:
If you want to make improvements then go ahead but it works with current master.

thierryH91200:
ok thanks

PR #2365, opened 2017-04-16
Fix Simple Bar Chart Demo, switch use of x and y values

franqueli:
The demo broke in the following commit fbba5156196ffa704ee52ac94c9c3a62f23ebd16
The x and y values were switched.


liuxuan30:
Thanks! I saw bar chart behaves weird, and just came across this PR!

PR #2377, opened 2017-04-20
fix #2356 crash if floor(10.0 * intervalMagnitude) is 0.0

### LINKED ISSUE #2356: Crash in YAxisRendererRadarChart
Crashes on line 137: axis.decimals = Int(ceil(-log10(interval)))

When interval is 0.
In my case this is happening when rawInterval is 0.05, interval then gets set to 0, and -log10(0) results in NaN which causes ceil to crash.

liuxuan30:
For #2356 

This is a corner case when `interval` and `intervalMagnitude` meet certain conditions, like `interval=0.06` and `intervalMagnitude=0.01` will lead `floor(10.0 * intervalMagnitude)` to `0.0`, and crash at `axis.decimals = Int(ceil(-log10(interval)))`

I added a check if it's 0, then we keep the current interval. 

@danielgindi Need you to review this as this is a fundamental change.
The problem is I don't know whether to keep the interval value, or we use `intervalMagnitude * 5` such as to get 0.05 rather than 0.06 or 0.08.

Or, do we have other options?

### DIFF
--- a/Source/Charts/Renderers/AxisRendererBase.swift
+++ b/Source/Charts/Renderers/AxisRendererBase.swift
@@ -127,7 +127,8 @@ open class AxisRendererBase: Renderer
         if intervalSigDigit > 5
         {
             // Use one order of magnitude higher, to avoid intervals like 0.9 or 90
-            interval = floor(10.0 * Double(intervalMagnitude))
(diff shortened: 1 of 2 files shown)


liuxuan30:
@danielgindi have sometime to review this? Should not take long :)

PR #2390, opened 2017-04-25
Added drawGridLinesOnTopEnabled boolean flag to draw grid lines on top of Bar Charts

JacopoMangiavacchi:
I had this UI requirements for a projects and added this boolean flag, and relative behaviour, to draw the grid lines above the Bar Charts

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -198,16 +198,16 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             _xAxisRenderer?.computeAxis(min: _xAxis._axisMinimum, max: _xAxis._axisMaximum, inverted: false)
         }
         
-        if !isdrawGridLinesOnTopEnabled {
-            _xAxisRenderer?.renderAxisLine(context: context)
-            _leftYAxisRenderer?.renderAxisLine(context: context)
-            _rightYAxisRenderer?.renderAxisLine(context: context)
-        }
+        _xAxisRenderer?.renderAxisLine(context: context)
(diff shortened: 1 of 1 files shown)


liuxuan30:
What's this for? I don't understand. Any screenshot to demonstrate?  From the code I don't see the difference for grid line in front or background?

JacopoMangiavacchi:
Sure thing @liuxuan30 , an image is better than 1000 words ;-)

grid line on top

liuxuan30:
😂 however I can't see the grid line difference?

JacopoMangiavacchi:
Oh, maybe there's another way but I didn't found an existing settings for having the (white) vertical lines draw on top of the (blue/orange) horizontal lines like in the picture above. I just added this setting and in the draw() method I moved the drawing of these lines after the bar lines if this property is true.

liuxuan30:
I just don't see any grid line in your screenshot. And have no idea what's front and behind, since grid line postion and the `contentRect` should not have intersect. That's why I am asking.

JacopoMangiavacchi:
Yes @liuxuan30 . I'm not a designer but as said I just had this UX request to implement a chart like this, with bars cutted off in small rect by the grid line. Other charts frameworks have this feature but I preferred to use Charts so I've implemented this.

JacopoMangiavacchi:
Btw, of course this is just a new optional configuration parameter (drawGridLinesOnTopEnabled) where the default value (false) draw a  traditional chart with grid lines below the bar line.

liuxuan30:
Sure, I see. I will leave this PR to @danielgindi , as I'm not sure if we should add this. I am fine, but adding more 'switches' also needs to be careful

jjatie:
Grid lines should be behind the data. If this is a requirement for your project, that's fine, but it's not something for the shared framework.

PR #2393, opened 2017-04-26
Minor improvements in BalloonMarker.swift

valeriyvan:
Getting rid of not needed String-NSString bridging


ChartsDemo/Classes/Components/BalloonMarker.swift
```diff
@@ -51,10 +51,7 @@ open class BalloonMarker: MarkerImage
     
     open override func draw(context: CGContext, point: CGPoint)
     {
-        if labelns == nil
-        {
-            return
-        }
+		guard let label = label else { return }
```
liuxuan30: seeing more indent here, please double check

Issue #2316, opened 2017-04-02
Too much Warnings on the comments


buisinam:
How can remove warning "Empty paragraph passed tp \params command" ?

liuxuan30:
I don't know where this is happened. Reopen when you have more details

Issue #135, opened 2015-06-08
Warnings on the comments


noais:
The generated file Charts-Swift.h have warnings on all comments like:

"Charts-Swift.h:149:20: Empty paragraph passed to '\param’ command”.

Every param is not described producing warnings.

danielgindi:
It's probably some configuration in your project, this is not the default behaviour

noais:
Yes, I'm using clang static analyser. I'm going to exclude this file from this analise. Thanks

michael-snapptec:
Can someone give some more information about this? I'm having the same issue, but all of the "solutions" I see would involve an edit on the code of the Charts library itself. I'd love to know what the setting is in my project that is causing this. I have this same issue in two related projects, but one of them is newly created, and I haven't monkeyed with the settings in Xcode or anything. Any ideas?

noais:
Have you some analyser like clang?

michael-snapptec:
I am assuming that Xcode is using clang. I'm using Xcode 8, and I haven't changed any static analyzer settings on my project.

ugenlik:
When I set `Apple Clang - Warnings -> Documentation Comments` to NO this warnings disappeared

PR #2399, opened 2017-04-27
Corrected multiple comments to use correct Swift syntax

elaewin:
Multiple comments in multiple files were commented with a triple slash (`///`) which is the correct syntax for C#, but not for Swift. This was causing multiple warnings in Xcode. 

A slash was removed where necessary to give the comments the correct double slash syntax for Swift comments.

### DIFF
--- a/Source/Charts/Animation/Animator.swift
+++ b/Source/Charts/Animation/Animator.swift
@@ -19,10 +19,10 @@ import CoreGraphics
 @objc(ChartAnimatorDelegate)
 public protocol AnimatorDelegate
 {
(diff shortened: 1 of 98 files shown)


thierryH91200:
**are you sure ?????**

capture d ecran 2017-04-27 a 20 03 08

 on a function with Xcode
capture d ecran 2017-04-27 a 20 12 23

and the result
capture d ecran 2017-04-27 a 20 12 34

elaewin:
The triple slashes for document markup are what is causing the errors in issues #135, and #2316.

PR #2413, opened 2017-05-05
Allow turning off drag in X and Y axes separately.

maciejtrybilo:
Currently if a chart is inside a scroll view and drag is enabled, the chart pan gesture recognizes all pan events and the containing scroll view doesn't work. 

This PR allows enabling/disabling drag in either axis separately. Probably the most common use case would be enabling scale only in the X direction and enabling drag only in the X direction, so a vertically scrolling scroll view can work normally:
```
chartView.scaleYEnabled = false
chartView.scaleXEnabled = true
chartView.dragYEnabled = false
```

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -29,6 +29,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     fileprivate var _pinchZoomEnabled = false
     fileprivate var _doubleTapToZoomEnabled = true
     fileprivate var _dragEnabled = true
+    fileprivate var _dragYEnabled = true
+    fileprivate var _dragXEnabled = true
     
     fileprivate var _scaleXEnabled = true
     fileprivate var _scaleYEnabled = true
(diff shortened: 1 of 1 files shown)


hernanarber:
+1 on that Request

kashifshaikh:
+1 - I made a similar change, but your is much better!

liuxuan30:
wow I missed this.. Will try to review and merge

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -29,6 +29,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     fileprivate var _pinchZoomEnabled = false
     fileprivate var _doubleTapToZoomEnabled = true
     fileprivate var _dragEnabled = true
+    fileprivate var _dragYEnabled = true
```
liuxuan30: Shall we change to `false` by default? Ideally this should be set by user explicitly

maciejtrybilo: `dragEnabled` is true by default. Having `dragYEnabled` false by default would change the current default behaviour.

liuxuan30: oops, my bad. Lots of `||` and `!`..

liuxuan30:
waiting for Travis results;

danielgindi:
Seems like you are disabling the beginning of the drag gesture if it's in the wrong direction,
but then changing directions in the middle of a gesture is allowed... Am I missing something?

maciejtrybilo:
I've only tested it when the chart is in a scroll view itself, so this didn't come up. Thanks!

liuxuan30:
@danielgindi you are back 😃

danielgindi:
Hey I'm always here, watching over you ;-)
Since I have a sweet little daughter, my time distribution is different.
Now she entered a little kindergarden so I can login sometimes at nights and do some maintenance and improvements.
By the way - you and @petester42 are doing such a great job. It's great that you're taking care of this when I can't!

liuxuan30:
Thanks! Hope your daughter sleep well. My 8-month boy would wake up several times in the evening, kind of killing me.. haha
still, some feature PRs need you help, especially when thinking about android. We are able to help review and merge some:)

PR #2429, opened 2017-05-11
Adds a property to disable the pie chart values inside the chart

FrankKair:
Context:

Uses the variable `_drawEntryLabelsEnabled`  to check whether to draw the values in the pie chart or not using `renderer!.drawValues(context: context)`.


liuxuan30:
it's controlled by 
`let drawValues = dataSet.isDrawValuesEnabled`
so you have fine-grained controls.

Issue #1830, opened 2016-11-14
A problem of CombinedChartView's YAxis


brxren:
I think there might be a problem with the YAxis of CombinedChartView.

I have a CombinedChartView with one bar and one line. 
The bar depends on the left YAxis.
The line depends on the right YAxis.
After the dependencies are well configured, I get the chart below, which is definitely not what I want.
The labels of the left YAxis and the right YAxis are the same even though y values are totally different.
I'm not sure if it's designed like that intentionally. If so, just ignore what I say below.

![simulator screen shot 2016 11 14 8 56 22](https://cloud.githubusercontent.com/assets/2941552/20265634/2cc3db94-aaad-11e6-98d2-45f0be4023e5.png)

I am confused, so I review the source code. Finally I find a problem of the function calcMinMax() in CombinedChartData.swift.

            if data.yMax > _leftAxisMax
            {
                _leftAxisMax = data.yMax
            }

            if data.yMin < _leftAxisMin
            {
                _leftAxisMin = data.yMin
            }

            if data.yMax > _rightAxisMax
            {
                _rightAxisMax = data.yMax
            }

            if data.yMin < _rightAxisMin
            {
                _rightAxisMin = data.yMin
            }

It seems that dataset's ydependency is NOT considered at all. So I change it.

            for i in 0 ..< sets.count
            {
                let dataset = sets[i]
                if dataset.axisDependency == .left
                {
                    if dataset.yMax > _leftAxisMax
                    {
                        _leftAxisMax = dataset.yMax
                    }
                    if dataset.yMin < _leftAxisMin
                    {
                        _leftAxisMin = dataset.yMin
                    }
                } else { // right
                    if dataset.yMax > _rightAxisMax
                    {
                        _rightAxisMax = dataset.yMax
                    }
                    if dataset.yMin < _rightAxisMin
                    {
                        _rightAxisMin = dataset.yMin
                    }
                }
            }

Now I get what I want.

![simulator screen shot 2016 11 14 8 55 02](https://cloud.githubusercontent.com/assets/2941552/20265620/1b0a0fcc-aaad-11e6-86e2-191e1bd94365.png)

I think it might be helpful to someone like myself.

liuxuan30:
To me, I would say setting up `axisDependency` can solve it.

brxren:
I've set up axisDependency, but it seems it doesn't work. As far as I know, it doesn't work in CombinedChartView. Other charts work well.

[...]

PR #2433, opened 2017-05-12
fix issues #1830

zysSS:
I have the same problem as issue #1830,this problem remains unsolved，the CombinedChartData is unable to correctly calculate _leftAxisMax and  _rightAxisMax. 
this commit fix this problem.


liuxuan30:
closing. Thanks for your help. I will  take care this in https://github.com/danielgindi/Charts/pull/2874 (this has conflict and using 2874 is easier for me)

PR #2439, opened 2017-05-14
remove an argument not necessarily required

sidepelican:
```ChartViewBase.swift
open func save(to path: String, format: ImageFormat, compressionQuality: Double) -> Bool
```

`compressionQuality` is not used for .png Format.
I think it is better to use Associated Value.

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -855,7 +855,8 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     
     public enum ImageFormat
     {
-        case jpeg
+        /// - parameter quality: compression quality for lossless formats (JPEG)
+        case jpeg(quality: Double)
         case png
     }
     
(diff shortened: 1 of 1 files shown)


liuxuan30:
For those don't not familiar with swift associated values(like me): [Associated Values](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID148)

sidepelican:
That makes sense. Thanks for your review.

liuxuan30:
Why close??

sidepelican:
Sorry, I misunderstood that you think this change is not need because of not kind for swift beginner.

liuxuan30:
sure thing. I think the new style looks neat, but not tracking new swift features for a while, so for whoever like me, they can find a link to understand
BTW, you can add that link in your commits so anyone can see if they don't understand and find your commit log.

sidepelican:
Omg, I forgot about objective-c 😱
This is not available.

https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html#//apple_ref/doc/uid/TP40014216-CH4-ID53
`Enumerations defined in Swift without Int raw value type`

liuxuan30:
what a pity...

PR #2449, opened 2017-05-17
remove an argument not necessarily required (from #2439)

sidepelican:
```ChartViewBase.swift
open func save(to path: String, format: ImageFormat, compressionQuality: Double) -> Bool
```

`compressionQuality` is not used for .png Format.
I think it is better to use Associated Value.

(this is from #2439)

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -869,17 +869,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     ///
     /// - returns: `true` if the image was saved successfully
     open func save(to path: String, format: ImageFormat) -> Bool
-    {
-        let saveUrl = URL(fileURLWithPath: path)
-        if #available(iOS 9.0, *)
-        {
-            if saveUrl.hasDirectoryPath
-            {
-                // the path represents a directory. not include filename.
-                return false
-            }
-        }
-        
+    {   
         var imageData: Data?
         switch (format)
         {
@@ -896,7 +886,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         {
             do
             {
-                try imageData.write(to: saveUrl, options: .atomic)
+                try imageData.write(to: URL(fileURLWithPath: path), options: .atomic)
             }
             catch
             {
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -871,10 +871,13 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     open func save(to path: String, format: ImageFormat) -> Bool
     {
         let saveUrl = URL(fileURLWithPath: path)
-        if saveUrl.hasDirectoryPath
+        if #available(iOS 9.0, *)
         {
-            // the path represents a directory. not include filename.
-            return false
+            if saveUrl.hasDirectoryPath
+            {
+                // the path represents a directory. not include filename.
+                return false
+            }
         }
         
         var imageData: Data?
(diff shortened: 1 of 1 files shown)


thierryH91200:
Chart is for macOS, iOS and tvOS
@sidepelican 
fix travis compile error for iOS ????

sidepelican:
error has only occurred on iOS build.
I had used newer method for iOS, so I add guard about iOS version.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -855,47 +855,59 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
     /// - parameter format: the format to save
-    /// - parameter compressionQuality: compression quality for lossless formats (JPEG)
     ///
     /// - returns: `true` if the image was saved successfully
-    open func save(to path: String, format: ImageFormat, compressionQuality: Double) -> Bool
+    open func save(to path: String, format: ImageFormat) -> Bool
     {
-		guard let image = getChartImage(transparent: format != .jpeg)
-            else { return false }
+        let saveUrl = URL(fileURLWithPath: path)
```
liuxuan30: What is this for? I think we can lower the requirement not to use 9.0 method?

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -855,47 +855,59 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+                return false
+            }
+        }
         
-        var imageData: Data!
+        var imageData: Data?
         switch (format)
         {
         case .png:
+            guard let image = getChartImage(transparent: true) else { return false }
```
liuxuan30: Want to know why move image here?

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -855,47 +855,59 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
             
-        case .jpeg:
+        case .jpeg(let compressionQuality):
+            guard let image = getChartImage(transparent: false) else { return false }
             imageData = NSUIImageJPEGRepresentation(image, CGFloat(compressionQuality))
-            break
         }
         
-        do
+        if let imageData = imageData
```
liuxuan30: why use `imageData: Data?` instead of `imageData: Data!` here? Any advantages?

Source/Charts/Charts/ChartViewBase.swift
sidepelican: `data.write(to: url)`  dose not throw an error when the path represents a directory(and do not make any files). This is for that.
But now,I think this seemed excessive.

Source/Charts/Charts/ChartViewBase.swift
sidepelican: old way, `.jpeg != format` is not available because of associated value enum. Those enum does not have `Equatable` (We have to define by myself).

Source/Charts/Charts/ChartViewBase.swift
sidepelican: Both `NSUIImagePNGRepresentation` and `NSUIImageJPEGRepresentation` returns `Data?`. 
I thought we need to check `nil`.

sidepelican:
petester42 taught me, associated value enum cannot use from Objective-C😱. I forgot this.

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: But we have try catch right? If it's nil, it will be caught.

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: I see, I don't realize `.jpeg` is gone here.

liuxuan30:
It's a nice change but considering Objective-C...  awkward

Source/Charts/Charts/ChartViewBase.swift
sidepelican: Swift's try ~ catch dose not handle forced unwrapping error.
on Playground, macOS 10.12.4, Xcode 8.3.2, Swift 3.1

liuxuan30: oh I see. My bad

PR #2477, opened 2017-05-25
add multi axis

thierryH91200:
I needed 4 axes because I had data in very different scales and the problem to internationalize the data implied great variations
Celcius in farenheit
Miles in km
The variations were not compatible with 2 axes and 4 curves
As I knew it did not exist it's done
Now for me it's ok otherwise it has not been easy

capture d ecran 2017-05-25 a 18 47 19

capture d ecran 2017-05-25 a 17 57 41

capture d ecran 2017-05-25 a 17 56 41

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -29,6 +29,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     fileprivate var _pinchZoomEnabled = false
     fileprivate var _doubleTapToZoomEnabled = true
     fileprivate var _dragEnabled = true
+    fileprivate var _dragYEnabled = true
+    fileprivate var _dragXEnabled = true
     
     fileprivate var _scaleXEnabled = true
     fileprivate var _scaleYEnabled = true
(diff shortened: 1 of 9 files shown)


liuxuan30:
Where is @danielgindi 😂 Someone is pulling features and pending

thierryH91200:
I forgot the use of: extraTopOffset, extraBottomOffset, extraLeftOffset, extraRightOffset

thierryH91200:
Request help


I have a little problem
let me explain
Charts consists of a graph and these components
- Legend
- LimitLine
- Description
- Marker
- xAxis
- yAxis
- _pinchGestureRecognizer
- _panGestureRecognizer
- etc ....

On my side I have added 2 secondary axes

The problem is that all these components are true by default. Is that it would have been better if some were true and some false by default and the user adds these components one by one or removes them


```
Open class ComponentBase: NSObject
{

// flag that indicates if this component is enabled or not
Open var enabled = true

Open var isEnabled: Bool {return enabled}
}
```

I would like the 2 secondary axes to be false by default, but I can not use anything other than enabled and isEnabled, as it is a component

Otherwise adding a variable of the type "axisSecondaryEnabled" this causes to modify a lot of code in many class
- YAxisRenderer
- YAxisRendererHorizontalBarChart
- BarLineChartViewBase
- etc

I do not know what to do
- many code changes but nothing is impossible with a risk of error
- do you have another great idea that I did not think ??

thierryH91200:
update with axisSecondaryEnabled = false
enabled is not used

DenZuck:
I also need this feature. When will it available?

PR #2507, opened 2017-06-08
Changes for Swift 4

liuxuan30:
Thanks to Swift 4's compatibility, this year is easier to upgrade on our own pace. But still works to do :(

## Part I 
modify All `NSAttributedStringKey` related methods for Swift 4

#### Note:
- `extension NSString for func size(attributes attrs: [String : Any]? = nil) -> NSSize`
is deleted due to iOS 11 uses `size(withAttributes: attrs)` now

## Known issues:
- There are tons of warnings about `@objc` usage, a new enhancement to reduce binary size since Swift 4, which is quite confusing, still waiting for newer Xcode 9 beta to see if it can help and investigating a better solution. Related SE: [SE-0160](https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md). One [SO Thread](https://stackoverflow.com/questions/44379348/the-use-of-swift-3-objc-inference-in-swift-4-mode-is-deprecated) discuss it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,7 @@
 language: objective-c
-osx_image: xcode8.3
+osx_image: xcode9
 matrix:
   include:
-    - osx_image: xcode8.3
-      env: PLATFORM="iOS"
-    - osx_image: xcode8.3
-      env: PLATFORM="tvOS"
-    - osx_image: xcode8.3
-      env: PLATFORM="macOS"
     - osx_image: xcode9
       env: PLATFORM="iOS"
     - osx_image: xcode9
--- a/Charts.xcodeproj/project.pbxproj
+++ b/Charts.xcodeproj/project.pbxproj
@@ -1125,32 +1125,6 @@
 				SDKROOT = macosx;
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
 				TARGETED_DEVICE_FAMILY = "1,2,3,4";
-				TVOS_DEPLOYMENT_TARGET = 9.0;
-				VALIDATE_PRODUCT = YES;
-				VERSIONING_SYSTEM = "apple-generic";
-				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
-			};
-			name = Release;
-		};
-		06165F391D8110E700722320 /* Debug */ = {
-			isa = XCBuildConfiguration;
-			buildSettings = {
-				APPLICATION_EXTENSION_API_ONLY = YES;
-				COMBINE_HIDPI_IMAGES = YES;
-				DEFINES_MODULE = YES;
-				DYLIB_COMPATIBILITY_VERSION = 1;
-				DYLIB_CURRENT_VERSION = 1;
-				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				FRAMEWORK_VERSION = A;
-				INFOPLIST_FILE = "$(SRCROOT)/Source/Supporting Files/Info.plist";
-				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = com.dcg.Charts;
-				PRODUCT_NAME = "$(TARGET_NAME)";
-				SKIP_INSTALL = YES;
-				SWIFT_SWIFT3_OBJC_INFERENCE = On;
-				SWIFT_VERSION = 4.0;
 			};
 			name = Debug;
 		};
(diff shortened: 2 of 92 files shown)


coty10:
Any news on swift 4 version?

liuxuan30:
This is the least working change now. Waiting for Xcode 9 release and finalize it

coty10:
so basically not before September.... 
i'm working on it in Xcode 9... i just have the language warning.. but everything else it's working... hopefully when the final swift 4 version will be released... i won't have to change all the code... but just use a nice pod install... :)
Thanks for the answer..

liuxuan30:
the warning is because the @objc tag. looks like we have to add @objc for class scope, but this dims the idea it asks for. Anyway, based on last year experience, just wait the GM

jjatie:
To clarify, SE-0160 doesn't make any recommendations as to where to put `@objc` tags, only that you now need to put them everywhere you want Objective-C compatibility (which Charts already does).

liuxuan30:
@jjatie here: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html#//apple_ref/doc/uid/TP40014097-CH35-ID348
>Most code should use the objc attribute instead, to expose only the declarations that are needed. If you need to expose many declarations, you can group them in an extension that has the objc attribute. This attribute is a convenience for libraries that make heavy use of the introspection facilities of the Objective-C runtime. Applying the objc attribute when it isn’t needed can increase your binary size and adversely effect performance.

it seems do not suggest add `@objc` to class scope,  but suggest "to expose only the declarations that are needed", that's what I mean. Currently if we add @objc before class, we loose the optimization? Yes eventually we might have to do this way, but I "wish" something can change when Xcode GM 😹

mrafayfarooq:
any update on Swift 4 conversion?

jjatie:
On any class that inherits from `NSObject` we don't lose any optimizations. We also need to use `@objc` at the type scope to provide an Objective-C name.

jjatie:
@mrafayfarooq if you're looking for a Swift 4 version, there is a fully functional one [here](https://github.com/jjatie/Charts) without objective-c compatibility. I haven't run into any bugs yet, and am migrating more and more to a "pure Swift" implementation, keeping the same API until at least Swift 5.

petester42:
If CI passes then this should be good to go!

liuxuan30:
@petester42 finally we have to add `@objc` for each func and vars? How about adding it before class (though it's not suggested)?

liuxuan30:
@petester42 we almost add `@objc` for each func and var. How about add `objcMembers` for the class?
According to
>objcMembers
    Apply this attribute to any class declaration that can have the objc attribute. The objc attribute is implicitly added to Objective-C compatible members of the class, its extensions, its subclasses, and all of the extensions of its subclasses.

>Most code should use the objc attribute instead, to expose only the declarations that are needed. If you need to expose many declarations, you can group them in an extension that has the objc attribute. The objcMembers attribute is a convenience for libraries that make heavy use of the introspection facilities of the Objective-C runtime. Applying the objc attribute when it isn’t needed can increase your binary size and adversely affect performance.

petester42:
It will increase the binary size but the functionality will be exactly the same as someone using swift 3. So it should not break someone's objective-c code written with an older version of the library.

FasterThanLlamas:
Is there going to be an update to the CocoaPod?

petester42:
When this is merged yes.

osianSmith:
When is this likely to be merged with main branch?

bardonadam:
Can we merge this now that XCode 9 is available?

liuxuan30:
I will check with latest Xcode9 and merge later if no issue found

liuxuan30:
Build via Xcode9 and no buildtime errors
merging
BTW, I made a legacy branch for current master branch, if anyone complains about Swift 4 upgrade, use the legacy branch https://github.com/liuxuan30/Charts/tree/Swift3.2-legacy

danielgindi:
Are all these `@objc` tags really required?

jjatie:
Many of them are not. Only those types, properties, and methods that consumers of the framework need access to in objective-c require the `@objc` tags. It is tedious to go through the framework and make the decision on each property.
For example: `ChartAnimator`'s methods are exposed to the consumer through `ChartViewBase`, therefor `ChartAnimator` doesn't need any `@objc` tags.

jjatie:
I may be wrong on this one, but another good example is all of the `ChartRenderer` classes don't seem to be meant for the consumer to use. In which case, not only can we remove the `@objc` tags on them, but also make them internal types.

danielgindi:
`ChartRenderer` actually are supposed to be subclasses as necessary...

danielgindi:
Maybe Apple will do this automatically in the future as a refactoring feature.
And maybe we will phase out objc support.... I don't know.

petester42:
They give you a option when migrating to swift 4 but picking the one with less caused too many errors. You can toggle it off in the project settings and it will give warnings for the onces that aren’t needed.

jjatie:
The documentation on the changes to `@objc` inference:
https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md

PR #2509, opened 2017-06-08
fix #2356

### LINKED ISSUE #2356: Crash in YAxisRendererRadarChart

pouriaalmassi:
## Description
- Step 2 of 2 for fixing the bug detailed at https://cha-ching.atlassian.net/browse/DEV-4680

> When `floor(10.0 * intervalMagnitude)` is 0.0, we use the old value to avoid crash at `axis.decimals = Int(ceil(-log10(interval)))`

https://github.com/danielgindi/Charts/pull/2377


pouriaalmassi:
Accidentally pushed this branch. Please ignore.

Issue #1631, opened 2016-10-10
How to make points interactive/draggable


3ph:
Hi, I'm looking at using ios-charts for displaying bezier curves (charts). I would like to be able to adjust to control points interactively. Can somebody please point me into the right direction? What needs to be overriden, etc? Thanks

liuxuan30:
hmm it's challenging. If I were you, I would do a prototype based on CoreGraphics, and if you do need the cool features in this library, then you have to walk through the library and find a way to integrate them.

PR #2522, opened 2017-06-15
Interactive chart view (Issue #1631)

3ph:
Initial idea linked to Issue #1631 though since that the Bezier bit has been dropped.

Interactive chart view makes charts interactive:
- chart entries (values) can be dragged around the chart
- data set can be (un)selected as well (currently calculation based on straight lines)
- added demo to demonstrate the chart view
- new flag determining if value can be selected without selecting the data set first

Added two new methods to the delegate:
- chartValueMoved - called when entry was moved
- chartDataSetSelected - called when data set was (un)selected

This might not ready to be included in the Charts master but still somebody might find it useful.


liuxuan30:
wow, thank you for your effort. This is a huge work and need a careful review from the author. However he has been busy with other stuff for quite a time. So please wait

acegreen:
Great work @3ph! You should assign @danielgindi as reviewer.

liuxuan30:
actually, any Collaborator is fine to review. But you have to solve the conflict.

I'm sorry you waited so long :( lack of man power

anhdle14:
Recently come across this feature, i have a very bad time to implement what #2522 's PR is. 

My solution is to implement `ChartDataEntryBase/ChartDataEntry` implement the `NSItemProviderReading/Writing`, `NSObject`, & `Codable` to get drag and drop. At least that is my assumption. 

Is this a good approach or not? I really want this feature, so please help me understand how to solve this problem. I will try to submit a PR after implementing it successfully.

PR #2568, opened 2017-06-29
Avoid passing NaN to CoreGraphics API (Fixes #1626)

### LINKED ISSUE #1626: Charts 3.0 - Error: this application, or a library it uses, has passed an invalid numeric value (NaN, or not-a-number) to CoreGraphics API and this value is being ignored
Hi Charts Team,

I'm currently updating my project from Switch 2.2 to 3.0 using the master branch of Charts via Cocoapods.

So far I don't meet big issues to update, but I just wanted to report these warnings from my console logs. They are emitted from the Charts framework, and even if they are not blocking me in my update, I thought it would be useful for you if I report them.

`<Error>: Error: this application, or a library it uses, has passed an invalid numeric value (NaN, or not-a-number) to CoreGraphics API and this value is being ignored. Please fix this problem.
If you want to see the backtrace, please set CG_NUMERICS_SHOW_BACKTRACE environmental variable.`

`<Error>: If you want to see the backtrace, please set CG_NUMERICS_SHOW_BACKTRACE environmental variable.
 CGContextAddLineToPoint: no current point.
<Error>: CGContextAddLineToPoint: no current point.`

PS: If you are aware of these warning you can just close this issue.

Good luck with portage, your framework is awesome! 👍

chiahan1123:
Fix issue #1626 
The `Highlight` object generated in the `ChartViewBase` class contains a `Double.NaN` y-value when calling the `highlightValue` functions that does not require a y-value.

In the `LineChartRenderer` class, the `drawHighlighted` function passes the NaN y-value into the `Transformer` `pixelForValues` to apply the `valueToPixelMatrix`, which will result in the x-value being a NaN; therefore, causing the `LineScatterCandleRadarRenderer` `drawHighlightLines` to move to a CGPoint with a NaN x-value.

To fix this, the y-value being passed into `pixelForValues` should be from the `entryForXValue`.

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -746,8 +746,8 @@ open class LineChartRenderer: LineRadarRenderer
                 context.setLineDash(phase: 0.0, lengths: [])
             }
             
-            let x = high.x // get the x-position
-            let y = high.y * Double(animator.phaseY)
+            let x = e.x // get the x-position
+            let y = e.y * Double(animator.phaseY)
             
             if x > chartXMax * animator.phaseX
             {


jjatie:
Thank you for the description! I'd like to merge this this week.

I'm unclear as to why we should be using the y value from `entryForXValue` instead of the one from the `Highlight`. @chiahan1123 @liuxuan30 is it safe to assume that `let e = set.entryForXValue(high.x, closestToY: high.y)` will always give the coordinates we want? I haven't looked into highlighting much, but it seems like this should be true. If I can get this verified, I can merge right away.

liuxuan30:
I think we should first figure out why there is NaN during the process, rather than bypass it.
Any NaN directly reflects something wrong with the data, or the interval(min/max value) we calculated based on the data, otherwise it should be safe.
NaN generated by user's invalid data should not be our responsibilities, they should be warned. The interval/min/max value we have is our things, like the one we merged: https://github.com/danielgindi/Charts/pull/2377

liuxuan30:
Hi there, sorry it took a long time to realize this is a good catch! Thank you!

PR #2604, opened 2017-07-13
Adding a third party tutorial

osianSmith:
I created a tutorial along with putting code on GitHub and was hopping to have it publicised?


liuxuan30:
Any highlight for creating the line chart?

osianSmith:
This was written was my use case for charts was for line charts and I found it a little hard to figure how to do them out at first with the demo being in object-c. 
Certain use cases need different graphs and I thought I could help out on creating one for line graphs, such as showing in-between recordings a statistical estimate without a artificial intelligence being taught (such as how many people are likely to be in a shop if you take samples every 15 minutes).

PR #2668, opened 2017-08-02
Axis ticks support

MatteoK:
This adds support for tickmarks in the x and y axis


codigodelsur:
This feature will be added?

MatteoK:
I wanted to create this PR on my fork and accidentally created it here. If I find the time to do this the right way i'll reopen this PR, since the feature is useful imo

ikbenben:
Hi there, I'm guessing this is not possible yet? I was just looking for this feature :)

PR #2679, opened 2017-08-04
Reduce build time with minor reference refactor

xinranw:
Adding a `self.` reference to a instance variable reduced the compile
time significantly (over 10x improvement). A single line evaluation of a
variable was further broken out into separate steps to reduce the
compile time as well.

This cuts at least 2 seconds from the build time

Compile times were analyzed with the following bash command:
`xcodebuild -scheme Charts -sdk iphonesimulator PLATFORM_NAME=iphonesimulator clean build OTHER_SWIFT_FLAGS="-Xfrontend -debug-time-function-bodies" 2>&1 | grep -e [0-9][0-9].[0-9].ms | sort -nr`

Compile times before the change:
before-change

Compile times after the change:
after-change

### DIFF
--- a/Source/Charts/Renderers/ChartDataRendererBase.swift
+++ b/Source/Charts/Renderers/ChartDataRendererBase.swift
@@ -55,7 +55,6 @@ open class DataRenderer: Renderer
     {
         guard let data = dataProvider?.data
             else { return false }
-        
-        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (viewPortHandler?.scaleX ?? 1.0))
+        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (self.viewPortHandler?.scaleX ?? 1.0))
     }
 }
--- a/Source/Charts/Renderers/HorizontalBarChartRenderer.swift
+++ b/Source/Charts/Renderers/HorizontalBarChartRenderer.swift
@@ -610,8 +610,7 @@ open class HorizontalBarChartRenderer: BarChartRenderer
     {
         guard let data = dataProvider?.data
             else { return false }
-        
-        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (viewPortHandler?.scaleY ?? 1.0))
+        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (self.viewPortHandler?.scaleY ?? 1.0))
     }
     
     /// Sets the drawing position of the highlight object based on the riven bar-rect.


liuxuan30:
Hi there, thanks for the input. Do you have any doc or post could explain this change?

xinranw:
The part at the bottom of [this article](https://atelierclockwork.net/2016/02/26/improving-swift-build-times.html) references a similar problem, where chaining operations can sometimes cause dramatic increases in compile time (also referenced by an Apple build engineer at WWDC).

For the `self.` addition that actually accounted for the majority of the build time improvement...I honestly don't have a good answer. Based on articles such as [this](https://medium.com/@RobertGummesson/regarding-swift-build-time-optimizations-fc92cdd91e31) and [this](https://medium.com/swift-programming/swift-build-time-optimizations-part-2-37b0a7514cbe), it seems a lot of the optimizations that others have found can be somewhat arbitrary. I ended up just trying various suggestions and noticed that adding the `self.` made a 1000x improvement. I wonder if this change is somehow helping the compiler figure out the reference to `viewPortHandler`.

xinranw:
Bump

liuxuan30:
Thank you for your information. I'm fine with this PR. But if it helps, do we need to change the entire code base?

xinranw:
There are certainly other parts of the code that can be optimized but these two lines were taking the longest amount of time (over 2 seconds if being used as a cocoapod). So I only tackled these two as the easiest way to cut down a somewhat significant part of the build time.

Source/Charts/Renderers/ChartDataRendererBase.swift
```diff
@@ -55,7 +55,9 @@ open class DataRenderer: Renderer
     {
         guard let data = dataProvider?.data
             else { return false }
-        
-        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (viewPortHandler?.scaleX ?? 1.0))
+
+        let maxVisibleCount = CGFloat(dataProvider?.maxVisibleCount ?? 0)
+        let scale = self.viewPortHandler?.scaleX ?? 1.0
```
liuxuan30: I got `self` is improving the speed. But can we just uses self, without using two instants? like
```swift
return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (self.viewPortHandler?.scaleX ?? 1.0))
```
I assume it has the same effect?

xinranw: The `self` change does account for most of the improvements. Chained operations do tack on some build time increase but it's probably not significant for this case. Happy to go ahead and update with that change.

danielgindi:
Cool! Thanks!

PR #2687, opened 2017-08-07
Updated all `Bool` ivars to get rid of special Obj-C backing variables.

jjatie:
- There is no more confusion for `Bool` ivars in the exposed API, as there is only one exposed variable (i.e. no `pinchToZoomEnabled` and `isPinchToZoomEnabled`).
- All backing variables are `private`, not `fileprivate`. The backing variables should only be called in the getter/setter of the public ivar.
- All names internal to Charts have been updated for Swift naming conventions
- All names external to Charts have obj-c setters with the appropriate obj-c naming convention
- Almost all `open` access control on `Bool` ivars has been changed to `public`. This is what they should have been in the first place, as overriding all but a few of these ivars would result in unintended consequences.
- Minor update to demos to support changes.

There are still some names that I would like to be changed, but I left them alone as that is outside the scope of this PR.


petester42:
This would be a breaking change. I think we should hold off on this a bit until we make other breaking changes. If you have ideas on how to improve this to not make it a breaking change I'd be glad to hear them.

jjatie:
Unfortunately this needs be a breaking change. As you commented on a few other of my changes, it makes sense to wait until Swift 4/Charts (4?) is released.

jjatie:
Can we either merge this, or create a branch for breaking changes (3.1 or 4.0 I'm guessing?)

jjatie:
@petester42 This should be 4.0.0 I guess.

jjatie:
@petester42 While this change works, it would be ideal to be able to just declare
```
// in objc, this can be accessed as:
// BOOL x = isBool
// x.isBool = false
// [x setIsBool: false]
@objc var isBool: Bool
```
This removes the custom setter name `setBool:`, but given that dot notation for properties is quite prevalent in obj-c now, I think it's acceptable to lose the custom setter name. Thoughts?

danielgindi:
Seems reasonable to me to do this, for 4.0.
Please notice that some properties' names changed beyond the Bool issue, like "horizontalHighlightIndicatorEnabled " which was "drawHorizontalHighlightIndicatorEnabled".

One of the main goals of this library is being compatible with its Android counterpart, so we need to keep the naming close (while doing our best to conform to Swift standards).
Btw some things are named in multiple ways, as the Swift standard evolved and things changed. At first it was not clear how to do ObjC compatibility, and some things just couldn't be compatible and we needed bridging functions and variables.  
Now ObjC is much less relevant, as Swift is more widely accepted, so personally I worry about that less.

jjatie:
@danielgindi Great. I will reevaluate the names this week (maybe today) to try and get the most consistency with android and Swift possible.

danielgindi:
Wherever we can drop the custom objc getter/setters - it's best. They are just another thing that we need to manage and we don't want that...
You wouldn't believe what we had to do for interoperability with objc in Swift 1/2

liuxuan30:
That makes me recall this project is started with Swift 1.1, at that time, we have to do something weird from what you see now :)

jjatie:
This is a huge change, and as you can see many conflicts are building up. I will resubmit this to 4.0.0 later (likely as in pieces as part of other refactors) to make this more manageable.

PR #2698, opened 2017-08-09
unwrap optionals

russellbstephens:
an attempt to reduce the number of force unwraps

### DIFF
--- a/Source/Charts/Animation/Animator.swift
+++ b/Source/Charts/Animation/Animator.swift
@@ -103,14 +103,7 @@ open class Animator: NSObject
                 elapsed = duration
             }
            
-            if let _easingX = _easingX
-            {
-                phaseX = _easingX(elapsed, duration)
-            }
-            else
-            {
-                phaseX = Double(elapsed / duration)
-            }
+            phaseX = _easingX?(elapsed, duration) ?? Double(elapsed / duration)
         }
         
         if _enabledY
--- a/Source/Charts/Animation/Animator.swift
+++ b/Source/Charts/Animation/Animator.swift
@@ -68,39 +68,27 @@ open class Animator: NSObject
     
     @objc open func stop()
     {
-        if _displayLink != nil
-        {
-            _displayLink?.remove(from: RunLoop.main, forMode: RunLoopMode.commonModes)
-            _displayLink = nil
-            
-            _enabledX = false
-            _enabledY = false
-            
-            // If we stopped an animation in the middle, we do not want to leave it like this
-            if phaseX != 1.0 || phaseY != 1.0
-            {
-                phaseX = 1.0
-                phaseY = 1.0
-                
-                if delegate != nil
-                {
-                    delegate!.animatorUpdated(self)
-                }
-                if updateBlock != nil
-                {
-                    updateBlock!()
-                }
-            }
-            
-            if delegate != nil
-            {
-                delegate!.animatorStopped(self)
-            }
-            if stopBlock != nil
-            {
-                stopBlock?()
-            }
+        guard let _displayLink = _displayLink else {
+            return
         }
+        _displayLink.remove(from: RunLoop.main, forMode: RunLoopMode.commonModes)
+        self._displayLink = nil
+
+        _enabledX = false
+        _enabledY = false
+
+        // If we stopped an animation in the middle, we do not want to leave it like this
+        if phaseX != 1.0 || phaseY != 1.0
+        {
+            phaseX = 1.0
+            phaseY = 1.0
+
+            delegate?.animatorUpdated(self)
+            updateBlock?()
+        }
+
+        delegate?.animatorStopped(self)
+        stopBlock?()
     }
     
     fileprivate func updateAnimationPhases(_ currentTime: TimeInterval)
@@ -115,9 +103,9 @@ open class Animator: NSObject
                 elapsed = duration
             }
            
-            if _easingX != nil
+            if let _easingX = _easingX
             {
-                phaseX = _easingX!(elapsed, duration)
+                phaseX = _easingX(elapsed, duration)
             }
             else
             {
@@ -135,9 +123,9 @@ open class Animator: NSObject
                 elapsed = duration
             }
             
-            if _easingY != nil
+            if let _easingY = _easingY
             {
-                phaseY = _easingY!(elapsed, duration)
+                phaseY = _easingY(elapsed, duration)
             }
             else
             {
@@ -151,15 +139,10 @@ open class Animator: NSObject
         let currentTime: TimeInterval = CACurrentMediaTime()
         
         updateAnimationPhases(currentTime)
-        
-        if delegate != nil
-        {
-            delegate!.animatorUpdated(self)
-        }
-        if updateBlock != nil
-        {
-            updateBlock!()
-        }
+
+        delegate?.animatorUpdated(self)
+
+        updateBlock?()
         
         if currentTime >= _endTime
         {
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -233,7 +233,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         
         context.restoreGState()
         
-        renderer!.drawExtras(context: context)
+        renderer?.drawExtras(context: context)
         
         if _xAxis.isEnabled && !_xAxis.isDrawLimitLinesBehindDataEnabled
         {
@@ -259,13 +259,13 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             context.saveGState()
             context.clip(to: _viewPortHandler.contentRect)
             
-            renderer!.drawValues(context: context)
+            renderer?.drawValues(context: context)
             
             context.restoreGState()
         }
         else
         {
-            renderer!.drawValues(context: context)
+            renderer?.drawValues(context: context)
         }
 
         _legendRenderer.renderLegend(context: context)
@@ -533,7 +533,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             
             let h = getHighlightByTouchPoint(recognizer.location(in: self))
             
-            if h === nil || h!.isEqual(self.lastHighlighted)
+            if h === nil || h?.isEqual(self.lastHighlighted) ?? false
             {
                 self.highlightValue(nil, callDelegate: true)
                 self.lastHighlighted = nil
@@ -753,7 +753,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 
                 if ((h === nil && lastHighlighted !== nil) ||
                     (h !== nil && lastHighlighted === nil) ||
-                    (h !== nil && lastHighlighted !== nil && !h!.isEqual(lastHighlighted)))
+                    (h !== nil && lastHighlighted !== nil && !(h?.isEqual(lastHighlighted) ?? false)))
                 {
                     self.lastHighlighted = h
                     self.highlightValue(h, callDelegate: true)
@@ -931,7 +931,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             ))
         {
             var scrollView = self.superview
-            while (scrollView !== nil && !scrollView!.isKind(of: NSUIScrollView.self))
+            while (scrollView !== nil && !(scrollView?.isKind(of: NSUIScrollView.self) ?? false))
             {
                 scrollView = scrollView?.superview
             }
@@ -946,22 +946,17 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
 
             var foundScrollView = scrollView as? NSUIScrollView
             
-            if foundScrollView !== nil && !foundScrollView!.nsuiIsScrollEnabled
+            if !(foundScrollView?.nsuiIsScrollEnabled ?? true)
             {
                 foundScrollView = nil
             }
             
             var scrollViewPanGestureRecognizer: NSUIGestureRecognizer!
             
-            if foundScrollView !== nil
-            {
-                for scrollRecognizer in foundScrollView!.nsuiGestureRecognizers!
+            foundScrollView?.nsuiGestureRecognizers?.forEach { scrollRecognizer in
+                if let panGestureRecognizer = scrollRecognizer as? NSUIPanGestureRecognizer
                 {
-                    if scrollRecognizer.isKind(of: NSUIPanGestureRecognizer.self)
-                    {
-                        scrollViewPanGestureRecognizer = scrollRecognizer as! NSUIPanGestureRecognizer
-                        break
-                    }
+                    scrollViewPanGestureRecognizer = panGestureRecognizer
                 }
             }
             
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -163,11 +163,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     open override func draw(_ rect: CGRect)
     {
         super.draw(rect)
-        
-        if _data === nil
-        {
-            return
-        }
+
+        guard data != nil, let renderer = renderer else { return }
         
         let optionalContext = NSUIGraphicsGetCurrentContext()
         guard let context = optionalContext else { return }
@@ -223,17 +220,17 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         // make sure the data cannot be drawn outside the content-rect
         context.saveGState()
         context.clip(to: _viewPortHandler.contentRect)
-        renderer?.drawData(context: context)
+        renderer.drawData(context: context)
         
         // if highlighting is enabled
         if (valuesToHighlight())
         {
-            renderer?.drawHighlighted(context: context, indices: _indicesToHighlight)
+            renderer.drawHighlighted(context: context, indices: _indicesToHighlight)
         }
         
         context.restoreGState()
         
-        renderer?.drawExtras(context: context)
+        renderer.drawExtras(context: context)
         
         if _xAxis.isEnabled && !_xAxis.isDrawLimitLinesBehindDataEnabled
         {
@@ -259,13 +256,13 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             context.saveGState()
             context.clip(to: _viewPortHandler.contentRect)
             
-            renderer?.drawValues(context: context)
+            renderer.drawValues(context: context)
             
             context.restoreGState()
         }
         else
         {
-            renderer?.drawValues(context: context)
+            renderer.drawValues(context: context)
         }
 
         _legendRenderer.renderLegend(context: context)
@@ -529,19 +526,19 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         
         if recognizer.state == NSUIGestureRecognizerState.ended
         {
-            if !self.isHighLightPerTapEnabled { return }
+            if !isHighLightPerTapEnabled { return }
             
             let h = getHighlightByTouchPoint(recognizer.location(in: self))
             
-            if h === nil || h?.isEqual(self.lastHighlighted) ?? false
+            if h?.isEqual(lastHighlighted) ?? true
             {
-                self.highlightValue(nil, callDelegate: true)
-                self.lastHighlighted = nil
+                highlightValue(nil, callDelegate: true)
+                lastHighlighted = nil
             }
             else
             {
-                self.highlightValue(h, callDelegate: true)
-                self.lastHighlighted = h
+                highlightValue(h, callDelegate: true)
+                lastHighlighted = h
             }
         }
     }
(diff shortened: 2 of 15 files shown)


liuxuan30:
Mind to give some details and what this for?

russellbstephens:
I was just looking through the repo and noticed some areas that could be refactored to remove `!`.  Overall it adds a more swifty feel to your codebase

petester42:
Thanks for the submission. Moving to a more swift style is great. I've wanted to do that myself but hadn't had the time. I'd like to merge this change after swift 4 support is merge. Should be soon.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -927,22 +927,17 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
...
             
-            if foundScrollView !== nil && !foundScrollView!.nsuiIsScrollEnabled
+            if !(foundScrollView?.nsuiIsScrollEnabled ?? true)
             {
                 foundScrollView = nil
             }
             
             var scrollViewPanGestureRecognizer: NSUIGestureRecognizer!
             
-            if foundScrollView !== nil
```
jjatie: replace the loop with
```
        let scrollViewPanGestureRecognizer = foundScrollView?.nsuiGestureRecognizers?.first {
            $0 is NSUIPanGestureRecognizer
        }

```

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -562,12 +561,12 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         {
             if h == nil
             {
-                delegate!.chartValueNothingSelected?(self)
+                delegate?.chartValueNothingSelected?(self)
```
jjatie: Why leave `delegate != nil` in the if statement if you're going to use optional chaining?
Maybe 
```if callDelegate, let delegate = delegate { ... } ``` instead?

File: Source/Charts/Charts/ChartViewBase.swift [hunk 3]
```diff
@@ -562,12 +561,12 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         {
             if h == nil
```
jjatie: To follow the spirit of your PR, you should flip the order of the if-else, i.e.
```
if let h = h { ... }
else { ... }
```

jjatie:
@russellbstephens I'd encourage you to look at PR #2737. It takes this a lot further and reduces many algorithms to one-liners. Perhaps you could bring some of those changes into here, as currently I can't get that PR to build.

russellbstephens:
@jjatie sure no problem, I put this up to see if you all would be interested in cleaning up the code a bit.  Now that I know there is interest, I would be happy to go further with this

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: I'm not confident in using `.first` here to achieve the same effect

Source/Charts/Charts/ChartViewBase.swift
russellbstephens: 👍

Source/Charts/Charts/ChartViewBase.swift
russellbstephens: 👍

russellbstephens:
@jjatie after looking at #2737 I feel you should be able to resolve the conflicts and get it into running order, as these two refactors are separate with different goals

russellbstephens:
@petester42 is this ready to go in?

liuxuan30:
sorry to jump in. Is on swift 4 or need another convert on your side?

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -888,7 +887,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         
         do
         {
-            try imageData.write(to: URL(fileURLWithPath: path), options: .atomic)
+            try imageData?.write(to: URL(fileURLWithPath: path), options: .atomic)
```
liuxuan30: Optional chaining can cover up the exception if the image type is not supported?

Source/Charts/Charts/CombinedChartView.swift
```diff
@@ -120,7 +120,7 @@ open class CombinedChartView: BarLineChartViewBase, CombinedChartDataProvider
             {
                 return nil
             }
-            return (_data as! CombinedChartData!).lineData
```
liuxuan30: It has been sometime we review this part. But seems `CombinedChartData!` can be `CombinedChartData` now,

Source/Charts/Charts/CombinedChartView.swift
```diff
@@ -120,7 +120,7 @@ open class CombinedChartView: BarLineChartViewBase, CombinedChartDataProvider
...
+            return (_data as? CombinedChartData)?.lineData
```
liuxuan30: shall we use `(_data as! CombinedChartData)` or `(_data as? CombinedChartData)?` ? seems `as!` is safe here though as? is safer

Source/Charts/Charts/CombinedChartView.swift
```diff
@@ -55,7 +55,7 @@ open class CombinedChartView: BarLineChartViewBase, CombinedChartDataProvider
             
             self.highlighter = CombinedHighlighter(chart: self, barDataProvider: self)
             
-            (renderer as! CombinedChartRenderer?)!.createRenderers()
+            (renderer as? CombinedChartRenderer)?.createRenderers()
```
liuxuan30: old code is a bit weird (maybe the compiler bug). But if it's safe unwrap here, why not use `(renderer as! CombinedChartRenderer).createRenderers()`? `renderer` has been setup as combined chart renderer in `initialize()` above.

Source/Charts/Charts/PieChartView.swift
```diff
@@ -104,16 +104,16 @@ open class PieChartView: PieRadarChartViewBase
         let optionalContext = NSUIGraphicsGetCurrentContext()
         guard let context = optionalContext else { return }
         
-        renderer!.drawData(context: context)
+        renderer?.drawData(context: context)
```
liuxuan30: why `?` here and below?

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -555,11 +555,10 @@ open class PieRadarChartViewBase: ChartViewBase
             
             if !rotationWithTwoFingers
             {
-                let touch = touches.first as NSUITouch!
-                
-                let touchLocation = touch?.location(in: self)
-                
-                processRotationGestureBegan(location: touchLocation!)
+                if let touch = touches.first {
```
liuxuan30: I'm not sure why use if here. Any ideas?

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -102,17 +102,17 @@ open class BarChartDataEntry: ChartDataEntry
     
     open func sumBelow(stackIndex :Int) -> Double
     {
-        if _yVals == nil
+        guard let _yVals = _yVals else
```
liuxuan30: shall we keep using `_` for local instant here? seems a little bit weird, though we don't have to change other code.

russellbstephens: updated

Source/Charts/Charts/PieRadarChartViewBase.swift
russellbstephens: updated

Source/Charts/Charts/PieChartView.swift
russellbstephens: updated

russellbstephens: its best to avoid `!` in swift

Source/Charts/Charts/CombinedChartView.swift
russellbstephens: I want to avoid using `!`, since in a case where `_data === nil` `!` will crash

Source/Charts/Charts/CombinedChartView.swift
russellbstephens: updated

Source/Charts/Charts/CombinedChartView.swift
russellbstephens: I'd rather not rely on the assumption `renderer !== nil`

Source/Charts/Charts/ChartViewBase.swift
russellbstephens: updated

russellbstephens:
is there any hope of this going in?

Source/Charts/Animation/Animator.swift
```diff
@@ -115,9 +103,9 @@ open class Animator: NSObject
                 elapsed = duration
             }
            
-            if _easingX != nil
+            if let _easingX = _easingX
```
jjatie: Why isn't this one line?
`phaseX = _easingX?(elapsed, duration) ?? Double(elapsed / duration)

Source/Charts/Animation/Animator.swift
```diff
@@ -135,9 +123,9 @@ open class Animator: NSObject
                 elapsed = duration
             }
             
-            if _easingY != nil
+            if let _easingY = _easingY
             {
-                phaseY = _easingY!(elapsed, duration)
```
jjatie: This can be one line as well.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -233,7 +233,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         
         context.restoreGState()
         
-        renderer!.drawExtras(context: context)
```
jjatie: Why not use
`guard let data = data, let renderer = renderer else { return }`
at the beginning of the function?

Logically, a `draw()` function would require a renderer to draw anything, and looking through the implementation of the function this is true.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -533,7 +533,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             
             let h = getHighlightByTouchPoint(recognizer.location(in: self))
             
-            if h === nil || h!.isEqual(self.lastHighlighted)
+            if h === nil || h?.isEqual(self.lastHighlighted) ?? false
```
jjatie: this can be just `if h?.isEqual(self.lastHighlighted) ?? true {`

jjatie: The former force unwrapping was also acceptable, because if the first condition passed on the or statement, we would have already entered the if statement.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -753,7 +753,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 
                 if ((h === nil && lastHighlighted !== nil) ||
                     (h !== nil && lastHighlighted === nil) ||
-                    (h !== nil && lastHighlighted !== nil && !h!.isEqual(lastHighlighted)))
+                    (h !== nil && lastHighlighted !== nil && !(h?.isEqual(lastHighlighted) ?? false)))
```
jjatie: Here we've already checked that `h != nil` in the AND statement. If `h == nil` escape early from the AND statement, so it is safe to force unwrap `h`

jjatie: Currently there is an issue in the last statement. If the whole expression was evaluated, then in the last component (`!(h?.isEqual(lastHighlighted) ?? false)`) you would return true when `h == nil` which is not what we want.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -931,7 +931,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             ))
         {
             var scrollView = self.superview
-            while (scrollView !== nil && !scrollView!.isKind(of: NSUIScrollView.self))
+            while (scrollView !== nil && !(scrollView?.isKind(of: NSUIScrollView.self) ?? false))
```
jjatie: Same issue here where if the last component is evaluated and `scrollView == nil` we return `true`
Easiest way to solve this is to replace
`(scrollView !== nil && !(scrollView?.isKind(of: NSUIScrollView.self) ?? false))`
with
`!(scrollView?.isKind(of: NSUIScrollView.self) ?? true)`

Source/Charts/Charts/BarLineChartViewBase.swift
jjatie: Why not? The code you've replaced the for loop with is not the same. In the for loop, we escape after we find the first `NSUIPanGestureRecognizer`, in your code it will continue to loop through and find more.

jjatie: Rather than leave it the old way (using our own implementation of `first`) why not simply use the built-in method and replace 5 lines with 2.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -394,11 +394,10 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
             descriptionText.characters.count > 0
             else { return }
         
-        var position = description.position
-
-        // if no position specified, draw on default position
-        if position == nil
-        {
+        let position: CGPoint
```
jjatie: Could be cleaner to do this.
```
        let p = description.position ?? CGPoint(x: bounds.width - _viewPortHandler.offsetRight - description.xOffset,
                                                y: bounds.height - _viewPortHandler.offsetBottom - description.yOffset - description.font.lineHeight)
```

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -394,11 +394,10 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+        if let descriptionPosition = description.position {
+            position = descriptionPosition
+        } else {
             let frame = self.bounds
```
jjatie: It's generally good practice to short a variable call  (i.e. `let width = self.frame.size.width`), but in this case it is unnecessary and wrong. Unnecessary because you don't need to explicitly state `self`, and wrong because in Cocoa `frame` and `bounds` mean two different things. `frame` should have never been here in the first place.

Source/Charts/Charts/CombinedChartView.swift
jjatie: I agree with @russellbstephens . However I think a future ticket should involve making `renderer` a required property since you can't do anything with a chart without a renderer.

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -573,11 +569,10 @@ open class PieRadarChartViewBase: ChartViewBase
     {
         if rotationEnabled && !rotationWithTwoFingers
         {
-            let touch = touches.first as NSUITouch!
-            
-            let touchLocation = touch?.location(in: self)
-            
-            processRotationGestureMoved(location: touchLocation!)
+            if let touch = touches.first {
```
jjatie: Can we move the `if let touch = touches.first` into `if rotationEnabled && !rotationWithTwoFingers`
i.e.
```
if rotationEnabled && !rotationWithTwoFingers, let touch = touches.first { ... }
```

jjatie: Actually, we don't need to check to see if `touches.first` is non-nil. We know there is at least one touch because the delegate method is called. Even Apple force unwraps this their sample code.

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -595,11 +590,10 @@ open class PieRadarChartViewBase: ChartViewBase
         
         if rotationEnabled && !rotationWithTwoFingers
         {
-            let touch = touches.first as NSUITouch!
-            
-            let touchLocation = touch?.location(in: self)
-            
-            processRotationGestureEnded(location: touchLocation!)
+            if let touch = touches.first {
```
jjatie: Same as previous comment

Source/Charts/Charts/RadarChartView.swift
```diff
@@ -118,19 +118,19 @@ open class RadarChartView: PieRadarChartViewBase
         
         if drawWeb
         {
-            renderer!.drawExtras(context: context)
+            renderer?.drawExtras(context: context)
```
jjatie: `guard let data = _data, let renderer = renderer else { return }` at the beginning of the function,

Source/Charts/Components/MarkerImage.swift
```diff
@@ -102,7 +102,7 @@ open class MarkerImage: NSObject, IMarker
             height: size.height)
         
         NSUIGraphicsPushContext(context)
-        image!.draw(in: rect)
+        image?.draw(in: rect)
```
jjatie: Seems like there should be a `guard let image = image else { return }` at the beginning of the method. Drawing will do nothing if `image == nil`

Source/Charts/Components/MarkerView.swift
```diff
@@ -36,18 +36,18 @@ open class MarkerView: NSUIView, IMarker
         {
             offset.x = -point.x
         }
-        else if chart != nil && point.x + width + offset.x > chart!.bounds.size.width
+        else if let chart = chart, point.x + width + offset.x > chart.bounds.size.width
```
jjatie: `guard let chart = chartView else { return }` at the beginning.

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -102,17 +102,17 @@ open class BarChartDataEntry: ChartDataEntry
     
     @objc open func sumBelow(stackIndex :Int) -> Double
     {
-        if _yVals == nil
+        guard let yVals = _yVals else
```
jjatie: Throughout the project we're being inconsistent in how we write closures. Sometimes we open the closure on a new line, sometimes not, and sometimes we write one liners. Since you're going through a significant portion of the project already, can we consolidate on a format?

Source/Charts/Data/Implementations/Standard/BubbleChartData.swift
```diff
@@ -27,8 +27,7 @@ open class BubbleChartData: BarLineScatterCandleBubbleChartData
     /// Sets the width of the circle that surrounds the bubble when highlighted for all DataSet objects this data object contains
     @objc open func setHighlightCircleWidth(_ width: CGFloat)
     {
-        for set in (_dataSets as? [IBubbleChartDataSet])!
-        {
+        (_dataSets as? [IBubbleChartDataSet])?.forEach { set in
```
jjatie: Could be a one liner as 
`(_dataSets as? [IBubbleChartDataSet])?.forEach { $0.highlightCircleWidth = width }`

Source/Charts/Animation/Animator.swift
russellbstephens: done

Source/Charts/Animation/Animator.swift
russellbstephens: done

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/BarLineChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/ChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/ChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/PieRadarChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/PieRadarChartViewBase.swift
russellbstephens: done

Source/Charts/Charts/RadarChartView.swift
russellbstephens: done

Source/Charts/Components/MarkerImage.swift
russellbstephens: done

Source/Charts/Components/MarkerView.swift
russellbstephens: this func must return a `CGPoint`

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
russellbstephens: I would love too, personally I think opening on the same line is fine

Source/Charts/Data/Implementations/Standard/BubbleChartData.swift
russellbstephens: done

russellbstephens:
@jjatie how's it looking?

Source/Charts/Components/MarkerView.swift
jjatie: `guard let chart = chartView else { return offset }`

jjatie:
Looks fine to me!

russellbstephens:
nice!

Source/Charts/Components/MarkerView.swift
russellbstephens: done

russellbstephens:
so merge?

russellbstephens:
:)

jjatie:
@liuxuan30

liuxuan30:
seems I don't have further questions. @danielgindi @petester42, any comments?

petester42:
The review that has been done by everyone seems adequate to me.

liuxuan30:
ok, merging.

PR #2711, opened 2017-08-13
Added ability to fill highlighted circle.

matijag1:
Added ability to fill highlighted circle with highlighted color and also to make it larger when it is highlighted.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/BubbleChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/BubbleChartDataSet.swift
@@ -40,6 +40,16 @@ open class BubbleChartDataSet: BarLineScatterCandleBubbleChartDataSet, IBubbleCh
     
     // MARK: - Styling functions and accessors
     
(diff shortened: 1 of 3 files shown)


liuxuan30:
Can't
```swift
    /// `true` if drawing circles for this DataSet is enabled, `false` ifnot
    var drawCircleHoleEnabled: Bool { get set }
```
solve the problem? It's in `fileprivate func drawCircles(context: CGContext)`

matijag1:
Function 'drawCircles(context: CGContext)' is part of LineChartRenderer, and this is BubbleChartRenderer, so it can't help.

liuxuan30:
Then we need some context about this PR.
Could you please provide some and maybe screenshots to demo? 

We need to evaluate if this is useful for most of the people or just your personal demand. If former, we may merge but if latter, you could subclass bubble chart and keep it in your class.

jjatie:
Feel free to reopen @matijag1

PR #2717, opened 2017-08-14
Conform to macOS api changes in swift 3.2

ohbargain:
Compiling the framework with Xcode 9 beta 4 results in error due to macOS Swift api changes.
This pull request resolves that.


liuxuan30:
We will hold all Xcode 9 PRs after it's released.

petester42:
in xcode 9 beta 6 there is no warnings but there this is more swifty.

PR #2730, opened 2017-08-18
Adopted `==` in favour of `isEqual(to:)`

jjatie:
Moved `public func ==(_:_:) -> Bool` into type extensions as static functions to properly adopt the `Equatable` protocol.
All instances of `isEqual(to:)` have been replaced with `==`, but the `isEqual(to:)` methods remain for compatibility with objective-c

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
@@ -69,15 +69,7 @@ open class ChartDataEntry: ChartDataEntryBase
         
         self.x = x
     }
-    
-    // MARK: NSObject
 
-    open override func isEqual(_ object: Any?) -> Bool
-    {
-        guard let object = object as? ChartDataEntry else { return false }
-        return self == object
-    }
-    
     // MARK: NSObject
     
     open override var description: String
@@ -101,17 +93,16 @@ open class ChartDataEntry: ChartDataEntryBase
 
 // MARK: Equatable
 extension ChartDataEntry/*: Equatable*/ {
-    public static func ==(lhs: ChartDataEntry, rhs: ChartDataEntry) -> Bool
-    {
-        if lhs === rhs
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let object = object as? ChartDataEntry else { return false }
+
+        if self === object
         {
             return true
         }
 
-        return lhs.isKind(of: type(of: rhs))
-            && ((lhs.data == nil && rhs.data == nil)
-                || (lhs.data?.isEqual(rhs.data) ?? false))
-            && lhs.y == rhs.y
-            && lhs.x == rhs.x
+        return ((data == nil && object.data == nil) || (data?.isEqual(object.data) ?? false))
+            && y == object.y
+            && x == object.x
     }
 }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
@@ -71,20 +71,11 @@ open class ChartDataEntry: ChartDataEntryBase
     }
     
     // MARK: NSObject
-    
+
     open override func isEqual(_ object: Any?) -> Bool
     {
-        if !super.isEqual(object)
-        {
-            return false
-        }
-        
-        if fabs((object! as AnyObject).x - x) > Double.ulpOfOne
-        {
-            return false
-        }
-        
-        return true
+        guard let object = object as? ChartDataEntry else { return false }
+        return self == object
     }
     
     // MARK: NSObject
@@ -108,32 +99,18 @@ open class ChartDataEntry: ChartDataEntryBase
     }
 }
 
-public func ==(lhs: ChartDataEntry, rhs: ChartDataEntry) -> Bool
-{
-    if lhs === rhs
+// MARK: Equatable
+extension ChartDataEntry/*: Equatable*/ {
+    public static func ==(lhs: ChartDataEntry, rhs: ChartDataEntry) -> Bool
     {
-        return true
+        if lhs === rhs
+        {
+            return true
+        }
+
+        return lhs.isKind(of: type(of: rhs))
+            && (lhs.data?.isEqual(rhs.data) ?? true)
+            && lhs.y == rhs.y
+            && lhs.x == rhs.x
     }
-    
-    if !lhs.isKind(of: type(of: rhs))
-    {
-        return false
-    }
-    
-    if lhs.data !== rhs.data && !lhs.data!.isEqual(rhs.data)
-    {
-        return false
-    }
-    
-    if fabs(lhs.x - rhs.x) > Double.ulpOfOne
-    {
-        return false
-    }
-    
-    if fabs(lhs.y - rhs.y) > Double.ulpOfOne
-    {
-        return false
-    }
-    
-    return true
 }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
@@ -109,7 +109,8 @@ extension ChartDataEntry/*: Equatable*/ {
         }
 
         return lhs.isKind(of: type(of: rhs))
-            && (lhs.data?.isEqual(rhs.data) ?? true)
+            && ((lhs.data == nil && rhs.data == nil)
+                || (lhs.data?.isEqual(rhs.data) ?? false))
             && lhs.y == rhs.y
             && lhs.x == rhs.x
     }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
@@ -96,7 +96,8 @@ extension ChartDataEntryBase/*: Equatable*/ {
             return true
         }
         return lhs.isKind(of: type(of: rhs))
-            && (lhs.data?.isEqual(rhs.data) ?? true)
+            && ((lhs.data == nil && rhs.data == nil)
+                || (lhs.data?.isEqual(rhs.data) ?? false))
             && lhs.y == rhs.y
     }
 }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
@@ -73,14 +73,6 @@ open class ChartDataEntryBase: NSObject
     
     // MARK: NSObject
     
-    open override func isEqual(_ object: Any?) -> Bool
-    {
-        guard let object = object as? ChartDataEntryBase else { return false }
-        return self == object
-    }
-    
-    // MARK: NSObject
-    
     open override var description: String
     {
         return "ChartDataEntryBase, y \(y)"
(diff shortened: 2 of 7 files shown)


Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
```diff
@@ -108,32 +99,18 @@ open class ChartDataEntry: ChartDataEntryBase
...
-    if fabs(lhs.y - rhs.y) > Double.ulpOfOne
-    {
-        return false
+        if lhs === rhs
+        {
+            return true
+        }
+
+        return lhs.isKind(of: type(of: rhs))
+            && (lhs.data?.isEqual(rhs.data) ?? true)
```
liuxuan30: why `?? true` here?

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
```diff
@@ -108,32 +99,18 @@ open class ChartDataEntry: ChartDataEntryBase
...
+            && lhs.y == rhs.y
```
liuxuan30: It seems you replaced `if fabs(lhs.x - rhs.x) > Double.ulpOfOne` with `lhs.x == rhs.x`. Want to be cautious about this change.

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
liuxuan30: I think we should be careful with this `?? true`. what if lhs.data is nil when rhs.data is nil, or rhs.data is not nil?

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
jjatie: I agree. The Swift core says it implements `==` for `FloatingPoint` according to IEEE 754-2008. Unfortunately I can't quote this due to the nature of IEEE documents, but we should be fine using the built in `==` and not have to rely on `ulpOfOne`.

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
jjatie: I will fix it now.

jjatie: On second thought, do we want to include the `data` property in the equality test at all? I can see an argument for either side, but I think equality can omit `data`. Testing identity (`===`) will compare `data` inherently.

liuxuan30: I am not sure why data is included here. But @danielgindi is missing :(
Why can't we just keep the old conditions like 
```swift
&& lhs.data !== rhs.data
&& !lhs.data!.isEqual(rhs.data)
```
ideally data will not be nil, if it's nil we will get the exception, but the new code will cover it up by returning false

Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
```diff
@@ -106,27 +87,16 @@ open class ChartDataEntryBase: NSObject
...
-    
-    if fabs(lhs.y - rhs.y) > Double.ulpOfOne
-    {
-        return false
+        if lhs === rhs
+        {
+            return true
+        }
+        return lhs.isKind(of: type(of: rhs))
+            && (lhs.data?.isEqual(rhs.data) ?? true)
```
liuxuan30: after we come to agreement, this needs change too

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
jjatie: `!==` shouldn't be used. Equality and identity are two different things. Also, my next step is to turn `ChartDataEntry` into a struct which means it's impossible to to test identity.

The new condition I just added tests equality properly.

jjatie:
The requested changes have been made...

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
liuxuan30: >!== shouldn't be used

Why?
The old code wanted to check the data identity first (`lhs.data !== rhs.data`) and later the data equality. However new code just checks `(lhs.data == nil && rhs.data == nil)`, which is different I think. If someone put same data in LineChartData and BarChartData, the new code shall return  true, while it is supposed to be false?

BTW, if you mean you want to turn `ChartDatEntry` into struct and make a PR, we are grateful and thank for your effort, but the main author has been missing for a while, I am not sure if all your work could get merged in time. So take your time and don't let this occupied most of your time and you find it's not getting attention will make you kind of frustrated.

jjatie: The final section of this [API Doc](https://developer.apple.com/documentation/swift/equatable) explains the difference between equality and identity, and why identity should not be considered when implementing `Equatable`. There is an argument to be made for the first check (`if lhs === rhs { return true }`), because it allows a quick escape without having to check each parameter individually (though I still disagree with it. I'm curious to see the performance of that identity check vs simply checking the equality of the ivars).

With data there are 5 possible scenarios
```
   lhs, rhs
1. nil, nil
2. val1, nil
3. nil, val1
4. val1, val2
5. val1, val1 
```
using `lhs.data?.isEqual(rhs.data) ?? false` will provide the correct result in cases 2, 3, 4, and 5. In order to account for case 1, we must also check to see if they are both `nil`.

jjatie: To address your other point, there are a few areas where Charts has some issues (e.g. large data sets, performance with real time data, some API oddities). It is also clearly not a Swift project. I understand the language has evolved quite a bit since it's original release, but there are some key areas (e.g. why do `ChartData` and `ChartDataSets` not conform to the appropriate `Collection` and `Equatable` protocols? Why is `ChartDataEntry` [and ideally `ChartData` and `ChartDataSets`] a class and not a struct?) that could use enhancements to make it feel more like a Swift library.

I get the impression from the readme that this is supposed to be a community project. It's fine that the lead maintainer is away to be with their family. It's been a few months now without him though, so how long do we wait until we choose someone else to lead the project? Ideally more than one person so as to not fall in the same trap of having nobody qualified to make bigger decisions again? From reading responses to issues and PRs, it seems nobody currently reviewing them really has sufficient knowledge of Swift to make a decision on more fundamental changes to the framework (which will be required to handle larger datasets).

The frustrating part isn't that someone doesn't have time to look at a given PR, it's that even if they do nothing will happen. For example: this is a pretty simple PR that doesn't change the usage or performance of the framework, and is something that is required to make the framework "more Swifty" and (I can't say required) is a helpful step in making the framework be much more performant. With the exception of missing the case when both `lhs.data` and `rhs.data` are `nil`, there is no reason for it not to be accepted.

liuxuan30: please allow me to find spare time to digest a little. You are giving too much information for me :) but that's good swift expert is helping us

liuxuan30: for your second reply, yes it's bothering us (me) that we all have work to do and not having time to track the swift evolution. It could be lazy or just too many family errands. I am currently the only active person reviewing issues and PRs. But I am not swift expert and for those swift specific improvements, I need to find a time sit down and digest first. If it's small changes, I could merge for sure. But for very big changes, especially that will remove the support to objc usage, this gonna take time.

You are welcome on board to help us. You can try gitter or send emails to @danielgindi. We need to find out a solution for how to embrace swift better.

liuxuan30: for the first point regards the code, sure we have `if lhs === rhs { return true }` already, but `ChartDatEntry`'s `data` is `AnyObject?` first. Checking identify first will avoid checking equatable if they are not same class at first, right?

petester42:
This change looks ok for me but i'd like to add this after swift 4 support is added.

liuxuan30:
@petester42 
shall we remove
```swift
if lhs.data !== rhs.data
```
here? My concern is in extreme cases different data class may be passed through, and they have the same entries... Like a BarData and LineData hold the same entries

petester42:
That wouldn't be true though right? `lhs.data` and `rhs.data` would have to be the same instance of an object.

liuxuan30:
My bad, `lhs.data` refers to `open var data: AnyObject?`, not `ChartData`. In this case, it's not safe to remove the identity check? Old code has `if lhs.data !== rhs.data` check in `ChartDataEntryBase` but I don't obviously see the new code check the class type

jjatie:
@petester42  What are your thoughts on removing `data` from the equality check altogether?

petester42:
Depends on what the intent of the equality should be. The equality to me should be that the objects are the same. If there is need to compare only part of the object like x and y values then some compare function should exists like `a.hasSamePoint(b)`.

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
```diff
@@ -108,32 +99,18 @@ open class ChartDataEntry: ChartDataEntryBase
...
-public func ==(lhs: ChartDataEntry, rhs: ChartDataEntry) -> Bool
-{
-    if lhs === rhs
-    {
-        return true
-    }
-    
-    if !lhs.isKind(of: type(of: rhs))
+// MARK: Equatable
+extension ChartDataEntry/*: Equatable*/ {
```
petester42: Why is equatable commented out?

Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
```diff
@@ -106,27 +87,16 @@ open class ChartDataEntryBase: NSObject
     }
 }
 
-public func ==(lhs: ChartDataEntryBase, rhs: ChartDataEntryBase) -> Bool
-{
-    if lhs === rhs
+// MARK: Equatable
+extension ChartDataEntryBase/*: Equatable*/ {
```
petester42: Why is equatable commented out?

Source/Charts/Highlight/Highlight.swift
```diff
@@ -176,81 +176,25 @@ open class Highlight: NSObject
...
+        guard let object = object as? Highlight else { return false }
+        return self == object
     }
 }
 
-func ==(lhs: Highlight, rhs: Highlight) -> Bool
-{
-    if lhs === rhs
+// MARK: Equatable
+extension Highlight /*: Equatable*/ {
```
petester42: Why is equatable commented out?

petester42:
@jjatie I still have issues with this. If tests could be added for this change that would be great since it would be fairly easy to add.

jjatie:
@petester42 I'm not sure what tests you want exactly. I'll write one for the equality/inequality of the couple types that are now using `==`. Are there any more you'd like?

Source/Charts/Highlight/Highlight.swift
jjatie: Because `NSObject` adopts `Equatable` and you can't declare conformance twice. When Obj-C support is dropped, then we will need to uncomment it.

Source/Charts/Data/Implementations/Standard/ChartDataEntryBase.swift
jjatie: Because `NSObject` adopts `Equatable` and you can't declare conformance twice. When Obj-C support is dropped, then we will need to uncomment it.

Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
jjatie: Because `NSObject` adopts `Equatable` and you can't declare conformance twice. When Obj-C support is dropped, then we will need to uncomment it.

jjatie:
I have reverted back to `isEqual` declarations, but using `==` to follow Swift convention because I found this:

According to [Apple](https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html#//apple_ref/doc/uid/TP40014216-CH4-ID35)
>Swift provides default implementations of the == and === operators and adopts the Equatable protocol for objects that derive from the NSObject class. The default implementation of the == operator invokes the isEqual: method, and the default implementation of the === operator checks pointer equality. You should not override the equality or identity operators for types imported from Objective-C.

liuxuan30:
can this be closed if you replaced with a new one?

PR #2737, opened 2017-08-20
Minor changes to improve clarity and reduce amount of code

jjatie:
Making use of `defer` where appropriate (e.g. `CGContext().saveGState()` / `CGContext().restoreGState()`)
Using replaced `if` where really meaning `guard`, and `if case` with `switch` where appropriate. Much less nesting as a result.
Making use of `Swift.min` and `Swift.max`
Changed some `Double` to `CGFloat` where it makes sense to. This also reduces typecasting.
Refined `for` loops, in some cases replacing them with `map`, `filter`, and/or `reduce`


jjatie:
There appears to be an inconsistent compiler bug preventing this from building. I am investigating further this week.

petester42:
Thanks for the improvements but I think we should wait until after swift 4 support is merged to do this.

jjatie:
Tests are passing fine for me locally, any thoughts on why they're failing in Travis CI?

jjatie:
I'm going to take a more structured approach to this.

PR #2767, opened 2017-09-02
Update xcode project for xcode 9

### LINKED ISSUE #2593: Invalid Bundle - Xcode 9 Beta 2
First, thank you for this great library. It works pretty nicely. The problem is not coding issue but configuration. It seems that Apple has added some new verification rules for submitted binaries. 
I have an app which uses Charts as a dependency using Carthage.
Carthage config: ```github "danielgindi/Charts" ~> 3.0.0``` resolved Charts version ```v3.0.2```

I've built and tested my app and uploaded to iTunes Connect and the upload was successful but the binary processing failed:
```
Dear developer,

We have discovered one or more issues with your recent delivery for "<app name>". To process your delivery, the following issues must be corrected:

Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

Once these issues have been corrected, you can then redeliver the corrected binary.

Regards,

The App Store team
```

Of course, it was very difficult to figure out what was the problem. I've check all Xcode settings in my app but didn't find anything suspicious. I've also posted a question in Apple forum: [Invalid Bundle - Xcode 9 Beta 2](https://forums.developer.apple.com/thread/81893?q=Xcode%209%20beta)

Luckily, I've got response from on the Apple forum that was helpful. 
I've went through all my 3rd party library dependencies and checked build settings and build scheme settings for all of them.
During this search I've found that _Charts_ scheme has ```Gather coverage data``` enabled. 
I needed to disable this settings during the build to be able to get my app's binary successfully processed by iTunes Connect.

So, I guess we need to turn this setting off otherwise people will fail their binary submission.

petester42:
This update the Charts project to support Xcode 9 and also fix #2593.


liuxuan30:
Hi @petester42, this also included https://github.com/danielgindi/Charts/pull/2507, right?

petester42:
No. Not updated for swift 4. Just making sure project structure is good, code coverage bug is fixed and ci passes.

liuxuan30:
ah ok, thanks

PR #2796, opened 2017-09-18
Draworder type update

jjatie:
`CombinedChartView` ivar `drawOrder` is now of type `[DrawOrder]`


liuxuan30:
CI failed. looks like not complete change?

jjatie:
@liuxuan30 That's a weird one. `DrawOrder` can be represented in obj-c though because it as a raw value type of `Int`... am I missing something?

jjatie:
Update. Objective-C can't have an array of enum cases. I'll remove the PR.

PR #2800, opened 2017-09-19
Custom text alignment for no data case

Browncoat:
Add NSTextAlignment option to noDataText to allow for custom text alignment


jjatie:
@liuxuan30 This seems like a good an improvement. Do you think we should accept it?

liuxuan30:
sure. @Browncoat let's fix the conflict and follow the code style

Issue #2168, opened 2017-02-15
Crash Observed when AxisRenderer interval is becoming NaN


nanduahmed-sikka:
I have a chart graph and based upon Segment Controller, We change the value of graph. When the axis graphs values are changed and segment is changed, the zoom level remains the same as it was earlier zoomed to. The following line is crashing in AxisRendererBase.swift since variable are becoming NaN. 
        let intervalSigDigit = Int(interval / intervalMagnitude)
My Chart Release shows following in Lock file
  - Charts (3.0.1):
    - Charts/Core (= 3.0.1)

nanduahmed-sikka:
This is one of the lldb log: fatal error: Double value cannot be converted to Int because it is either infinite or NaN

liuxuan30:
need more info why some of the values become NaN or inf.

nanduahmed-sikka:
Not sure. I am just using the API and it was crashing on this If in near future I get to reproduce I will surely paste the info. As of now I would like someone to fix it for invalid values. Any help would be appreciated.

liuxuan30:
yep but we need more info on how this happens

capecrawler:
Also getting few crashes from fabric that is somehow related to this.

![screen_shot_2017-04-20_at_10_58_43_am](https://cloud.githubusercontent.com/assets/1147223/25211652/c067ce10-25b8-11e7-84b9-3775813b10a7.png)

liuxuan30:
are you having same issue as in #2356?

capecrawler:
I am not sure since I was not able to replicate this crash and it points to different line of code which is `let intervalSigDigit = Int(interval / intervalMagnitude)`. My guess is that intervalMagnitude value is 0.0 or NaN that is causing the crash.

liuxuan30:
that's weird, I need to know when `let intervalMagnitude = ChartUtils.roundToNextSignificant(number: pow(10.0, Double(Int(log10(interval)))))` will make `intervalMagnitude` zero. I don't see obvious number will do so.

josman185hotm:
Hello @liuxuan30 , @danielgindi 
I'm getting an error in line 125: fatal error: Double value cannot be converted to Int because it is either infinite or NaN

this is the line: let intervalMagnitude = ChartUtils.roundToNextSignificant(number: pow(10.0, Double(Int(log10(interval)))))

please if you have any suggestions I'll be very thankful.

here you have more data:
[...]

PR #2804, opened 2017-09-20
fix for crash when interval is NaN

pingd:
After clearing a working chart and then adding new data (dataset), the chart crashes in AxisRendererBase.swift line 125 with "Double value cannot be converted to Int because it is either infinite or NaN". The reason for that is "min" and "max" are NaN  (because the Transformer._matrixValueToPx is somewhat invalid(?), why does it happen?).

The issue is referenced here: https://stackoverflow.com/questions/45496912/how-to-clear-a-chart-and-then-add-data-to-it-in-swift-3-using-chartview-clear and here: https://github.com/danielgindi/Charts/issues/2168 (josman185hotm's comment from 17 Jun).

### DIFF
--- a/Source/Charts/Renderers/AxisRendererBase.swift
+++ b/Source/Charts/Renderers/AxisRendererBase.swift
@@ -103,7 +103,7 @@ open class AxisRendererBase: Renderer
         let labelCount = axis.labelCount
         let range = abs(yMax - yMin)
         
-        if labelCount == 0 || range <= 0 || range.isInfinite
+        if labelCount == 0 || range <= 0 || range.isInfinite || range.isNaN
         {
             axis.entries = [Double]()
             axis.centeredEntries = [Double]()


liuxuan30:
#2377 is my thought to fix. Have you checked it's the reason why you saw NaN on your side? 

BTW, Do you have the exact values that causes NaN? We'd like to investigate? Adding a NaN check is the last resort to prevent crash, but we don't want to silent the crash first, it will be more difficult to notice.

liuxuan30:
I looked at your change again, it seems not the same thing. `range` should not be NaN at all.
Do you have the exact values that causes NaN? We'd like to investigate. 

Adding a NaN check is the last resort to prevent crash, but we don't want to silent the crash first, it will be more difficult to notice.

liuxuan30:
please open an issue to track NaN crash for your case, with details or values that causes NaN, thanks.

pingd:
I am not able to open an issue with actual code for about two weeks, as I am away. However, it happens roughly when you do something like in the SO issue.

In my case, the steps go like this:
1. Set up a CandleStickChartView.
2. Create a new CandleChartData with no underlaying dataset. Set chart.data to it.
3. Create an empty CandleChartDataSet. Add it to chart data.
4. Add values to chart. (At this point everything works correctly, the chart is displayed, the values can be update real time and so on).
5. When I need to switch to another ViewController, I stop my data updates and clear chart.
6. When I go back to a ViewController with chart, I repeat the process from step 2. After adding values and calling notifyDataSetChanged(), I get "Double value cannot be converted to Int because it is either infinite or NaN" on line 125 in AxisRendererBase.

I will open an actual issue with code once I am back.

liuxuan30:
I log the reason in #2845 closing this.

SeRG1k17:
@liuxuan30 
I think this will be useful to many.
And, it is required to use only a ChartData class, because any subclass also calls crash( < ios11).

```swift    
override var data: ChartData? {
        get { return super.data }
        set {
            guard let newData = newValue, newData.entryCount > 0 else {
                super.data = ChartData(dataSet: ChartDataSet(values: [ChartDataEntry()]))
                return
            }
            
            super.data = newData
            notifyDataSetChanged()
        }
    }
```

atalayasa:
Hello @SeRG1k17 I had same issue too but no other solutions worked for me where should I put your code ? Thanks

atalayasa:
I have solved the issue using following answer https://stackoverflow.com/questions/45496912/how-to-clear-a-chart-and-then-add-data-to-it-in-swift-3-using-chartview-clear

Issue #2840, opened 2017-10-03
Crash when calling `func highlightValue(x: Double, dataSetIndex: Int, callDelegate: Bool)`


everlof:
I use charts version 3.0.2.

I call `highlightValue` like this:
`highlightValue(x: 2, dataSetIndex: 0, callDelegate: true)`.

I then get a crash here:
![screen shot 2017-10-03 at 09 08 42](https://user-images.githubusercontent.com/1321434/31113631-80ccff02-a81a-11e7-8235-634251d70dcc.png)

When following the code stack trace from the crash:

Stack trace step 5:
Row 522 in ChartViewBase.swift
The `Highlight` object is created with `x=2`, `y=Nan`, `dataSetIndex=0 `.

Step trace step 4:
The `Highlight` object is then passed to `entryForHighlight` on row 547 in ChartViewBase.swift.

Then it crashes with `Index out of range` on row 278 in CombinedChartData:
`let data = dataObjects[highlight.dataIndex]`

Because `highlight.dataIndex` is `-1`, since it's not set from the constructor. The constructor that is used is: 
```
    /// - parameter x: the x-value of the highlighted value
    /// - parameter y: the y-value of the highlighted value
    /// - parameter dataSetIndex: the index of the DataSet the highlighted value belongs to
    public init(x: Double, y: Double, dataSetIndex: Int)
    {
        _x = x
        _y = y
        _dataSetIndex = dataSetIndex
    }
```
Where `highlight.dataIndex` is not initialized.

liuxuan30:
You can pass `dataIndex` to 
>@objc open func highlightValue(_ highlight: Highlight?, callDelegate: Bool)

Before that you create the highlight object you need to calculate dataIndex carefully when you manually call it.

liuxuan30:
Or we add a new method to pass `dataIndex` into.

everlof:
Not sure how this is an `enhancement`, when calling a public API (`highlightValue`) causes a crash.

However, seems like the fixes you've created would solve it.

liuxuan30:
@everlof because you use the wrong API as they don't have `dataIndex` setup. Use `@objc open func highlightValue(_ highlight: Highlight?, callDelegate: Bool)` and manually create the highlight object and setup your correct dataIndex first.

jjatie:
@liuxuan30 I agree. This sounds like a problem with documentation more than anything else, but your #2852 also works ; )

PR #2852, opened 2017-10-08
For #2840. add dataIndex parameter in `highlightValue()` calls

liuxuan30:
for #2840. add dataIndex parameter in `highlightValue()` calls. Default is -1


jjatie:
@liuxuan30 I agree with you in #2840, but if you think we should add this, let's merge it.

liuxuan30:
what you think of this PR itself?

jjatie:
I think we can reduce them all to one function:
```
@objc open func highlightValue(x: Double, y: Double = .nan, dataSetIndex: Int, dataIndex: Int = -1, callDelegate: Bool = true)
```

liuxuan30:
I will take a look later tomorrow

liuxuan30:
actually there is only one method left to use the full method footprint, and chaining up maybe better now. I would stick to current one.

PR #2887, opened 2017-10-16
Update LICENSE

sDaniel:
Copyright 2016 Daniel Cohen Gindi & Philipp Jahoda < replaced template with content

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2016 Daniel Cohen Gindi & Philipp Jahoda
```
liuxuan30: @danielgindi is this intended to use a template or forgot to change?
BTW, should it be 2017 or just keep the first year?

petester42: It was missed. We can update the year later

PR #2889, opened 2017-10-16
animated update for barchart

kevinjtchen:


kevinjtchen:
no longer needed

kditraglia:
Why is this no longer needed, is there another work-around that can get the effect you added here?

cowgp:
Why is this not needed?  am I missing some other PR that has this functionality that did get merged?   I would really like to see this exact feature as build in this PR

Jauzee:
@kevinjtchen can you please answer the questions above. I was looking for this functionality and came here.

PR #2894, opened 2017-10-17
Give the users customizable axis label limits (Fixes #2085)

### LINKED ISSUE #2085: How to set LabelCount bigger than 25 in HorizatalBarView

igzrobertoestrada:
This change adds the ability to users to override the imposed limitation of at least 2 labels per axis and at most 25 labels per axis if they need to and at their own risk. 

By default these customizable limits are hardcoded to the previous limits, at least 2 labels, at most 25.

The rationale for this change is that when coming from previous versions of Charts, users are able to use –at least on the X axis– as many labels as they wish and when they upgrade to a newer version of Charts they find that they have lost functionality, and this change brings that functionality back.

### DIFF
--- a/Source/Charts/Components/AxisBase.swift
+++ b/Source/Charts/Components/AxisBase.swift
@@ -213,6 +213,16 @@ open class AxisBase: ComponentBase
     /// the total range of values this axis covers
     @objc open var axisRange = Double(0)
     
+    /// The minumum number of labels on the axis
+    @objc open var axisMinLabels: Int = Int(2) {
+        didSet { axisMinLabels = axisMinLabels > 0 ? axisMinLabels : oldValue }
+    }
+    
+    /// The maximum number of labels on the axis
+    @objc open var axisMaxLabels: Int = Int(25) {
+        didSet { axisMinLabels = axisMaxLabels > 0 ? axisMaxLabels : oldValue }
+    }
+    
     /// the number of label entries the axis should have
     /// max = 25,
     /// min = 2,
@@ -228,13 +238,13 @@ open class AxisBase: ComponentBase
         {
             _labelCount = newValue
             
-            if _labelCount > 25
+            if _labelCount > axisMaxLabels
             {
-                _labelCount = 25
+                _labelCount = axisMaxLabels
             }
-            if _labelCount < 2
+            if _labelCount < axisMinLabels
             {
-                _labelCount = 2
+                _labelCount = axisMinLabels
             }
             
             forceLabelsEnabled = false
--- a/Source/Charts/Components/AxisBase.swift
+++ b/Source/Charts/Components/AxisBase.swift
@@ -214,12 +214,12 @@ open class AxisBase: ComponentBase
     @objc open var axisRange = Double(0)
     
     /// The minumum number of labels on the axis
-    @objc open var axisMinLabels: Int = Int(2) {
+    @objc open var axisMinLabels = Int(2) {
         didSet { axisMinLabels = axisMinLabels > 0 ? axisMinLabels : oldValue }
     }
     
     /// The maximum number of labels on the axis
-    @objc open var axisMaxLabels: Int = Int(25) {
+    @objc open var axisMaxLabels = Int(25) {
         didSet { axisMinLabels = axisMaxLabels > 0 ? axisMaxLabels : oldValue }
     }
     


liuxuan30:
I am not sure if allowing other values could raise any issue when >25 ?

igzrobertoestrada:
I'm using 61 labels in the X axis in an AppStore production app and it's working just fine. 

In my case, this value comes as a client requirement for a feature. Many other people may be in the same or similar situation just as they tell in #2085 and they and/or their clients cannot afford a step back in functionality compared to what they could have in prior versions.

Many people will be safe with the 25-labels limitation. It will work wonders for them and they won't even notice it, but others may need more flexibility when it comes to customizing their charts in order to satisfy requirements that –in many cases– are non negotiable.

liuxuan30:
I wonder how can you draw 61 labels on x axis on a mobile screen? ipad or iPhone? I am OK with this change. But what I concerned is in somewhere, the is one piece of code that only thinks about <25 values, if >25, a bug triggered.
@danielgindi do you remember anything? If this is safe to merge?

alejandroluengo:
It's ok, merge needed for sure

igzrobertoestrada:
@liuxuan30 I have the barchart inside a UIScrollView so the users can swipe through the chart because having the entire chart on a relatively small region of an iPhone screen in portrait mode would be an awful user experience and really really difficult to read.

liuxuan30:
@igzrobertoestrada why your question is related to this PR?
It's your choice to show the full chart or zoom in to only show some; and there are APIs to restrict how many items you can see. e.g. visibleXrange.

igzrobertoestrada:
@liuxuan30 Because I need to maintain a piece of code that requires that charts is able to render more than 25 labels as it did in previous versions.

Source/Charts/Components/AxisBase.swift
```diff
@@ -213,6 +213,16 @@ open class AxisBase: ComponentBase
     /// the total range of values this axis covers
     @objc open var axisRange = Double(0)
     
+    /// The minumum number of labels on the axis
+    @objc open var axisMinLabels: Int = Int(2) {
```
jjatie: Type inference is sufficient. `@objc open var axisMinLabels = 2`

jjatie:
@igzrobertoestrada Isn't having the chart in a `UIScrollView` with many labels, the same things as just scrolling through a chart with fewer labels? The only difference is that in the latter case the you can always see the y-axis.

Source/Charts/Components/AxisBase.swift
igzrobertoestrada: Fixed in new commit. Removed ```: Int``` but maintained ```Int(2)``` for consistence with the rest of the codebase.

Source/Charts/Components/AxisBase.swift
```diff
@@ -214,12 +214,12 @@ open class AxisBase: ComponentBase
     @objc open var axisRange = Double(0)
     
     /// The minumum number of labels on the axis
-    @objc open var axisMinLabels: Int = Int(2) {
+    @objc open var axisMinLabels = Int(2) {
         didSet { axisMinLabels = axisMinLabels > 0 ? axisMinLabels : oldValue }
```
liuxuan30: should we use `didSet` or setter for this check here? @jjatie

liuxuan30:
Again I don't have issue to let user decide the value instead o capping it at 25. What concerns me is I don't know if there is any function that take 25 as granted, when we change to larger ones, it broke. Only @danielgindi might remember.
If we think this can be merged and later let user report what's breaking, it's good to go

Source/Charts/Components/AxisBase.swift
```diff
@@ -213,6 +213,16 @@ open class AxisBase: ComponentBase
...
+        didSet { axisMinLabels = axisMinLabels > 0 ? axisMinLabels : oldValue }
+    }
+    
+    /// The maximum number of labels on the axis
+    @objc open var axisMaxLabels = Int(25) {
+        didSet { axisMinLabels = axisMaxLabels > 0 ? axisMaxLabels : oldValue }
+    }
+    
     /// the number of label entries the axis should have
     /// max = 25,
```
liuxuan30: comment might consider to change as well.

jjatie:
@danielgindi What are your thoughts on adding this to the library? Technically there's no reason to impose a limit on the number of labels, but from a data visualization perspective it makes sense to limit it.

jjatie:
@liuxuan30 I think if there are functions that use require 25 already, we should consider those bugs that need to be fixed. If we accept this change, we will just have to fix those issues as they come up.

liuxuan30:
I just dont know if there is any. Just my guess. So far I dont see

liuxuan30:
let's merge it

jjatie:
@liuxuan30 FYI I'm available for about an hour.

PR #2913, opened 2017-10-26
fixed axisValueFormatter memleak

kanjanaSi:


liuxuan30:
not that simple. That requires you strong hold the formatter somewhere else. This can cause other issues as well. We had a thread for that. But for now, just not use self as your formatter seems better, or try weakSelf or owned

PR #2914, opened 2017-10-26
remove max label count

kanjanaSi:
I have to show more than 25 labels. I add chart view as subview to scrollview so I'm sure the labels will not be overlapping.


liuxuan30:
keep it in your branch. showing more labels seems not possible on a mobile screen. There is a PR to set min/max value for it. so this can be closed

kanjanaSi:
It's possible to show it in scroll view.

Issue #2933, opened 2017-10-31
Xcode 9.1 - Characters is deprecated


saldous:
7 new Xcode warnings:

> 'characters' is deprecated: Please use String or Substring directly

thierryH91200:
see #2866

saldous:
@thierryH91200 that issues was closed but not resolved... Xcode 9.1 is now GA and these warnings now show. Please re-open #2866

liuxuan30:
closing this

Issue #2866, opened 2017-10-10
'characters' is deprecated: Please use String or Substring directly


michael-snapptec:
This appears to show up 7 times in the current version.

One instance is specifically at `AxisBase.swift:137`.

I am not a Swift developer (Obj-C for the win!), but it appears that this can be corrected by changing this:

    if longest.characters.count < text.characters.count

to this:

    if longest.count < text.count

liuxuan30:
interesting, what's your swift version? As my swift 4 and xcode9 does not warn this at all.

saldous:
Please re-open, Xcode 9.1 is now released and shows these warnings.

liuxuan30:
not updated yet; but you are welcome to file a PR for this update

PR #2942, opened 2017-11-02
modify for Character Alert: characters is deprecated

suzuhiroruri:
character is deprecated.
As for "count character" code, delete "character" code.
You will delete deprecated alert.


liuxuan30:
https://github.com/danielgindi/Charts/issues/2933 #2866

Issue #753, opened 2016-02-12
Swift Package Manager


danielgindi:
It seems like there's a standard being emerged, as SPM... https://swift.org/package-manager/#example-usage

We've been waiting a long time for such a standard, something that will be the equivalent of NPM for JS. 
We need to keep track of this - and if/when possible, declare Charts as a Package.

doofyus:
Could this please be merged in.

PR #2950, opened 2017-11-02
Add Swift Package Manager support.

BrianDoig:
This implements https://github.com/danielgindi/Charts/issues/753 by adding a Package.swift file.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,15 @@
+// swift-tools-version:4.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+import PackageDescription
+
+let package = Package(
+    name: "Charts",
+    products: [
+        .library(name: "Charts", type: .dynamic, targets: ["Charts"])
+    ],
+    dependencies: [],
+    targets: [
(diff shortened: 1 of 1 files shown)


liuxuan30:
Just curious, swift pkg mgr does not need the library version? How do we update? (new to swift pkg, but will spend some time on it later)

liuxuan30:
seems fine. @petester42 or swift experts?

doofyus:
Could this please be merged in.

liuxuan30:
I will merge.

danielgindi:
Cool :-)

petester42:
What about the dependency on UIKit/AppKit? Is there a way to identify that this will only work on Mac and if you have the sdk installed?

petester42:
@doofyus could you elaborate what you needed this for? I was asked a question about spm support and from what I know about spm it can’t be used with apps right now.

doofyus:
It's true that SPM is a bit limited currently, but it can be used to keep the _dependencies_ up to date and with some workarounds, it's possible to use them in an app as well (i use SPM in a couple of them).

PR #2961, opened 2017-11-06
added highlightColor parameter for pie charts

pascalherrmann:
Added attribute `highlightColor` to specify a color for highlighted pie chart sectors.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/PieChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/PieChartDataSet.swift
@@ -105,6 +105,9 @@ open class PieChartDataSet: ChartDataSet, IPieChartDataSet
     /// the color for the slice-text labels
     open var entryLabelColor: NSUIColor? = nil
     
+    /// the color for the highlighted sector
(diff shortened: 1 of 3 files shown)


liuxuan30:
wow, I don't realize pie chart does not support highlightColor.
@danielgindi for feature or enhancements like this, shall we go ahead or still need sync with Android?

petester42:
Looks good to me. If this needs to be ported to android it's fairly simple. I think we should take ownership of the project's direction since @danielgindi isn't here a lot. The issues and pull requests are piling up and we can't always wait for @danielgindi to review. We should also add more contributors to help with the load.

liuxuan30:
ok sure. but how about the "more swift-y" changes? those are kind of heavy; and seems we need a swift expert to make some feedback. are you able to do that? I'm not having enough time to keep up with the swift trend. I'm afraid I am more learner than reviewer for those changes.

petester42:
Ya sure. Things being more “swifty” just really means the code should be easy to read. Read more like a sentence then instructions.

PR #2967, opened 2017-11-08
Add support for iPhone X

liuxuan30:
adopted safe area for demo chart controllers;
remove launch image, use launch screen storyboard instead

Note:
I don't apply all constraints to safe area, only for those that will violate the guidelines. For example, slider Y has a decent distance to its super view, so changing to safe area is not a must.


liuxuan30:
fine to merge. I played a little with iPhone X  
(the simulator)

PR #2981, opened 2017-11-13
Animator non nil

jjatie:
`Animator` is now non-optional in `Renderer` types


liuxuan30:
why non nil?

BTW, I suggest merge small commits into one with more meaningful message, like put this into single one commit with a message 'Animator non nil'. I saw you have multiple 'Merging with upstream/remote', kind of unneccessary. 

I can use 'Squash and merge' for this PR anyway, so next time maybe.

jjatie:
I can start flattening my commits.

The framework is not designed to allow for `animator == nil` as made clear by the fact I was able to simply remove the `guard-else { return }` for every area `animator` is used. This should be reflected in the property declaration and the initializers.

Source/Charts/Renderers/ChartDataRendererBase.swift
```diff
@@ -15,13 +15,13 @@ import CoreGraphics
 @objc(ChartDataRendererBase)
 open class DataRenderer: Renderer
 {
-    @objc open var animator: Animator?
+    @objc open let animator: Animator
```
liuxuan30: did Xcode tell you to use `let` instead of `var`? I wonder if any chance people would modify renderer's animator later.

liuxuan30:
oops, looks like there are dependency or orders. conflict when 3010 merged.

jjatie:
Fixed

Source/Charts/Renderers/ChartDataRendererBase.swift
jjatie: I made it `let`. What is the use case of changing the renderer?

liuxuan30: Not sure, just seeing it used to be `var`, though it was optional and some advanced user maybe? Using let will be more strict.
If we use `var`, will Xocde complain it is never re-assigned?

jjatie: It's an instance variable, so it's impossible for Xcode to know if it is never re-assigned.

It is more strict, but from what I can tell there is no consideration for having multiple animators. We have one class for it, and no protocols to define an interface. Overriding any functions in `Animator`, without diving into the framework to know exactly what needs to be there and what can be changed, will cause problems.

liuxuan30:
@petester42 do you want to take a look at this and https://github.com/danielgindi/Charts/pull/2982? I checked it is legit to allow animator and renderer non nil, as the callers passes an object already.

liuxuan30:
I will take this then. seems many of these PRs there I could handle :)

PR #2980, opened 2017-11-13
View port handler nonnil

jjatie:

PR #2982, opened 2017-11-13
`ChartRenderer`'s must be initialized with a chart

jjatie:


liuxuan30:
why is that?

jjatie:
Does it make sense that a ChartRenderer has no chart to render?

My guess is this initializer parameter was made optional because the `weak` property declaration forced the property to be declared as optional.

liuxuan30:
`@objc open weak var dataProvider: BarChartDataProvider?` is declared as weak, sounds like a historical reason or personal habit to use optional for params.

This change is in conjunction with https://github.com/danielgindi/Charts/pull/2981 to make animator non nil as well.

jjatie:
Oops, I missed that one. Fixing now.

jjatie:
Any PR that has a dependency on another as a comment indicating that. For the most part there are no dependencies.

liuxuan30:
oops, conflicts. when you solve I will check again.

jjatie:
@liuxuan30 Fixed, but pull #2980 first. While in some cases I might be able to consolidate some PRs, I think for the most part they should be separate.

liuxuan30:
this can be next to merge?

jjatie:
Yes, fixing now.

jjatie:
Fixed

liuxuan30:
need to use squash to merge

PR #2986, opened 2017-11-13
 add draw path for LineChartRenderer for personal processing

Orange-W:
it's more convenient  than using data set to draw.


Orange-W:
Supplement background path for drawing. Fixed the bug drawPath will be nil.

liuxuan30:
if it's personal, don't make a PR.

PR #2990, opened 2017-11-13
Axis renderer protocol

jjatie:
Slight more code, but allows for better type safety and removes a lot of unnecessary type conversion. It's a "Swift-ier" way.

Note that this builds off the changes made in #2991

PR #2991, opened 2017-11-13
Renderer protocol

jjatie:
`Renderer` and `DataRenderer` are now protocols instead of classes.


jjatie:
There are changes here that pave the way to much larger changes to the framework which allow better type safety and less reliance on optionals. This is immediately followed by #2990. Once these two are approved, there will be more changes that can be quickly and easily made.

jjatie:
This includes #2980

jjatie:
@liuxuan30 After #2982 I think this one should be done next (and #2990)

liuxuan30:
oops just read your comment.
I would suggest you create an issue to track your PRs, like which are for 4.0 branch and which can be merged into master directly. I feel dizzy switching around the PRs not sure which to start with. Some "minor" changes have hundreds of diffs, which are not minor to review.

I would first start with diffs that are just two digits and later with the big ones.

jjatie:
I'll just change the destination branch on the PR

jjatie:
In the future I will create Issues and name the PRs after them.

jjatie:
Continued in #3136

PR #2992, opened 2017-11-13
Minor changes to logic in `ViewPortJob` subclasses.

jjatie:


Source/Charts/Jobs/ViewPortJob.swift
```diff
@@ -17,11 +17,11 @@ import CoreGraphics
 open class ViewPortJob: NSObject
 {
     @objc internal var point: CGPoint = CGPoint()
-    @objc internal weak var viewPortHandler: ViewPortHandler?
+    @objc internal weak var viewPortHandler: ViewPortHandler? // TODO: Can this be unowned?
```
liuxuan30: before merge, do we need to finish TODO? I am willing to check if we can use `unowned` when I got time, but seems weak is totally fine here. What's your concern using weak?

liuxuan30:
Generally this PR looks great to use `@discardableResult` to remove those "let _" stuff and removed unused init().

Source/Charts/Jobs/ViewPortJob.swift
jjatie: `weak` does work here, but I think what is actually intended is `unowned`. In terms of function, nothing would change, but it does clean up a lot of `guard` statements if we can use `unowned` I feel like that is a separate PR and one that requires a bit more digging. I'm not sure what the policy is on TODOs

liuxuan30: I don't know the policy neither. but if it is another PR, could you remove the comment?  I will merge later.

jjatie: Done

PR #2993, opened 2017-11-13
Replaced `ChartUtils.Math` in favour of an extension on `FloatingPoint`

jjatie:
Increases readability, and in many cases removes a set of parentheses to ensure order of operations. Also can be applied to any `FloatingPoint` type, not just `Double` and `CGFloat`


jjatie:
@liuxuan30 @petester42 This is a quick and easy one.

Source/Charts/Utils/ChartUtils.swift
```diff
@@ -16,9 +16,26 @@ import CoreGraphics
...
+
+    var rad2deg: Self {
+        return self * 180 / .pi
+    }
+
+    /// - returns: An angle between 0.0 < 360.0 (not less than zero, less than 360)
+    /// NOTE: Value must be in degrees
+    var normalizedAngle: Self {
+        let angle = truncatingRemainder(dividingBy: 360)
+        return (sign == .minus) ? angle + 360 : angle
```
liuxuan30: the order is flipped. I need to refresh my mind first to see if they are equivalent. Or you already did that?

jjatie: They’re equivalent.

liuxuan30: made some stupid equations :P

Source/Charts/Utils/ChartUtils.swift
```diff
@@ -16,9 +16,26 @@ import CoreGraphics
     import UIKit
 #endif
 
+extension FloatingPoint {
+    var deg2rad: Self {
```
liuxuan30: Generally it's fine. But do you think we should follow the Camel-Case naming? like deg2Rad

jjatie: Up to you. I tried `deg2Rad` and `deg2RAD` and they both look silly to me. I also considered `inRadians` and `inDegrees`, but that leaves out grads

liuxuan30: OK. How about we keep old `FDEG2RAD` to keep consitent?

jjatie: I'll drop the F since it's implied as it's an extension on `FloatingPoint`

liuxuan30:
This is the one of the change I like most, indeed make it more swift and simple

PR #2994, opened 2017-11-13
Replaced relevant `ChartUtils` methods with `Double` extensions

jjatie:
Improves readability.
`nextUp` is built in and provides the same functionality.


jjatie:
@liuxuan30 @petester42  This is a quick and easy one.

Source/Charts/Utils/ChartUtils.swift
```diff
@@ -41,50 +75,7 @@ open class ChartUtils
...
-        
-        if i.isInfinite || i.isNaN
-        {
-            return 0
-        }
-        
-        return Int(ceil(-log10(i))) + 2
-    }
-    
-    internal class func nextUp(_ number: Double) -> Double
```
liuxuan30: swift doc:
>For any finite value x, x.nextUp is greater than x. For nan or infinity, x.nextUp is x itself. The following special cases also apply:
>
>If x is -infinity, then x.nextUp is -greatestFiniteMagnitude.
If x is -leastNonzeroMagnitude, then x.nextUp is -0.0.
If x is zero, then x.nextUp is leastNonzeroMagnitude.
If x is greatestFiniteMagnitude, then x.nextUp is infinity.

I'm kind of confused `If x is -infinity, then x.nextUp is -greatestFiniteMagnitude.` because `For nan or infinity, x.nextUp is x itself`

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -343,7 +343,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         if _defaultValueFormatter is DefaultValueFormatter
         {
             // setup the formatter with a new number of digits
-            let digits = ChartUtils.decimals(reference)
+            let digits = reference.decimalPlaces
```
liuxuan30: I'm no native speaker. is `decimalPlaces` ok? @petester42 @danielgindi

Source/Charts/Utils/ChartUtils.swift
jjatie: The only place this is used is in the axis renderers. I removed this implementation because the default one is safer to pass into the call sites.

liuxuan30: yep but it's different than the library implementation, towards the infinity check and result. I need to check several places.

liuxuan30: but the real issue is swift doc has paradox:
>For nan or infinity, x.nextUp is x itself. The following special cases also apply

does infinity here contains both +inf and -inf? or just +inf?

because it added at last line:
>If x is -infinity, then x.nextUp is -greatestFiniteMagnitude

jjatie: `.infinitiy != -.infinity`
It makes sense that the nextUp from negative infinity is the negative greatest finite magnitude. The opposite is true for `nextDown()`

liuxuan30:
Except for the naming, I have no more questions. If you think it's a proper name, let's squash and merge.

PR #2996, opened 2017-11-13
Removed methods and properties deprecated in 1.0

jjatie:
These produce quite a bit of clutter.
It has been 2-3 years since their deprecation, it seems reasonable to remove them now.

If this is not to be accepted now, when is a good time to resubmit this PR?


liuxuan30:
have you checked what's the date when adding those deprecated tag? 

BTW, it has conflicts :)

jjatie:
I looked for a bit, but gave up looking for the merge. They are marked as deprecated in 1.0, so it doesn't make sense to me that they would be introduced in 3.0.

petester42:
It was a misunderstanding on what the number there meant. It only applies to OS versions. It doesn’t read our frameworks version. They could have been deprecated in 2.0 or 3.0 I don’t know when though.

liuxuan30:
but we can remove them now, right?

jjatie:
Searching through the commits I've found that these methods were all deprecated between Apr 8, 2016 and Sept 21, 2016.

Source/Charts/Utils/ChartUtils.swift
```diff
@@ -287,60 +287,4 @@ open class ChartUtils
     {
         return _defaultValueFormatter
     }
-
-    /// MARK: - Bridging functions
-    
-    internal class func bridgedObjCGetNSUIColorArray (swift array: [NSUIColor?]) -> [NSObject]
```
liuxuan30: why this is included?

liuxuan30:
I'm supporting remove this soon, like after we finish refactoring and some other enhancements and make a major release, since many things changed and seems it's the right time to remove those APIs

Source/Charts/Utils/ChartUtils.swift
jjatie: These only existed to support the deprecated functions. Since I am removing the deprecated functions in this PR, these methods are no longer needed.

jjatie: Also, this could have been written as `let myObjArray = myColorArray as! [NSObject]`

liuxuan30:
@jjatie will your future commits involve these deprecated methods? If so, I favor to merge soon. If not, maybe put it to last?

jjatie:
My other PRs don't touch these deprecated methods at all. If we are not merging soon, then this should be in 4.0.0. Given the deprecations are between 14 and 20 months old, I think it's acceptable to do this before 4.0.0

liuxuan30:
OK, I will hold this for the moment, we shall have a release after the refinement, so just a matter of time and order.

liuxuan30:
could you switch to 4.0 branch? so we can merge this.

liuxuan30:
merging this, to make things easier.

PR #2997, opened 2017-11-14
Remove java interface convention

jjatie:
Updated naming to be in line with Swift API Guidelines


jjatie:
As per the [Swift API](https://swift.org/documentation/api-design-guidelines/) guide
> Protocols that describe what something is should read as nouns (e.g. Collection)

The "I" prefix needs to go because that's from Java. Dropping the "I" leaves us with names identical to the classes that implement them. For protocols with associated types, the Swift API guide says:
> If an associated type is so tightly bound to its protocol constraint that the protocol name is the role, avoid collision by appending Type to the associated type name:
>
>>protocol Sequence {
>>    associatedtype IteratorType : Iterator
>>}
but there is no case for this.

Looking at the Swift standard library and Foundation, there are a few examples:
`String` and `StringProtocol`
`Iterator` and `IteratorProtocol`
`StaticArray` and `StaticArrayProtocol` 
etc.
it seems that this is normal practice.

My original PR #2795 was to remove the protocols altogether. I still believe they should be removed as they do not provide any additional functionality over the classes that already exist to implement these methods and properties.

In summary: I don't think the protocols should remain in the framework, but if they are going to stay they need to be renamed to match Swift convention.

petester42:
The protocols are needed for different implementations of the datasets. An example of this is ChartsRealm. It provides an extension point for anyone to add their own way to get chart data.

petester42:
For the naming it’s fine then but this is going to be a breaking change so we need to hold off on it for a bit. I might create a vNext branch for these kinds of changes until we a ready to bump the version to 4.

jjatie:
Can you make that branch? I think I'm going to end up making approved, but breaking changes over the next few weeks.

jjatie:
I have a few questions about certain implementations in the framework, that aren't documented here or in MPAndroidCharts. Where is the best spot to ask these questions?

petester42:
Open a new issue with the questions would be the best.

petester42:
i've created a branch called `4.0.0` for breaking changes

jjatie:
Will do

jjatie:
It appears I don't have sufficient permissions to add milestones to PRs

liuxuan30:
@danielgindi could you add rights for @jjatie ? I cannot give him neither

PR #3000, opened 2017-11-14
The backing var is not necessary.

jjatie:
Appears to have been a way to expose the var to obj-c like so many others in the framework.
Removed duplicated calls to `notifyDataSetChanged`


Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -177,17 +158,17 @@ open class ChartDataSet: ChartBaseDataSet
...
-    open override var entryCount: Int { return _values?.count ?? 0 }
+    open override var entryCount: Int { return values.count }
     
     /// - returns: The entry object found at the given index (not x-value!)
     /// - throws: out of bounds
     /// if `i` is out of bounds, it may throw an out-of-bounds exception
     open override func entryForIndex(_ i: Int) -> ChartDataEntry?
     {
-        guard i >= 0 && i < _values.count else {
+        guard values.indices.contains(i) else {
```
liuxuan30: seems this is O(n) cost, while the old one is O(1)?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -228,27 +209,27 @@ open class ChartDataSet: ChartBaseDataSet
         var entries = [ChartDataEntry]()
         
         var low = 0
-        var high = _values.count - 1
+        var high = values.endIndex - 1
```
liuxuan30: just keep `.count` so it's consistent with others

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -26,24 +26,24 @@ open class ChartDataSet: ChartBaseDataSet
...
+        values = [ChartDataEntry]()
+
         super.init(label: label)
-        
-        _values = [ChartDataEntry]()
     }
     
     @objc public init(values: [ChartDataEntry]?, label: String?)
     {
+        self.values = values ?? []
```
liuxuan30: if `[]` and `[ChartDataEntry]()` are both valid, shall we just use one of them?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -461,21 +432,16 @@ open class ChartDataSet: ChartBaseDataSet
...
+            if values[i] === entry
             {
-                _values.remove(at: i)
+                values.remove(at: i)
                 removed = true
                 break
             }
         }
-        
-        if removed
```
liuxuan30: I don't see this can be deleted?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -484,40 +450,24 @@ open class ChartDataSet: ChartBaseDataSet
     /// - returns: `true` if successful, `false` ifnot.
     open override func removeFirst() -> Bool
     {
-        let entry: ChartDataEntry? = _values.isEmpty ? nil : _values.removeFirst()
-        
-        let removed = entry != nil
-        
-        if removed
```
liuxuan30: same as line 474, why delete here?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -484,40 +450,24 @@ open class ChartDataSet: ChartBaseDataSet
...
     /// Removes the last Entry (at index size-1) of this DataSet from the entries array.
     ///
     /// - returns: `true` if successful, `false` ifnot.
     open override func removeLast() -> Bool
     {
-        let entry: ChartDataEntry? = _values.isEmpty ? nil : _values.removeLast()
-        
-        let removed = entry != nil
-        
-        if removed
```
liuxuan30: same as line 474

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: Because `values` has a property observer that takes care of this now.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: same as line 474

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: same as line 474

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: OK

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: Fixed

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: `count` and `endIndex` aren't necessarily the same, and `endIndex` is safer (when that's really want we want). When we are able to properly add `Collection` conformance to `ChartDataSet` then it will become possible to call this on a subrange of a `ChartDataSet` where `startIndex` isn't guaranteed to be `0` and `endIndex` isn't guaranteed to be `count`.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: Just curious, do you know whose performance is better? like `[]` may need to check the type and call the initializer?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -227,28 +208,28 @@ open class ChartDataSet: ChartBaseDataSet
     {
         var entries = [ChartDataEntry]()
         
-        var low = 0
-        var high = _values.count - 1
+        var low = values.startIndex
+        var high = values.endIndex - 1
```
liuxuan30: >When we are able to properly add Collection conformance to ChartDataSet then it will become possible to call this on a subrange of a ChartDataSet where startIndex isn't guaranteed to be 0 and endIndex isn't guaranteed to be count.

Got it. But if a sub set of ChartDataSet is called upon, I think the startIndex and endIndex is still from 0 to count, since we are on a sub set?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -291,15 +272,15 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var low = 0
-        var high = _values.count - 1
+        var high = values.count - 1
```
liuxuan30: do you mean we shall change all `.count` into `.endIndex` in the future after collection adoption? Or just some.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: Sorry, did this observer already exist? Or new added? I don't see it.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: Depends on how the collection is implemented. If slicing is implemented, then no. `startIndex` and `endIndex` are safer and they work the same as `0` and `count` for standard collection implementations, but still keep it safe for more sophisticated implementations.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: We should use `count` when we actually want to know how many of something there is, and `endIndex` when we mean the end of the collection. I'll double check, but assuming this is like the other method implementation, this should be `endIndex`.

I'll review the PR again to make sure I'm consistent.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: The performance is identical in this case. This is a compile time check, not a runtime check. It just looks cleaner this way.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: It already existed.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: Sure, I don't mean you need to change others. Just to clarify what we should use. I checked the method names and they don't reflect clearly count or boundary. Sometimes we don't clearly which to use. I will wait.

jjatie: Are we good to go on this one then?

liuxuan30:
oops, just made some experiments on the "project" tab.
@jjatie are you able to edit or add new items in the project I created "More Swift"? It may be a better place to manage these PRs

jjatie:
I am not able to add/edit anything to the project

liuxuan30:
eh.. Do you like to join in? If so we can ping Daniel to set it up

jjatie:
I'd be happy to

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: the last commit seems just reflect the change about count/endIndex.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: just one about the observer. I can't see it?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: It was on `values`, not `_values`. It calls `notifyDataSetChanged()` which calls `calcMinMax()` and only `calcMinMax`

liuxuan30: why I feel the logic is different then. The old method will trigger `calcMinMax()` after something removed from _values. However in the new one, it does not call `calcMinMax()` if remove happens.
`values.remove(at: i)` will not call `notifyDataSetChanged()` I assume, it's not a setter call.

jjatie: It is a setter call. Try running this in a playground, you will see that the print statement is called.
```
var x = [1, 2, 3, 4] {
    didSet {
        print("ITEM REMOVED")
    }
}
x.remove(at: 1)
```

liuxuan30: what... that changed my understanding

liuxuan30: Is this behavior documented? I walked through https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Properties.html, not mentioning this specifically.

jjatie: In directly, yes. 
>If you pass a property that has observers to a function as an in-out parameter, the willSet and didSet observers are always called. This is because of the copy-in copy-out memory model for in-out parameters: The value is always written back to the property at the end of the function

Modification of array are all in-out (otherwise every time you modified an array, new memory would need to be allocated).

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -68,23 +75,7 @@ open class ChartDataSet: ChartBaseDataSet
...
-        set
-        {
-            _values = newValue
-            notifyDataSetChanged()
-        }
-    }
-    
+
     /// Use this method to tell the data set that the underlying data has changed
     open override func notifyDataSetChanged()
```
liuxuan30: It seems we can safely remove `open override func notifyDataSetChanged()`? Since its superclass `ChartBaseDataSet`->`notifyDataSetChanged()` already called `notifyDataSetChanged()`

jjatie: You are correct!

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: Jesus, thanks. All good

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: what you think? It's ok either keep or remove.

jjatie: I'm removing it now

jjatie: Done

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: @liuxuan30 For future reference, I've looked into this, and all `Range` types have O(1) complexity. They perform a bounds check internally instead of iterating through the range. So this:
```
(values.startIndex..<values.endIndex).contains(i)
```
Performs the same, and, I feel, is more clear.

PR #3003, opened 2017-11-14
Minor cleanup to Highlighter types

jjatie:


Source/Charts/Highlight/ChartHighlighter.swift
```diff
@@ -48,19 +48,15 @@ open class ChartHighlighter : NSObject, IHighlighter
...
-        guard let chart = chart
-            else { return nil }
+        guard let chart = chart else { return nil }
         
         let closestValues = getHighlights(xValue: xVal, x: x, y: y)
-        if closestValues.isEmpty
-        {
-            return nil
-        }
+        guard !closestValues.isEmpty else { return nil }
```
petester42: This guard is harder to read than how it was before. I have to think about it for a second when I try to understand this line. It was more straightforward before. It doesn't reduce nesting so we can switch it back.

Source/Charts/Highlight/PieHighlighter.swift
```diff
@@ -17,12 +17,10 @@ open class PieHighlighter: PieRadarHighlighter
 {
     open override func closestHighlight(index: Int, x: CGFloat, y: CGFloat) -> Highlight?
     {
-        guard let set = chart?.data?.dataSets[0]
+        guard let set = chart?.data?.dataSets[0],
```
petester42: Combining these guard statements makes it harder to read. I think it was easier to understand before

Source/Charts/Highlight/PieRadarHighlighter.swift
```diff
@@ -23,32 +23,30 @@ open class PieRadarHighlighter: ChartHighlighter
         let touchDistanceToCenter = chart.distanceToCenter(x: x, y: y)
         
         // check if a slice was touched
-        if touchDistanceToCenter > chart.radius
-        {
+        guard touchDistanceToCenter <= chart.radius else {
```
petester42: guard or if here doesn't really make a difference.

Source/Charts/Highlight/ChartHighlighter.swift
jjatie: While `guard` often does reduce nesting, it's purpose is to indicate that
> You use a guard statement to require that a condition must be true in order for the code after the guard statement to be executed. [Apple](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ControlFlow.html).

It seems reasonable to me to expect reading "only continue if `closestValues` is not empty"

Source/Charts/Highlight/PieHighlighter.swift
jjatie: What about all the other places in the framework where we are writing `guard` statements like this? Although usually it's written like 
```
guard
    let set = chart?.data?.dataSets[0],
    let entry = set.entryForIndex(index)
    else { return nil }
```

Source/Charts/Highlight/PieRadarHighlighter.swift
jjatie: Again:
> You use a guard statement to require that a condition must be true in order for the code after the guard statement to be executed.

So in this case the guard makes it extra clear that the method can return early here

jjatie:
The convention is as the [Swift Book](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ControlFlow.html) states:
> You use a guard statement to require that a condition must be true in order for the code after the guard statement to be executed.

While technically it doesn't matter if we use an `if` or a `guard`, the `guard` does make it clear that we need a condition to be `true` to continue.

As for the longer `guard` conditions I don't know of any convention, but I like the way it's written in many places in Charts and I will update this one to look the same. (e.g.
```
        guard
            let data = _data as? BarChartData,
            let set = data.getDataSetForEntry(e) as? IBarChartDataSet
            else { return .null }
```

jjatie:
@petester42 Do you have any further feedback or can this be merged?

jjatie:
Review has been discussed in #3034 and approved.

liuxuan30:
next time merging some old PRs, just merge master and trigger a new CI to make sure breaks nothing :)

jjatie:
I did : )

PR #3005, opened 2017-11-14
Minor changes to Animator

jjatie:


Source/Charts/Animation/Animator.swift
```diff
@@ -134,15 +114,8 @@ open class Animator: NSObject
...
-            if _easingY != nil
-            {
-                phaseY = _easingY!(elapsed, duration)
-            }
-            else
-            {
-                phaseY = Double(elapsed / duration)
-            }
+
+            phaseY = _easingY?(elapsed, duration) ?? elapsed / duration
```
petester42: I think using `if let` here instead of a one-liner here would be more readable

jjatie: I disagree on this one. `??` is a very common operator in Swift (it's all over the stdlib). Would you prefer something like
```
phaseY = _easingY?(elapsed, duration)
    ?? elapsed / duration
```
or 
```
phaseY = _easingY?(elapsed, duration) ??
    elapsed / duration
```

jjatie: It's also worth noting that on #2698 there were many cases of this being approved, including this line (thought it was written as `phaseY = _easingY?(elapsed, duration) ?? Double(elapsed / duration)`

petester42: My problem is not with the ?? My problem is with the optional closure. I think It would be clearer to unwrap it first with `if let` before calling it. It is minor though. It’s up to you.

jjatie: I would agree with you in a case like:
```
let x = myFuncWithNonEscaping {
     // Logic
} ?? myDefaultValue
```
But in this case I really think this is okay.

petester42: I find it cluttered but it doesn’t really matter.

PR #3008, opened 2017-11-14
BarChartRenderer Logic cleanup

jjatie:
Mostly using guard where appropriate
Few very minor performance improvements


Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -190,17 +177,11 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
         
         for i in 0 ..< barData.dataSetCount
         {
-            guard let set = barData.getDataSetByIndex(i) else { continue }
-            
-            if set.isVisible
-            {
-                if !(set is IBarChartDataSet)
-                {
-                    fatalError("Datasets for BarChartRenderer must conform to IBarChartDataset")
```
petester42: this messaging is important and should not be removed.

jjatie: Oops, got ahead of myself. I'll fix it in the other ones too.

jjatie: @petester42 This is corrected now. Can it be merged?

liuxuan30:
master or 4.0 branch?

jjatie:
@liuxuan30 master

jjatie:
Fixed according to review. Merging since it has been sitting for a month and a half.

Charts.xcodeproj/project.pbxproj
```diff
@@ -1,1253 +1,1253 @@
 // !$*UTF8*$!
 {
-	archiveVersion = 1;
```
liuxuan30: the weird indent issue happens here as well. I do think we need to check. Something must be wrong.
The original file on my side is totally fine but in your commit the old file seems wrong, and you fixed with one that just the same as my old one..

liuxuan30: @jjatie I could reset this and just force push the valid changes. Would like me to do that? but you have to check your tab/space otherwise it will still happen

jjatie: I have a feeling that someone else's PR change them all to tabs. I haven't changed any indentation settings in years, and I've never seen an xcodeproj file with tabs. Also, why would this be an issue now when it never has before?

jjatie: I'm okay with the force push.

MohanNB:
i'm getting this in xcode 14.0.1 Cannot find type 'BarChartView' in scope please any one can give the solution.

Issue #2999, opened 2017-11-14
Collection Conformance


jjatie:
`DataSets` are collections of `DataEntries`, and `ChartData` are collections of `DataSets`. Currently we have non-standard methods to access members of the collections (e.g. `entryForIndex`).

Due to the requirement of Objective-C compatibility, we are unable to do the proper thing and adopt `Collection` (and a few related protocols) in these classes. While we are able to implement specific functions to make them feel more "Swift-y", users will lose on out expected features that come for free with `Collection` conformance (such as `map`, `reduce`, `filter`).

Are there any thoughts with how to proceed in making collections feel more Swift like?

petester42:
Maybe extensions could do the trick here so that the normal implementation works as it has forever and if you use swift you get access to the extension methods that can't be exported to objective-c. Don't know how that could be implemented though.

jjatie:
I would be happy to start an implementation for this approach. Would you like me to do this within the same file or make a superset one(s)?

petester42:
Take a break. You’ve been working too much on this lately. I’ll take a look at it.

jjatie:
#3018

PR #3018, opened 2017-11-18
[#2999]

jjatie:
ChartDataSet adopts conformances:
MutableCollection
RandomAccessCollection
RangeReplaceableCollection

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -27,7 +27,6 @@ open class ChartDataSet: ChartBaseDataSet
     public required init()
     {
         super.init()
-        
         _values = [ChartDataEntry]()
     }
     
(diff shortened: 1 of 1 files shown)


jjatie:
I entirely agree, however protocol extensions are a Swift only feature and therefor cannot be implemented on an @objc protocol. Either we get rid of objective-c, get rid of the DataSetProtocols or we accept that this is the best we can do right now.

petester42:
If we can't do it right then I need a good reason to do it at all. Are we gonna rewrite the internal logic to use `map`, `filter` and such or it is just for the consumers of the library. I don't see much value for the consumers since they probably already do those operations to their data before creating their dataset. What use cases can you think of for needing this?

jjatie:
That's reasonable.

Being able to take advantage of the collection conformance in the framework itself will greatly reduce the amount of code there is to maintain, unfortunately we are unable to do that in this implementation.

For consumers of the framework, the benefit they get from this is not `map`, `filter`, `reduce`, but to be able to access data in a more standard way if they need to, using subscripting, `index(where:)`, etc. That's the only justification I have for this implementation. Any future improvements to how this implemented will not change the current functionality. Because of this I think it would make sense to 
allow this change and also add it to ChartsRealm, then in the future (likely when we drop Obj-C support) we can provide a better implementation that we can take advantage of internally.

I see your argument though, and am fine if you choose to just close the PR.

liuxuan30:
I feel dropping ObjC support is not gonna happen in near future as long as apple still use it.

petester42:
Ya I’m leaning towards closing this PR. The only gain would be subscript access and we can just add that ourselves.

PR #3023, opened 2017-11-20
Chartdata collection conformance

jjatie:
Implemented `ChartData` `Collection` conformances.


jjatie:
Important to note that this doesn't suffer from the same issues as #3018

jjatie:
Any idea why the codecov is so wrong?

liuxuan30:
no idea how codecov works.
I wonder if collection conformance goes for 4.0?

jjatie:
For chart data it's purely additive. I don't see why it can't be done before 4.0.

Source/Charts/Data/Implementations/Standard/BarChartData.swift
```diff
@@ -14,7 +14,7 @@ import CoreGraphics
 
 open class BarChartData: BarLineScatterCandleBubbleChartData
 {
-    public override init()
+    public required init()
```
liuxuan30: why many become required init() now?

jjatie: It's a requirement of `RangeReplaceableCollection`

liuxuan30: hmm.. My Dash API Doc does not show this to me

jjatie: From [Apple](https://developer.apple.com/documentation/swift/rangereplaceablecollection)

Source/Charts/Data/Implementations/Standard/CombinedChartData.swift
```diff
@@ -28,6 +28,11 @@ open class CombinedChartData: BarLineScatterCandleBubbleChartData
     {
         super.init(dataSets: dataSets)
     }
+
+    public required init(arrayLiteral elements: ChartDataSetProtocol...)
```
liuxuan30: have you checked if any tricky bug or trap (like index to a wrong position) regards CombinedChartData collection conformance? CombinedChartData is a wrapper putting other data together like barData, lineData.

Source/Charts/Data/Implementations/Standard/BarChartData.swift
liuxuan30: image
Found a Dash bug :)

Source/Charts/Data/Implementations/Standard/CombinedChartData.swift
jjatie: I made sure that all the new Swift methods/properties have the same implementations as the existing ones, so this should be an issue.

liuxuan30: sure. It will take more time, need to catch up the dataset protocol first.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,138 @@ open class ChartData: NSObject
...
+    {
+        get{ return _dataSets[position] }
+        set{ self._dataSets[position] = newValue }
+    }
+}
+
+// MARK: RandomAccessCollection
+extension ChartData: RandomAccessCollection
+{
+    public func index(before: Index) -> Index
```
liuxuan30: I saw `RandomAccessCollection` has more required methods to provide, but here only one `index(before: Index)` which just comes from `BidirectionalCollection`. What knowledge I missed here?

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,138 @@ open class ChartData: NSObject
...
+        return element
+    }
+
+    public func removeLast(_ n: Int)
+    {
+        self._dataSets.removeLast(n)
+        notifyDataChanged()
+    }
+
+//    public func removeSubrange(_ bounds: Range<Index>) {
```
liuxuan30: don't need to call notifyDatChanged here()? how about in `replaceSubrange`. Shall we consider add some unit tests for these type of changes.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -24,13 +24,20 @@ open class ChartData: NSObject
...
-        
-        _dataSets = [ChartDataSetProtocol]()
     }
-    
+
+    public required init(arrayLiteral elements: ChartDataSetProtocol...)
+    {
+        super.init()
+
+        _dataSets = dataSets
```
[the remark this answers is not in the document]
jjatie: For the `_`, this should be changed, but that isn't in the scope of this PR.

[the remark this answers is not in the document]
jjatie: The first part is an error I will fix.

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: From the `RandomAccessCollection` doc:
> The RandomAccessCollection protocol adds further constraints on the associated Indices and SubSequence types, but otherwise imposes no additional requirements over the BidirectionalCollection protocol. However, in order to meet the complexity guarantees of a random-access collection, either the index for your custom type must conform to the Strideable protocol or you must implement the index(_:offsetBy:) and distance(from:to:) methods with O(1) efficiency

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: We do because we are accessing the backing var so it's never called. In a future PR I will remove the backing var and refactor this. Currently this works the exact same way as the other add/remove methods, but are simply Swift Protocol versions (as they should be).

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -807,47 +807,47 @@ extension ChartData: RangeReplaceableCollection
...
-        self._dataSets.removeLast(n)
+        _dataSets.removeLast(n)
         notifyDataChanged()
     }
 
-//    public func removeSubrange(_ bounds: Range<Index>) {
-//        self.dataSets.removeSubrange(bounds)
-//        notifyDataChanged()
-//    }
+    public func removeSubrange<R>(_ bounds: R) where R : RangeExpression, ChartData.Index == R.Bound {
```
liuxuan30: What R stands for?

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,138 @@ open class ChartData: NSObject
...
+        return element
+    }
+
+    public func removeLast(_ n: Int)
+    {
+        _dataSets.removeLast(n)
+        notifyDataChanged()
+    }
+
+    public func removeSubrange<R>(_ bounds: R) where R : RangeExpression, ChartData.Index == R.Bound {
```
liuxuan30: code style

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,138 @@ open class ChartData: NSObject
...
+        _dataSets.removeSubrange(bounds)
+        notifyDataChanged()
+    }
+}
+
+// MARK: Swift Accessors
+extension ChartData
+{
+    //TODO: Reevaluate if warning is still true
```
liuxuan30: I'm close to finish review soon, do we handle TODO here or in future PRs?

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: It's a generic function.

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: I meant to delete it.

Source/Charts/Data/Implementations/Standard/ChartData.swift
liuxuan30: So we use R instead of T when we have where constrains, or it does not matter?

jjatie: You can pick whichever you want. `T` is often used, but I think Apple chose `R` here because it's a `R`ange

jjatie: You can choose anything that is also a valid type (Class, Struct Protocol or Enum) name

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,132 @@ open class ChartData: NSObject
...
+    public func append(_ newElement: Element)
+    {
+        _dataSets.append(newElement)
+        calcMinMax(dataSet: newElement)
+    }
+
+    public func remove(at position: Index) -> Element
+    {
+        let element = _dataSets.remove(at: position)
+        calcMinMax()
```
liuxuan30: shall we use `notifyDataChanged` instead of `calcMinMax` here? in CombinedChartData, they works differently then their super class.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -757,3 +764,132 @@ open class ChartData: NSObject
...
+extension ChartData: RandomAccessCollection
+{
+    public func index(before: Index) -> Index
+    {
+        return _dataSets.index(before: before)
+    }
+}
+
+// MARK: RangeReplaceableCollection
+extension ChartData: RangeReplaceableCollection
```
liuxuan30: Generally it should works fine with the collection protocol adoption. The devil may exist in CombinedChartData in the future, as it works differently than usual chart data. 

We actually don't append/remove data sets into a combined data, we just assign new sub chartData. e.g.
```swift
    open override func removeDataSetByIndex(_ index: Int) -> Bool
    {
        print("removeDataSet(index) not supported for CombinedData", terminator: "\n")
        return false
    }
```
I think we should override the protocols in CombinedChartData to not allow such operations?

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: I thought about that. I think we should keep the way it is now for this PR. The existing `open func removeDataSet(_ dataSet: ChartDataSetProtocol!) -> Bool` uses `calcMinMax()`

jjatie: Maybe we can rethink that in another PR though.

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: I will do that right now : )

jjatie: Fixed

liuxuan30:
@jjatie I'm thinking of disabling all operations(add/remove/update) on dataSet in ChartData for `CombinedChartData`, while let ChartData conforming the collection protocol. Is there any easy way we can do that.

jjatie:
Why? All the other ones are valid. Unless you're going to remove the existing methods too.

jjatie:
@liuxuan30 done

liuxuan30:
Thanks, let's wait CI finish up

PR #3024, opened 2017-11-20
Chartdata collection refactor

jjatie:
Refactored logic where possible to take advantage of `ChartData`'s `Collection` conformance


jjatie:
This is in conjunction with #3023

ChartsDemo/Swift/Demos/BubbleChartViewController.swift
```diff
@@ -104,7 +104,7 @@ class BubbleChartViewController: DemoBaseViewController {
         set3.setColor(ChartColorTemplates.colorful()[2], alpha: 0.5)
         set3.drawValuesEnabled = true
         
-        let data = BubbleChartData(dataSets: [set1, set2, set3])
+        let data = [set1, set2, set3] as BubbleChartData
```
liuxuan30: may I know which protocol method enables this convenient conversion?

ChartsDemo/Swift/DemoBaseViewController.swift
```diff
@@ -117,13 +117,13 @@ class DemoBaseViewController: UIViewController, ChartViewDelegate {
     func handleOption(_ option: Option, forChartView chartView: ChartViewBase) {
         switch option {
         case .toggleValues:
-            for set in chartView.data!.dataSets {
+            for set in chartView.data! {
```
liuxuan30: I had a feeling that `for set in data` is less natural than `for set in data.dataSets`, what you think? Maybe need time to adapt?
So in the future, we will deprecate `data.dataSet`?

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
```diff
@@ -144,31 +144,31 @@ class LineChart1ViewController: DemoBaseViewController {
     override func optionTapped(_ option: Option) {
         switch option {
         case .toggleFilled:
-            for set in chartView.data!.dataSets as! [LineChartDataSet] {
+            for set in chartView.data as! LineChartData {
```
liuxuan30: want to know which method enables this as well.
And, it seems weird that a `set` object is a `ChartData` type? what you think

Source/Charts/Highlight/ChartHighlighter.swift
```diff
@@ -73,17 +73,11 @@ open class ChartHighlighter : NSObject, Highlighter
...
         
-        for i in 0 ..< data.dataSetCount
-        {
-            guard
-                let dataSet = data.getDataSetByIndex(i),
-                dataSet.isHighlightEnabled      // don't include datasets that cannot be highlighted
-                else { continue }
-            
-
+        for (i, set) in zip(data.indices, data) where set.isHighlightEnabled
```
liuxuan30: I like this advanced use. Does this come with the collection adoption?

jjatie: `for d in data` comes from `Sequence` adoption which `Collection` inherits. This also allows the zip(:) function to be used.

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
jjatie: See comment below

ChartsDemo/Swift/Demos/BubbleChartViewController.swift
jjatie: `ExpressibleByArrayLiteral`

ChartsDemo/Swift/DemoBaseViewController.swift
jjatie: I think it's just new to you. We need to continue to support `data.dataSet` as long as we keep Obj-C compatibility.

liuxuan30:
I ran twice, still failure.

jjatie:
Any chance you can add an approval or additional review before you can go to bed? Travis CI is finally getting better and I could merge during my day. BTW my time zone is GMT-0400 hours

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
liuxuan30: @jjatie `for set in chartView.data as! LineChartData` here I'm confused, as `set` implies a dataset class, while it casts to a chartdata class. is the name good here?

liuxuan30:
and when I wake up and still pending LOL.
The only question is about some names. see comments above that you didn't answered. Otherwise, it seems good.

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
liuxuan30: oh wait. Seems I asked a wrong question?
it's `chartView.data` being casted to `LineChartData`?

liuxuan30:
Travis macOS builds have trouble again. Expect delay

PR #3032, opened 2017-11-22
BarLineChartViewBase cleanup

jjatie:
For the most part, condensing logic and using `guard` where appropriate
Removed optionality of many internal variables as they were only optional to allow for deferred initialization. This is now replaced with lazy vars.
Removed empty initializer overrides.
`fileprivate` is now `private`
`@objc` tags removed from internal properties as they will never be exposed to objc users.


liuxuan30:
what's this for? master or 4.0

jjatie:
@liuxuan30 master.

PR #3034, opened 2017-11-23
ChartViewBase cleanup

jjatie:
For the most part, condensing logic and using `guard` where appropriate
Removed optionality of many internal variables as they were only optional to allow for deferred initialization. This is now replaced with lazy vars.
Removed empty initializer overrides.
`fileprivate` is now `private`


Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -67,67 +67,55 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-    @objc open var chartDescription: Description?
+    @objc open lazy var chartDescription = Description()
     
     /// This property is deprecated - Use `chartDescription.text` instead.
     @objc @available(*, deprecated: 1.0, message: "Use `chartDescription.text` instead.")
     open var descriptionText: String
     {
-        get { return chartDescription?.text ?? "" }
-        set { chartDescription?.text = newValue }
+        get { return chartDescription.text ?? "" }
```
liuxuan30: it seems we have a future conflict that these will be removed, but we change it now?
I'm not sure if we should merge https://github.com/danielgindi/Charts/pull/2996 into master now or for 4.0. I don't think it's worthy to refactor what will be removed

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -500,8 +463,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     /// - parameter callDelegate: Should the delegate be called for this change
     @objc open func highlightValue(x: Double, y: Double, dataSetIndex: Int, callDelegate: Bool)
     {
-        guard let data = _data else
-        {
+        guard let data = _data else {
```
liuxuan30: code style is make `{` a separate line

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -443,19 +410,15 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     /// null or an empty array to undo all highlighting. 
     /// This should be used to programmatically highlight values.
     /// This method *will not* call the delegate.
-    @objc open func highlightValues(_ highs: [Highlight]?)
+    @objc open func highlightValues(_ highs: [Highlight])
```
liuxuan30: I remember highlight objects are allowed to be nil

Source/Charts/Charts/ChartViewBase.swift
jjatie: It’s not a big deal though, and I will likely be the one handling merge conflicts. In this case, these deprecations need to be changed in order for the project to build.

I think we should merge into master, and I’ll handle any merge conflicts when we merge into 4.0 in the future.

Source/Charts/Charts/ChartViewBase.swift
jjatie: Reverted

Source/Charts/Charts/ChartViewBase.swift
jjatie: Fixed

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: I have merged #2996 to make things easier

jjatie: Conflicts are resolved

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -475,41 +450,27 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-            _indicesToHighlight.removeAll(keepingCapacity: false)
-        }
-        else
-        {
-            // set the indices to highlight
-            entry = _data?.entryForHighlight(h!)
-            if entry == nil
-            {
-                h = nil
+        var high = highlight
```
liuxuan30: I'm feeling this rewrite makes harder to understand the logic (the guard else{} part). If we cannot find a better one, I suggest add simple comment here.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -519,7 +480,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     /// CandleStick-Chart.
     @objc open func getHighlightByTouchPoint(_ pt: CGPoint) -> Highlight?
     {
-        if _data === nil
+        guard _data != nil else
```
liuxuan30: Not feeling guard is better here.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -538,33 +499,25 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+                let set = data?.getDataSetByIndex(highlight.dataSetIndex),
                 let e = _data?.entryForHighlight(highlight)
                 else { continue }
             
             let entryIndex = set.entryIndex(entry: e)
-            if entryIndex > Int(Double(set.entryCount) * _animator.phaseX)
-            {
-                continue
-            }
+            guard entryIndex <= Int(Double(set.entryCount) * _animator.phaseX) else { continue }
```
liuxuan30: is guard in swift as fast as if, in terms of assembling code? I'm feeling we are using lots of guard, when it just means if condition. If guard is slower than if, I would rather not to use too many.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -905,17 +857,11 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-            var val = newValue
-            if val < 0.0
-            {
-                val = 0.0
-            }
-            if val >= 1.0
-            {
-                val = 0.999
+            switch newValue {
+            case ..<0.0: _dragDecelerationFrictionCoef = 0
```
liuxuan30: where can I find this magic definition.. I just can find things like `case 1..<5:` in the manual.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -156,55 +160,37 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
     }
     
     internal func initialize()
     {
         #if os(iOS)
-            self.backgroundColor = NSUIColor.clear
+            self.backgroundColor = .clear
         #endif
 
-        _animator = Animator()
```
liuxuan30: so if we uses lazy, how do we override the property in subclass to a new class object? e.g some users may write their own animator class, or XAxis class, and just override initialize() to replace them.

jjatie: They are still able to do this exactly the same as they have before.

Source/Charts/Charts/ChartViewBase.swift
jjatie: The convention is as the [Swift Book](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ControlFlow.html) states:
> You use a guard statement to require that a condition must be true in order for the code after the guard statement to be executed.

While technically it doesn't matter if we use an `if` or a `guard`, the `guard` does make it clear that we need a condition to be `true` to continue.

Source/Charts/Charts/ChartViewBase.swift
jjatie: They compile the same.

The convention is as the [Swift Book](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ControlFlow.html) states:
> You use a guard statement to require that a condition must be true in order for the code after the guard statement to be executed.

While technically it doesn't matter if we use an `if` or a `guard`, the `guard` does make it clear that we need a condition to be `true` to continue.

Source/Charts/Charts/ChartViewBase.swift
jjatie: [One Sided Ranges(https://github.com/apple/swift-evolution/blob/master/proposals/0172-one-sided-ranges.md) are new in Swift 4.

Source/Charts/Charts/ChartViewBase.swift
danielgindi: Yeah I agree with @jjatie on this one - we are used to C-style languages, while Swift does not aim at being one... For Swift it's more natural to put a `guard`

Source/Charts/Charts/ChartViewBase.swift
danielgindi: Most of what you know as "optionals", and stuff that work with optionals, and guards and stuff - are compile time. You want to be sure that something is not null and unwrap it, and unwrapping does not mean that you have more code there - it just means that the compiler trusts the non-null object and allows you to refer to it without the `?` operator

liuxuan30:
@jjatie I'm confused. We have a card for PRs merged to master, and 4.0 card merge to 4.0 branch. You moved most of them from master to 4.0 card, shall we change the branch?

jjatie:
Yep! I missed a couple I guess. Things will be easier if I just move them to 4.0 and they aren't adding/changing functionality so they can wait until 4.0

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: Then we should add some comments/notes that people should override initialize() to replace the objects with their classes?

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: When I have time, I would create a sample app and check what guard and if implementation looks like in asm code. Ideally should not "that" impact the performance. I was just obsessed a little recently about the asm code 😂

Source/Charts/Charts/ChartViewBase.swift
jjatie: Why? We never did before. As far as the consumer is concerned, nothing has changed other than it is no longer an implicitly unwrapped optional.

liuxuan30:
alright, one last comment. do you think any chance anyone would nil out some of those old optionals? Like legend.
Corner case: Someone displays legend and later want to remove it, rather than disable, he would like to nil legend out (like to save memory)

jjatie:
In this example, we already have a solution for this which is to set `Legend.Form` to `.none`. The amount of memory saved will be negligible. If we are going to have an optional `Legend` (which does make sense), then we shouldn't have the `.none` case.

jjatie:
You're looking at the internal properties. The consumer facing properties were already non-optional.

liuxuan30:
Let's try the new tests then. looks good to me

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -86,20 +86,24 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
     
     @objc open var highlighter: IHighlighter?
     
     /// object that manages the bounds and drawing constraints of the chart
-    internal var _viewPortHandler: ViewPortHandler!
+    internal lazy var _viewPortHandler: ViewPortHandler = ViewPortHandler(width: bounds.size.width, height: bounds.size.height)
     
     /// object responsible for animations
-    internal var _animator: Animator!
+    internal lazy var _animator: Animator = {
```
liuxuan30: code style for blocks? { another line?

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -344,10 +323,10 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         let position = description.position ?? CGPoint(x: bounds.width - _viewPortHandler.offsetRight - description.xOffset,
                                                        y: bounds.height - _viewPortHandler.offsetBottom - description.yOffset - description.font.lineHeight)
         
-        var attrs = [NSAttributedStringKey : Any]()
-        
-        attrs[NSAttributedStringKey.font] = description.font
-        attrs[NSAttributedStringKey.foregroundColor] = description.textColor
+        let attrs: [NSAttributedStringKey : Any] = [
```
liuxuan30: how about [

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -475,41 +447,28 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-        else
-        {
-            // set the indices to highlight
-            entry = _data?.entryForHighlight(h!)
-            if entry == nil
-            {
-                h = nil
+        var high = highlight
+        guard
+            let h = high,
```
liuxuan30: can we use single let here

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -475,41 +447,28 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+            let entry = _data?.entryForHighlight(h)
+            else {
```
liuxuan30: sorry about missing these code style, { another line

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -475,41 +447,28 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+            // If no highlight or entry, remove all highleted entries and notifgy delegate.
+                high = nil
                 _indicesToHighlight.removeAll(keepingCapacity: false)
-            }
-            else
-            {
-                _indicesToHighlight = [h!]
-            }
+                if callDelegate {
```
liuxuan30: sorry about missing these code style, { another line

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -475,41 +447,28 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-            }
-            else
-            {
-                delegate.chartValueNothingSelected?(self)
-            }
+
+        // set the indices to highlight
+       _indicesToHighlight = [h]
+
+        if callDelegate {
```
liuxuan30: sorry about missing these code style, { another line

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -905,17 +855,11 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         }
         set
         {
-            var val = newValue
-            if val < 0.0
-            {
-                val = 0.0
+            switch newValue {
```
liuxuan30: { another line or?

Source/Charts/Charts/ChartViewBase.swift
jjatie: I guess so?

Source/Charts/Charts/ChartViewBase.swift
jjatie: Looking through the rest of library, `[` and `(` seems to be on the same line.

Source/Charts/Charts/ChartViewBase.swift
jjatie: Unfortunately, no. I tried, but couldn't find a way because we need to assign the original highlight.

Source/Charts/Charts/ChartViewBase.swift
jjatie: I thought we agreed this was the style for multiline guard statements?

Source/Charts/Charts/ChartViewBase.swift
jjatie: I guess so. Seems weird to me on switch statements.

liuxuan30:
@jjatie looks like these commits are no with latest rake file. Feel free to just test on your side or merge master(like practice rebase)

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: oh is it? sorry don't remember.

jjatie:
Oops! Fixing now.

liuxuan30:
any idea why the UT is not latest?

jjatie:
4.0.0 is out of sync with master

liuxuan30:
please merge and rebase on master, once the CI passed, this can be merged I think.

jjatie:
I still don't understand how you want me to rebase 4.0.0 branch since I can't push it up without review. I rebased this PR, but obviously 4.0.0 has not been rebased. 

Also, I noticed that when I rebased from master, the merging took many old commits that I had to manually correct whereas pulling master did not. Very frustrating.

liuxuan30:
Oh I see. If you see the old conflicts you already resolved but once again appear when you rebase, you are on the "right rebase way" I'm talking about, and yes, it's frustrating.

What I meant in the other thread is, I hope you no longer need to have those "merging from upstream master..." commits in your commit log by using rebase.

When you rebase, the old conflicts you resolved will jump again, and if it's easy to fix again, just continue. If it's a lot of burden, let's simply pull master, and then we squash the commits when we want to merge the pull requests on github page.

The whole point using rebase with this project is to squash the commits into one or remove useless  commits before filing the PR or updating the PR. The purpose is to make the project commit logs simple and easy to search.

Hope this clarify? For this PR, let's just pull master, let CI pass and click "squash and merge".

jjatie:
You can't undo a rebase like this.

I figured out what my problem is with the rebasing in this case. You can't rebase from two different branches at the same time which is what we need to do for PRs going into 4.0.0. You can either rebase from master OR 4.0.0, not both. In order for rebasing to work here, 4.0.0 needs to be regularly rebased from master, and then PRs for 4.0.0 need to rebase off the freshly rebased 4.0.0. The problem with this is that there is no way rebase 4.0.0 without review.

Most guides will tell you that you should never rebase a branch someone else might touch, which in this case is the 4.0.0 branch. While I disagree with rebasing in general because you lose history, I'm okay doing it for PR branches. If we are going to rebase, all PRs should be rebasing from the branch they are PRing into. I'll create an issue about this and we can update the docs from there for future PRs to be accepted.

This PR is garbage now, but luckily #3045 includes all the changes I made here. I will update that PR with the current 4.0.0. They should have been the same PR in the first place, but I wasn't sure what changes might be accepted at the time I made them.

liuxuan30:
>You can't undo a rebase like this.

to clarify, can't `git reset --hard`do the job?

jjatie:
Not when you need to rebase from multiple branches

PR #3035, opened 2017-11-23
Fix BalloonMarker's text position calculation, consider insets

yangcaimu:
Consider insets, adjust the position of the label in the context of the center.

### DIFF
--- a/ChartsDemo/Swift/Components/BalloonMarker.swift
+++ b/ChartsDemo/Swift/Components/BalloonMarker.swift
@@ -85,7 +85,9 @@ public class BalloonMarker: MarkerImage {
             y: rect.origin.y))
         context.fillPath()
         
+        rect.origin.x += self.insets.left
         rect.origin.y += self.insets.top
+        rect.size.width -= self.insets.left + self.insets.right
         rect.size.height -= self.insets.top + self.insets.bottom
         
         UIGraphicsPushContext(context)


liuxuan30:
please give a context what's wrong.

jjatie:
This seems to be a good change, but as @liuxuan30 said: What is the problem in the current library?

yangcaimu:
@liuxuan30 @jjatie Has no effect on the current library, repair logic is not rigorous, ignored the horizontal center.

liuxuan30:
again, we understand this should not impact the current library, we just need to know what's the change is about. You changed the x position and width to respect inset, we want to know the user case. Did you meet any issue? Or the x direction inset broke something on your side?

I guess when it was implemented, @danielgindi might considered this already, and intended only y direction.

jjatie:
@yangcaimu This looks like a good change, we just want to document it better so if there is a question or issue in the future, we can quickly understand the justification behind what's there.

yangcaimu:
@liuxuan30 @jjatie 

Thank you for your patience reply, I'll make it clearer. 

At present, the text is centered because the `alignment` attribute of the `_paragraphStyle` attribute is assigned to `center`. When I change to `left`, the text will be completely on the left side of the border, even modify the insets. 

If you do not understand, I can provide Demo.

liuxuan30:
ok. So you change to `.left`, the text aligned on the left is desired result. When you add the inset, what you want to achieve? Adjust the frame by changing `inset`?

yangcaimu:
Yes, especially in the case of multiple lines, to achieve the effect of left and right edge by changing `inset`.

wx20180111-103250 2x

liuxuan30:
@yangcaimu can you post compare images before/after changing the inset?

thierryH91200:
there is also this solution

[RectMarker.swift.zip](https://github.com/danielgindi/Charts/files/1621842/RectMarker.swift.zip)


capture d ecran 2018-01-11 09 52 57
capture d ecran 2018-01-11 09 53 05

yangcaimu:
@liuxuan30 

Before:
`inset = UIEdgeInsets.zero`

wx20180111-103250 2x

After:
`insets = UIEdgeInsets(top: 8, left: 8, bottom: 20, right: 8)`

wx20180111-180248 2x

jjatie:
@liuxuan30 I like this change.

liuxuan30:
sure thing. once we know how it looks after fix, it makes sense then.

PR #3038, opened 2017-11-23
Removed @objc from internal properties

jjatie:
Since the framework is entirely Swift, internal properties do not need to be marked as @objc

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -21,7 +21,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
 {
     /// the maximum number of entries to which values will be drawn
     /// (entry numbers greater than this value will cause value-labels to disappear)
-    @objc internal var _maxVisibleCount = 100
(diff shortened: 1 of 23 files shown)


jjatie:
@liuxuan30 This is a quick and easy one.

liuxuan30:
if it contains more than 3 files it's not a quick one lol

liuxuan30:
I remember we just added those tags since swift 4.
>**Internal** access enables entities to be used within any source file from their defining module, but not in any source file outside of that module. You typically use internal access when defining an app’s or a framework’s internal structure.

Does this mean no one need access to these internals outside Charts framework anymore? If true, we can remove then.

jjatie:
the internal access method means that the properties/methods can only be used within the Charts framework. Any consumer of the framework cannot see them.

Since any consumer of the framework cannot see them, and the framework is written in Swift internally, all internal and private functions and properties do not need to be marked as `@objc` because they will never be used by objective-c

liuxuan30:
yes, but the trick is adding @objc will let Objc see it and break the rules

liuxuan30:
for math refineing I have to delay one day. get back to work

PR #3039, opened 2017-11-23
weak -> unowned

jjatie:
`ViewPortJob`s are owned by the Charts that make them. They are guaranteed to only exist while the chart exists. The `Transformer` and `ViewPortHandler` are supplied by the chart, so they will also only exist while the chart exists. Therefor none of them need to be `weak`, but can be `unowned` instead. It's a minor change in the code base (removing some guard statements), but it makes it much easier to discern how the framework is architected.


liuxuan30:
CI is not latest? Please pull 4.0

jjatie:
@liuxuan30 done

liuxuan30:
let's wait result

jjatie:
@liuxuan30 Any idea why Travis CI isn't running anything?

liuxuan30:
don't know.. I will cancel and requeue

liuxuan30:
I just found many CI is still in created state. I will try if I could contact Travis

liuxuan30:
they are saying 
>We are currently investigating an increased back log for container-based infrastructure. 

Guess that's the reason?

liuxuan30:
I have restarted the top 3 PRs' CI. Let's see if it can get started.

jjatie:
Travis says it's fully operational again. I just tried restarting this one again.

liuxuan30:
@jjatie we can move forward now.

PR #3042, opened 2017-11-23
fileprivate -> private

jjatie:
almsot all cases of fileprivate were meant to be private. If there is a need to use fileprivate in the future, it should be considered then. Not now.


jjatie:
@liuxuan30 This is a quick and easy one.

liuxuan30:
@danielgindi @petester42 do you remember why use `fileprivate` back then?

liuxuan30:
when we need `fileprivate`, we could add them later.

petester42:
Probably because private was not needed and the semantics were a little different back then.

PR #3043, opened 2017-11-23
Removed redundant ivars in BarLineChartViewBase

jjatie:
In favour of proper access control


Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -57,19 +57,28 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     /// **default**: false
     @objc open var keepPositionOnRotation: Bool = false
     
-    /// the object representing the left y-axis
-    internal var _leftAxis: YAxis!
+    /// The left y-axis object. In the horizontal bar-chart, this is the
+    /// top axis.
+    @objc open internal(set) var leftAxis: YAxis!
```
liuxuan30: what's the `set` purpose here?

liuxuan30:
I'm not sure if we want to remove `_` here. I'm kind of supporting using `_` as it implies "interval", is this against latest swift trend about naming?

Source/Charts/Charts/BarLineChartViewBase.swift
jjatie: Previously there were two variables
```
internal var _leftAxis: YAxis!
@objc open var leftAxis: YAxis! {
    return _leftAxis
}
```

There is no need for this as swift provides better access control than objective-c did previously. `open internal(set)` (can also be `open private(set)` but was holding off on that one until I evaluated whether that was appropriate) makes the variable `open`, but only allows it to be set by internally.

jjatie:
It's certainly a lot less common in Swift. In this case we don't need to have a special internal variable, as the backing ivar, and the internal ivar don't have any side effects.

Source/Charts/Charts/BarLineChartViewBase.swift
liuxuan30: wow, I don't realize swift add this already.

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -57,19 +57,28 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
...
+    /// **default**: An instance of YAxisRenderer
+    @objc open var leftYAxisRenderer: YAxisRenderer!
 
-    internal var _leftYAxisRenderer: YAxisRenderer!
-    internal var _rightYAxisRenderer: YAxisRenderer!
+    /// The right Y axis renderer. This is a read-write property so you can set your own custom renderer here.
+    /// **default**: An instance of YAxisRenderer
+    @objc open var rightYAxisRenderer: YAxisRenderer!
     
     internal var _leftAxisTransformer: Transformer!
```
liuxuan30: how about these two and some others with `_`?  Like `_tapGestureRecognizer` It means we still need `_` for some?

jjatie: It means I didn't touch them. I think they shouldn't have the `_`.

liuxuan30:
Generally I'm ok with this. 
 ideas and opinions @danielgindi @petester42? @jjatie let's wait sometime to see if they have any. If you need this merged in a short time or other PRs depend on this, let me know, if like 1-2 weeks passed, we can move forward.
BTW, is this better we put this into 4.0 branch or just a minor refactor?

jjatie:
Sounds good. This doesn't change the exposed interface for consumers, so I think it is okay to do the PR before 4.0. This and a few other PRs are setting me up for some more major changes which I will submit to 4.0 in the new year.

jjatie:
@liuxuan30 It's been two weeks. Is it safe to merge now?

liuxuan30:
I'm ok with this PR. But I think we should first merge the unit test PR, to make sure the demo in the future? If you manually checked the tests passed and demo builds, it's ok to 'squash and merge'

Source/Charts/Charts/BarLineChartViewBase.swift
liuxuan30: Just a moment. Have you checked if people can set axis outside from the framework? Just like other PRs similar issue?

liuxuan30: I just did a quick check, axis is not settable outside, seems safe. Just run the tests and demo:)

PR #3045, opened 2017-11-23
Chartviewbase redundant ivar

jjatie:


jjatie:
This includes #3034

liuxuan30:
do we need to close https://github.com/danielgindi/Charts/pull/3034, or it's a dependency

jjatie:
3034 should be done first is what I meant

liuxuan30:
sure, I'm taking #3034 first

jjatie:
@liuxuan30 I rebased this one off of 4.0.0 as I mentioned in #3034.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -39,43 +39,51 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-    
     /// The default IValueFormatter that has been determined by the chart considering the provided minimum and maximum values.
     internal var _defaultValueFormatter: ValueFormatter? = DefaultValueFormatter(decimals: 0)
-
-    /// object that holds all data that was originally set for the chart, before it was modified or any filtering algorithms had been applied
-    internal var _data: ChartData?
     
+    /// The data for the chart
+    @objc open var data: ChartData? {
+        didSet {
```
liuxuan30: does `didSet` have any advantage over the old setter? Or just `didSet` describe the job more properly?

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -103,9 +125,9 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     
     /// flag that indicates if offsets calculation has already been done or not
     private var _offsetsCalculated = false
-    
-    /// array of Highlight objects that reference the highlighted slices in the chart
-    internal var _indicesToHighlight = [Highlight]()
+
+    /// The array of currently highlighted values. This might an empty if nothing is highlighted.
+    @objc open internal(set) var highlighted = [Highlight]()
```
liuxuan30: A) Can we keep old name, or the new name is more swift favored? 
B) If we gonna change the public var names like this, we might need to write them down and when publishing 4.0, users could know what's thre replacement.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -384,26 +349,22 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
-        return _indicesToHighlight.count > 0
+        return highlighted.count > 0
     }
 
     /// Highlights the values at the given indices in the given DataSets. Provide
     /// null or an empty array to undo all highlighting. 
     /// This should be used to programmatically highlight values.
     /// This method *will not* call the delegate.
-    @objc open func highlightValues(_ highs: [Highlight]?)
+    @objc open func highlightValues(_ highs: [Highlight])
```
liuxuan30: Looks we have a regression here

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -448,8 +409,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     /// - parameter callDelegate: Should the delegate be called for this change
     @objc open func highlightValue(x: Double, y: Double, dataSetIndex: Int, callDelegate: Bool)
     {
-        guard let data = _data else
-        {
+        guard let data = data else {
```
liuxuan30: regression code style. Please check #3034 for code style issues and update here? If you can manage to reset hard with #3034, we can merge #3034 so you won't have to do dup work. up to you

liuxuan30:
Besides, can you pull master into this PR, so we have the newer CI?

Source/Charts/Charts/ChartViewBase.swift
jjatie: I don't understand. The old way had two variables, one stored and one computed. This just has the one stored with a property observer to do the calculations.

Source/Charts/Charts/ChartViewBase.swift
jjatie: The public facing name was `highlighted` which is why I changed the internal var name.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -39,43 +39,53 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
     /// If set to true, chart continues to scroll after touch up
     @objc open var dragDecelerationEnabled = true
-    
-    /// Deceleration friction coefficient in [0 ; 1] interval, higher values indicate that speed will decrease slowly, for example if it set to 0, it will stop immediately.
-    /// 1 is an invalid value, and will be converted to 0.999 automatically.
-    private var _dragDecelerationFrictionCoef: CGFloat = 0.9
-    
+
     /// if true, units are drawn next to the values in the chart
+    // TODO: This is used nowhere and can't be used by a consumer. Can we remove this property?
```
jjatie: What's the deal with this variable?

jjatie:
If you let me know what the next PR you're going to tackle is, I can stay one ahead of you ; )

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: I mean setter vs. didSet.

Source/Charts/Charts/ChartViewBase.swift
liuxuan30: if this is really not used anywhere, we can safely delete, or add @deprecated first.

liuxuan30:
The review order is up to you, I'm ok. I usually prefer finish off small ones, and then larger one. But if any PR can ease your workflow, I can take that first.

It looks like CI failed because of the demo controllers' `chartDescription` deprecation. Could you fix it so we can merge?

Source/Charts/Charts/ChartViewBase.swift
jjatie: It's not used anywhere so I will delete.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -439,10 +442,12 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         guard
             let h = high,
             let entry = data?.entryForHighlight(h)
-            else {
+            else
+        {
```
liuxuan30: indent seems wrong here.

jjatie:
Pushing now

Source/Charts/Charts/ChartViewBase.swift
jjatie: It's what Xcode auto indents to. It lines up with the guard instead of the else. I think it makes sense because in a `guard` statement the "else" is syntactic sugar. `guard` is really just a fancy `if` and in `if` statements the "{" lines up the same way.

jjatie:
#3086 next would be good : )

liuxuan30:
all green finally

PR #3054, opened 2017-11-25
Utils Cleanup

jjatie:
Replaced unnecessary getters with proper access control
Replaced unnecessary convenience inits with default parameters
Minor refactoring


Rakefile
```diff
@@ -3,7 +3,11 @@ def type
 end
 
 def project_name
-  'Charts.xcodeproj'
+  'ChartsDemo/ChartsDemo.xcodeproj'
```
liuxuan30: I think we have merged master into 4.0 already? But these jumps out again

jjatie:
FYI, new PRs from me will be rebasing. I'm scared to do it with the current ones since they switched from master -> 4.0.0.

Rakefile
jjatie: Just haven't pulled from 4.0.0 in a while.

liuxuan30:
what you mean? I thought since 4.0 is  already sync with master, the rest of PRs can easily pull 4.0 again and simply resolve conflicts.

Source/Charts/Utils/Transformer.swift
```diff
@@ -17,50 +17,50 @@ import CoreGraphics
...
         {
             scaleY = 0.0
         }
 
         // setup all matrices
-        _matrixValueToPx = CGAffineTransform.identity
-        _matrixValueToPx = _matrixValueToPx.scaledBy(x: scaleX, y: -scaleY)
-        _matrixValueToPx = _matrixValueToPx.translatedBy(x: CGFloat(-chartXMin), y: CGFloat(-chartYMin))
+        matrixValueToPx = CGAffineTransform.identity
+            .scaledBy(x: scaleX, y: -scaleY)
```
liuxuan30: is this a chained call? how about putting in on line. seems a little distractive

Source/Charts/Utils/ViewPortHandler.swift
```diff
@@ -250,160 +227,91 @@ open class ViewPortHandler: NSObject
...
-        
-        if newValue == 0.0
-        {
-            newValue = CGFloat.greatestFiniteMagnitude
-        }
-        
-        _maxScaleX = newValue
-        
-        limitTransAndScale(matrix: &_touchMatrix, content: _contentRect)
+        maxScaleX = xScale == 0 ? .greatestFiniteMagnitude : xScale
```
liuxuan30: is it a  good practice to use 0 as hint to  CGFloat?

Source/Charts/Utils/ViewPortHandler.swift
```diff
@@ -250,160 +227,91 @@ open class ViewPortHandler: NSObject
...
         
         chart.setNeedsDisplay()
         
-        return _touchMatrix
+        return touchMatrix
     }
     
     /// limits the maximum scale and X translation of the given matrix
-    private func limitTransAndScale(matrix: inout CGAffineTransform, content: CGRect?)
+    private func limitTransAndScale(matrix: inout CGAffineTransform, content: CGRect)
```
liuxuan30: I'm not on my mac now, are we sure all `limitTransAndScale` callers only pass `self.contentRect`?

Source/Charts/Utils/Transformer.swift
jjatie: This is a fairly standard Swift practice.

Source/Charts/Utils/ViewPortHandler.swift
jjatie: I am sure.

Source/Charts/Utils/ViewPortHandler.swift
jjatie: `0 == 0.0`. I think the reason `0.0` is used elsewhere is carryover from Objective-C

liuxuan30: yes, 0.0 looks like more natural as float, while 0 gives me an int imply..

PR #3079, opened 2017-12-08
Rename notifyDataSetChanged() to reloadData()

jbehrens94:
As `notifyDataSetChanged()` is a Java thing, I changed the method to `reloadData()`. This improves the API for Swift users, as this feels more natural to them.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -70,7 +70,7 @@ open class ChartDataSet: ChartBaseDataSet
     internal var _xMin: Double = Double.greatestFiniteMagnitude
     
     /// *
-    /// - note: Calls `notifyDataSetChanged()` after setting a new value.
+    /// - note: Calls `reloadData()` after setting a new value.
     /// - returns: The array of y-values that this DataSet represents.
     @objc open var values: [ChartDataEntry]
     {
@@ -81,12 +81,12 @@ open class ChartDataSet: ChartBaseDataSet
         set
         {
             _values = newValue
-            notifyDataSetChanged()
+            reloadData()
         }
     }
     
     /// Use this method to tell the data set that the underlying data has changed
-    open override func notifyDataSetChanged()
+    open override func reloadData()
     {
         calcMinMax()
     }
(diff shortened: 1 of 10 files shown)


liuxuan30:
I guess keep this name is for alignment for it's android portion MPAndroidChart, to keep the learning curve singleton.

jbehrens94:
I would disagree with the approach, then. As iOS use `reloadData()` for `UITableView`, `UICollectionView` for example, it would make sense to use that. I personally think this makes life easier for Swift developers who don't have an Android background.

liuxuan30:
I would agree the name is a bit distracting. @danielgindi @PhilJay @petester42  what you think about this.

jbehrens94:
There's some hiccups going on, suddenly the tests don't pass anymore and there was a merge conflict. However, I think this has to do with a PR merged to master before mine, could someone check this? @liuxuan30 @danielgindi

liuxuan30:
A recently swift refactor may change some code and lead to fail. 
@jjatie what you think about changing the name and either @jbehrens94 should fix it or you. FYI I have Daniel reply, he should add you in soon.

jjatie:
While it does appear to fall in line with other iOS API (e.g. UITableView), I disagree with the name change as it is misleading. `reloadData()` on a `UITableView` is an expensive method call as it actually reloads the table view's datasource. In this case, we aren't reloading anything, but instead updating helper variables using the data of the new datasource.

jjatie:
We shouldn't be able to do any merges without all the tests passing, but there seems to have been something that's gotten through. I will look into it.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -75,6 +75,28 @@ open class ChartDataSet: ChartBaseDataSet
     
     /// minimum x-value in the value array
     internal var _xMin: Double = Double.greatestFiniteMagnitude
+
+    /// *
+    /// - note: Calls `reloadData()` after setting a new value.
+    /// - returns: The array of y-values that this DataSet represents.
+    @objc open var values: [ChartDataEntry]
```
jjatie: This variable no longer exists. Please remove and update the `didSet` method on the new variable of the same name.

jjatie:
`reloadData` is misrepresents what the method is actually doing.

PR #3088, opened 2017-12-11
Moved the default value formatter

jjatie:
It is now simply `DefaultValueFomatter()`
Removed unnecessary backing ivars in `DefaultValuetFormatter` in favour of property observers
Deprecated static func constructor in favour of initializer


liuxuan30:
do we need to wait #3024 ?

jjatie:
Nope

Source/Charts/Utils/ChartUtils.swift
```diff
@@ -226,7 +226,7 @@ extension CGContext {
         NSUIGraphicsPopContext()
     }
 
-    internal func drawMultilineText(_ text: String, at point: CGPoint, constrainedTo size: CGSize, anchor: CGPoint, knownTextSize: CGSize, angleRadians: CGFloat, attributes: [NSAttributedStringKey : Any]?)
+    func drawMultilineText(_ text: String, at point: CGPoint, constrainedTo size: CGSize, anchor: CGPoint, knownTextSize: CGSize, angleRadians: CGFloat, attributes: [NSAttributedStringKey : Any]?)
```
liuxuan30: how about we make these two `drawMultilineText` as open? I don't see why only internal.

Source/Charts/Formatters/DefaultValueFormatter.swift
```diff
@@ -22,69 +23,65 @@ open class DefaultValueFormatter: NSObject, ValueFormatter
     
     @objc open var block: Block?
     
-    @objc open var hasAutoDecimals: Bool = false
+    @objc open var hasAutoDecimals: Bool
```
liuxuan30: why remove default value here but initialize later in init() many times?

Source/Charts/Formatters/DefaultValueFormatter.swift
```diff
@@ -22,69 +23,65 @@ open class DefaultValueFormatter: NSObject, ValueFormatter
...
         }
     }
     
     public override init()
     {
-        super.init()
-        
-        self.formatter = NumberFormatter()
+        formatter = NumberFormatter()
+        formatter?.usesGroupingSeparator = true
```
liuxuan30: why set usesGroupingSeparator and decimals here while the old code does not? did I miss anything

liuxuan30:
there is a `DefaultAxisValueFormatter` looks very same as `DefaultValueFormatter`, except for some method fingerprints. Shall we unify them?

Source/Charts/Utils/ChartUtils.swift
jjatie: First, it's always been internal. Second, we expose a way to extend the draw methods by wrapping them in another open method. I think it's safe to keep it internal. If we get requests to expose it, maybe we can change our mind then.

Source/Charts/Formatters/DefaultValueFormatter.swift
jjatie: Because it has different values based on which initializer. This way it's not assigned twice.

Source/Charts/Formatters/DefaultValueFormatter.swift
jjatie: I made the empty initializer provide the same formatter that `ChartUtils.defaultFormatter` had which set it to true

Source/Charts/Utils/ChartUtils.swift
jjatie: Also, I don't think it makes sense to have any of these be open, because they would have to subclass `CGContext` which isn't recommended.

Source/Charts/Formatters/DefaultValueFormatter.swift
liuxuan30: alright, makes sense then.

Source/Charts/Formatters/DefaultValueFormatter.swift
liuxuan30: Oh sorry. I missed the removal

jjatie:
There is a lot that is the same between the two classes, but I don't see an obvious way to merge them. Maybe we should do it if it's possible, but I think that's a separate PR.

liuxuan30:
conflicts to merge. Take your time, as CI will be longer.

PR #3104, opened 2017-12-14
Replaced backing variables with proper access control

jjatie:
This is a non-breaking change
No logic was changed, and the external interface remains the same. Only internal names have been changed (dropping "_")
This excludes Bool values, as to change those breaks the external interface. Those changes are handled by #2687 
Greatly reduces the code base, and keeps logic together.


jjatie:
@liuxuan30 Could you make this one a priority? It will cause merge conflicts in many of my PRs and I would like to get them out of the way.

There are no logic changes, just removing backing variables across the board (except for Bools as noted in PR notes) like you've seen in a few other PRs already.

jjatie:
@liuxuan30 I'm investigating why the test are failing. I'll let you know when it's fixed.

liuxuan30:
Sorry just saw this. #3034 and this, which one first?

jjatie:
#3034

liuxuan30:
It's the end of year, so everyone is quite occpuied :) I will try review 1-2 PR a week 😂

Source/Charts/Components/AxisBase.swift
```diff
@@ -200,15 +192,31 @@ open class AxisBase: ComponentBase
...
-    /// Do not touch this directly, instead, use axisMinimum.
-    /// This is automatically calculated to represent the real min value,
-    /// and is used when calculating the effective minimum.
-    internal var _axisMinimum = Double(0)
+    /// The minimum value for this axis.
+    /// If set, this value will not be calculated automatically depending on the provided data.
+    /// Use `resetCustomAxisMin()` to undo this.
+    @objc open internal(set) var axisMinimum = 0.0 {
+        willSet {
+            _customAxisMax = true
```
liuxuan30: why willSet here? can't be in didSet?

Source/Charts/Components/AxisBase.swift
```diff
@@ -200,15 +192,31 @@ open class AxisBase: ComponentBase
...
     internal var _customAxisMax: Bool = false
     
-    /// Do not touch this directly, instead, use axisMinimum.
-    /// This is automatically calculated to represent the real min value,
-    /// and is used when calculating the effective minimum.
-    internal var _axisMinimum = Double(0)
+    /// The minimum value for this axis.
+    /// If set, this value will not be calculated automatically depending on the provided data.
+    /// Use `resetCustomAxisMin()` to undo this.
+    @objc open internal(set) var axisMinimum = 0.0 {
```
liuxuan30: axisMinimum is also accessed from outside of framework.

Source/Charts/Components/AxisBase.swift
```diff
@@ -200,15 +192,31 @@ open class AxisBase: ComponentBase
...
+    }
     
-    /// Do not touch this directly, instead, use axisMaximum.
-    /// This is automatically calculated to represent the real max value,
-    /// and is used when calculating the effective maximum.
-    internal var _axisMaximum = Double(0)
+    /// The maximum value for this axis.
+    /// If set, this value will not be calculated automatically depending on the provided data.
+    /// Use `resetCustomAxisMax()` to undo this.
+    @objc open internal(set) var axisMaximum = 0.0 {
```
liuxuan30: Same to axisMinimum. Seems you missed some context to make setter internal?

Source/Charts/Components/AxisBase.swift
jjatie: I was following the order in the former setters because I didn't want to change any logic in this PR. I agree it can all be done in didSet. Should I change it?

Source/Charts/Components/AxisBase.swift
jjatie: I think I was just tired and missed a couple things ; )

Source/Charts/Components/AxisBase.swift
liuxuan30: I think so. The old setter is atomic I assume, so will/did seems the same in our case.

Source/Charts/Components/AxisBase.swift
liuxuan30: Using Demo to test could help trigger the errors?

jjatie:
This is too big of a change to switch from master to 4.0.0. Will implement this in pieces as part of other changes.

PR #3106, opened 2017-12-14
Removed optionality from valueFormatter where appropriate

jjatie:
In ChartBaseDataSet, `valueFormatter` never returned nil, and escaped early if trying to set it to nil. It appears this was made optional solely to provide lazy initialization. We now use a lazy var instead. This also removes the need for `needsFormatter`

In AxisBase, the backing var `_axisValueFormatter` would never be treated as nil, and appears to be made optional solely to provide lazy initialization. We now use a lazy var instead. In `valueFormatter` we can remove the `nil` check, but leave it optional to keep the same functionality.


jjatie:
@liuxuan30 This has the xcodeproj issue again. I think it comes from being originally on master, and me changing it to 4.0.0

liuxuan30:
I think it's your tab/space mess up issue as I said in another thread. You can look it up see what's wrong. By default Xcode project is using tab. However in your commit it changes to space.

liuxuan30:
conflict and issue with project.pbxproj

jjatie:
Fixed

jjatie:
@liuxuan30 fixed

liuxuan30:
weird, the xproj issue happens again...

jjatie:
ugh. I'll fix it.... Let's see what happens after #3039 gets merged first though. Can you review the rest of it and approve/comment. That way I can merge it when I fix the xproj issue if you're not around.

liuxuan30:
This change is ok except for the project file. Do you think you can fix it? 
two options if cannot fix:
1. mix reset, only commit other changes
2. merge, and we revert project file change

jjatie:
I will fix it if #3039 doesn't fix it.

liuxuan30:
we can merge #3039 now

liuxuan30:
travis is really slow now.. almost 48H and tvOS is still pending.. the other two passed.

jjatie:
Yea. When I check (about 12 hours ago) the backlog was at 1800

Issue #2257, opened 2017-03-18
HorizontalBarChart does not update labels correctly


defranke:
I'm using the HorizontalBarChart and have lots of bars to display. Each bar has its own label.
If I'm zoomed out, the HorizontalBarCharts only display as expected ~10 Labels, but if I zoom in, the labels, which should be shown, are not updated and thus it's still just showing the same 10 labels.
Thus, if I zoom in more, I don't have any labels any more. Basically it doesn't recompute min/max-values of the axis and updates the labels accordingly.

For example, here is the BarChart completely zoomed out, which display the correct labels:
![screen shot 2017-03-18 at 23 54 57](https://cloud.githubusercontent.com/assets/12988965/24076659/9a3c9b94-0c36-11e7-84b3-041842c67d5f.png)

And here is a screenshot after zooming in, which now displays still the same labels:
![screen shot 2017-03-18 at 23 55 02](https://cloud.githubusercontent.com/assets/12988965/24076662/abce57d0-0c36-11e7-93ab-6917e67ad1d1.png)

I configured the xAxis as follows:

        xAxis.granularityEnabled = false
        xAxis.granularity = 1.0
        xAxis.valueFormatter = IndexAxisValueFormatter(values: dataPoints)
        xAxis.avoidFirstLastClippingEnabled = false
        xAxis.labelCount = dataPoints.count

I already tried to fix it myself and it seems to me, that there is a Bug in XAxisRendererHorizontalBarChart Line 42:
`if viewPortHandler.contentWidth > 10 && !viewPortHandler.isFullyZoomedOutX`

Since everything is rotated and in an HorizontalBarChart the xAxis is actually the yAxis, we need to check if the y-Axis is fully zoomed out:
`if viewPortHandler.contentWidth > 10 && !viewPortHandler.isFullyZoomedOutY`
 
In my case this fixes the issue.
Maybe someone with more experience in the library can also take a look at it.

joshiaProgram:
thanks for share, it's effective for me

liuxuan30:
are you able to fix it and send out a pull request?

siyu6974:
#this fix actually caused a bug in my app. It created more labels than needed. Weird enough it only starts after I have more than 6 entries.

<img width="231" alt="screen shot 2018-02-26 at 00 50 34" src="https://user-images.githubusercontent.com/15876323/36648100-16133c84-1a8f-11e8-9a20-460609f9b197.png"
#3114

[...]

PR #3114, opened 2017-12-18
Fixed using wrong axis (Issue #2257)

defranke:
Fix for Issue #2257


liuxuan30:
I found another issue while  testing, the label sometimes disappeared when zoomed in only one full bar remains on the screen. I will file a fix soon.

Issue #3116, opened 2017-12-19
ChartsDemo doesn't compile


valeriyvan:
ChartsDemo target of ChartsDemo project is not compiling after last changes

liuxuan30:
what.. The CI indicates it's all green and then could I merge. Why not compiling

liuxuan30:
I just checked the CI log. I just compile the framework and run UT, not compiling the demos. I think we should avoid that in the future. Like adding a UT to build demos.

jjatie:
Sounds like a good idea to add compiling demos to the CI. I often forget to check them.

jjatie:
I just checked the master and the demos (Mac and iOS objc/swift) are building fine.

liuxuan30:
Shall we merge #3121 then? BTW, please check #3121 comments, do you know anything we can reduce the compile times?

PR #3121, opened 2017-12-20
add option to build demo projects unit tests on iOS

liuxuan30:
To avoid any refactoring change that will impact demo builds like #3116, I added the build process for ChartsDemo and its swift portion.

Completely new to rake, it's weird google does not give enough tutorials towards rake  basis.

It took me a wild guess what's the difference between `is_build_demo = false` and `is_build_demo: false` as I just follow `xcprety_args: ''` to guess it's the right way to specify default value.

I change to 'ChartsDemo/ChartsDemo.xcodeproj' as umbrella project. 
I specifically put build before test, to list test results at the bottom of the output.

One thing I don't know why is it builds Charts framework triple times, though I don't ask for clean. It's what we could improve, so welcome suggestions.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@ def type
 end
 
 def project_name
-  'Charts.xcodeproj'
+  'ChartsDemo/ChartsDemo.xcodeproj'
 end
 
 def configuration
@@ -26,6 +26,13 @@ end
 def build_schemes
   %w[
     Charts
+  ]
+end
+
+def build_demo_schemes
+  %i[
+    ChartsDemo
+    ChartsDemo-Swift
   ]
 end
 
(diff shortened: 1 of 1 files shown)


petester42:
The Rakefile is just a ruby script that has special semantics for tasks. I’m sure you can find plenty of ruby resources out there. 

I like the idea of building the demos too to make sure they still run but I’d like to refactor the script to not have to change the execute command.

It builds 3 time because it builds iOS, macOS and tvOS.

liuxuan30:
@petester42 no the 3 times means when running for ChartsDemo, ChartsDemo-Swift, and test, each target build once.

petester42:
Test runs build also. You don’t have to build first.

petester42:
I don’t think I understand what the problem is.

liuxuan30:
I already changed the procedure, take a look at the script. I already build first. So I assume test shouldn't need it.

petester42:
I don’t think it’s that smart about that.

liuxuan30:
I'm not sure, it's still xcode, and I don't put them into separate machine or state, so I think building once should be enough, the framework is already there, xcodebuild won't find it?
Anyway, if nothing we can do, shall we merge this?

jjatie:
We should also test the Mac demos.

I don't think we can reduce the number of compiles. Even if we did it wouldn't make much of a difference on Travis CI (the project is fairly small).

liuxuan30:
I'm adding OS X demos.

liuxuan30:
there is space to improve the rake script. but make it work for now.

liuxuan30:
Let's merge this first.

jjatie:
Sounds good. I don't know how to better improve the script. Shall we go ahead and merge?

liuxuan30:
let's make it work first. from now on, safer for demos

PR #3130, opened 2017-12-24
Updating 4.0.0 with latest changes in master

jjatie:


liuxuan30:
@jjatie please try squash the commits into 4.0 as well. Now you have the access, I think we could come up a new model how do we deal with 4.0 refactoring. e.g. do you push all changes into 4.0, and we review once for all, or still we review every PR?

The cons and pros are that, with a huge single PR from 4.0 to master, it would took lot of time to review. With sub PRs, the workload is split, but the review progress may slow you down.

jjatie:
We want to squash commits coming from the master too? I assumed I would squash everything except for what was in master.

liuxuan30:
the idea is to make the commit logs simple, so the old fashion is we squash into one commit before filing the PR. But now since Github provide such feature, we are not forcing quash on client side anymore.

jjatie:
Will do!

liuxuan30:
welcome on board :)

PR #3132, opened 2017-12-25
Fix axis label disappear when zooming in deep enough

liuxuan30:
Find another issue when testing  #3114
fix axis label missing when zoom in deep enough, sometimes the label will disappear while scrolling.
Turned out `first` and `last` is equal, so `n` remains 0, and `axis.entries` is empty then.

I do see `var n = axis.centerAxisLabelsEnabled ? 1 : 0` at first, but rather than force setting n = 1, I think checking n == 0 specifically may be better. Welcome suggestions.

### DIFF
--- a/Source/Charts/Renderers/AxisRendererBase.swift
+++ b/Source/Charts/Renderers/AxisRendererBase.swift
@@ -163,7 +163,11 @@ open class AxisRendererBase: Renderer
                     n += 1
                 }
             }
-            
+            else if last == first && n == 0
+            {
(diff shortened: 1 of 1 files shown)


jjatie:
Should we be concerned that Codecov wasn't able to diff it?

liuxuan30:
have no idea. Codecov is like blackbox to me.

liuxuan30:
@jjatie please review this. a small fix seems right before 3.0.5

PR #3135, opened 2017-12-25
Update 4.0.0 with master

jjatie:
No changes to 4.0.0 yet, so I will run tests/demo and merge.


liuxuan30:
what are these empty commits for?
Do you know how to squash (or rebase children of xxx interactively from SourceTree) on your side? I'm feeling you can use that on your side to avoid some commits.

If you never tried, you can now. Daniel used to tell me about this, and it makes my commit logs much more clear.

jjatie:
So we are rebasing and (force) pushing?

I can do that, but wasn't sure so I stuck with the safer choice.

liuxuan30:
why need force push here? (force push to 4.0 branch may required, but force push to master should never happen)
for example, you have lots of commits in 4.0, ahead master. But you need to merge master into 4.0 now, and after merge, 
if you use SourceTree, just rebase on the master's last commit, and start rebase interactively (you may need to solve conflicts again), and then 4.0 branch commits will be upon latest master commit, without any additional commit(like merging upstream master...)
You can even choose what commits to be squash into one, so it will very easy to mange the commits.

If you already knew this, ignore what I said, but if you don't, just give some practice.

jjatie:
So to be clear, I can push up a rebase without any review?

liuxuan30:
I don't think so. What I asked is about this PR, without any change but only commits in master, which is unnecessary.
What I meant is you can manage your PR's commits more clear and straightforward, without those annoying "merging remote-tracking.." commits

PR #3136, opened 2017-12-26
Renderer protocols

jjatie:
`Renderer`, `DataRenderer` and `AxisRenderer` are now protocols instead of classes.

Slight more code, but allows for better type safety and removes a lot of unnecessary type conversion. It's a "Swift-ier" way.

This a resubmit of #2990 because there were major merge conflicts in that one.


jjatie:
This was approved in #2990

PR #649, opened 2016-01-05
Add a gradient line chart

Regno:
For [#537](https://github.com/danielgindi/ios-charts/issues/537)
Add functionality to draw a gradient line for line chart. It use exist property "colors" and new property "gradientPositions" for rendering gradient line.
Add two new public properties: "gradientPositions" and "drawGradientEnabled" for LineChartDataSet.
Add two new private functions: "drawGradientLine" and "generatePath" for LineChartRenderer

Example of usage:

``` swift
override func viewDidLoad() {
        super.viewDidLoad()
        //set x values
        let count = 10
        var xVals: [String]! = []
        for (var i = 0; i < count; i++)
            {
                xVals.append("\(i)")
            }
        //set y values
        let range = 5.0
        var yVals: [ChartDataEntry] = []
        for (var i = 0; i < count; i++)
            {
                let val = drand48() * range + 35
                yVals.append(ChartDataEntry(value: val, xIndex: i))
            }
        let set1 = LineChartDataSet(yVals: yVals, label: "DataSet 1")
        //Set gradient
        set1.drawCubicEnabled = true
        set1.drawGradientEnabled = true
        chartView.drawGridBackgroundEnabled = false
        //set1.drawFilledEnabled = true
        set1.lineWidth = 2.0
        set1.circleRadius = 2.0
        set1.colors = [UIColor(red: 0.192, green: 0.686, blue: 0.980, alpha: 1.00), UIColor(red: 0.212, green: 0.863, blue: 0.318, alpha: 1.00), UIColor(red: 0.996, green: 0.867, blue: 0.275, alpha: 1.00), UIColor(red: 0.980, green: 0.090, blue: 0.157, alpha: 1.00)]
        set1.gradientPositions = [35.0, 36.6, 38.0, 40.0]
        // set Data
        var dataSets: [LineChartDataSet] = []
        dataSets.append(set1)
        let chartData = LineChartData(xVals: xVals, dataSets: dataSets)
        chartView.data = chartData
}
```

Result:

PR #3141, opened 2017-12-28
Added gradient line drawing to LineChartRenderer

jjatie:
Reimplemented #649 for modern Swift and Charts


liuxuan30:
In the future, you can create a feature branch directly in this repo, so we can help pull and test, rather than your fork. (for bugfix, either is ok)

jjatie:
#3142

PR #3149, opened 2018-01-03
Subclassing of LegendRenderer didn't take any effect

l-lemesev:
I was trying to subclass the LegendRenderer for a custom pie chart legend, but swapping the `legendRenderer` property didn't take any effect because the internal var was used instead of the open one.


liuxuan30:
@jjatie I don't remember if this will be conflict with 4.0 refactoring? Or we can close this if it'x fixed in 4.0

jjatie:
It won't be a conflict. It is fixed in 4.0 (by accident), but since this is a bug fix it should also go in master.

liuxuan30:
ok.

PR #3159, opened 2018-01-07
fix programatical unhighlighting for BarCharView

jekahy:
There was an issue with reseting the _highlight_ manually from code. The _lastHighlighted_ property of **ChartViewBase** was not set to nil. Thus, calling  **chartView.highlightValue(nil)** in the delegate method  _chartValueSelected(_ chartView: entry: highlight:)_ did not do anything, and the next time user tapped a bar the chartValueNothingSelected method was called instead of the intended chartValueSelected.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -533,13 +533,13 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             
             if h === nil || h == self.lastHighlighted
             {
-                highlightValue(nil, callDelegate: true)
                 lastHighlighted = nil
+                highlightValue(nil, callDelegate: true)
             }
             else
             {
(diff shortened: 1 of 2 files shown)


jjatie:
Thanks!

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -533,13 +533,13 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
...
             {
-                highlightValue(nil, callDelegate: true)
                 lastHighlighted = nil
+                highlightValue(nil, callDelegate: true)
             }
             else
             {
-                highlightValue(h, callDelegate: true)
                 lastHighlighted = h
+                highlightValue(h, callDelegate: true)
```
liuxuan30: why we change the order here?

jjatie: #3178

PR #3178, opened 2018-01-13
Removed redundant `lastHighlighted = nil`

jjatie:
`highlightValue` already takes care of it.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -533,7 +533,6 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
             
             if h === nil || h == self.lastHighlighted
             {
-                lastHighlighted = nil
                 highlightValue(nil, callDelegate: true)
             }
             else


liuxuan30:
I only see `highlightValues(_ highs: [Highlight]?)` nil out `lastHighlighted`, but `highlightValue` does not. 
There used to be a PR in `func clear()` specifically set `lastHighlighted` to nil because some small bugs, not cleaning `lastHighlighted` state and causes wrong highlight.
I think we need a deeper look to make sure this PR is safe.

liuxuan30:
@jjatie where you say `lastHighlighted = nil` is handled?

jjatie:
ChartViewBase line 483

jjatie:
This doesn't actually solve the bug the original author intended. I feel like highlighting is messy right now, and we should reevaluate it in the future. We probably can just close this PR, but I'll let you decide.

liuxuan30:
ah I see. right a little messy. But why put `lastHighlighted = h` outside.. 
Could we put both of them in `@objc open func highlightValue(_ highlight: Highlight?, callDelegate: Bool)`? just a quick look, not sure if any side effect..

jjatie:
I didn't want to make the change without evaluating the impact everywhere. I'm going to close this, and reevaluate highlighting at a later time.

PR #3179, opened 2018-01-14
Fixed a duplicated assignment compared with obj-c code.

canapio:


canapio:
@jjatie thanks for reply. It’s my mistake. I removed the unnecessary changed file.

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
```diff
@@ -126,7 +126,7 @@ class LineChart1ViewController: DemoBaseViewController {
         set1.valueFont = .systemFont(ofSize: 9)
         set1.formLineDashLengths = [5, 2.5]
         set1.formLineWidth = 1
-        set1.formLineWidth = 15
+        set1.formSize = 15
```
liuxuan30: why set formSize here but not just delete it?

jjatie: The objective-c version sets `formSize`

liuxuan30: oh I don't see it's swift controller.

PR #3199, opened 2018-01-21
Added custom text alignment for noData

jjatie:
Reimplemented #2800


Pranoy1c:
The `open var noDataTextAlignment: NSTextAlignment = .left` is missing the `@objc` so it's not exposed in Objective C.

Issue #2567, opened 2017-06-28
Line Chart do not draw line for Data Set


kbilyk:
I need to update draw line chart to display result of calculations. 

For some reason for data set:
`Charts.LineChartDataSet, 13 entries:
ChartDataEntry, x: 450.479217529297, y 7217.4140625
ChartDataEntry, x: 450.068511962891, y 7090.0
ChartDataEntry, x: 449.577758789062, y 6932.0
ChartDataEntry, x: 449.156005859375, y 6774.0
ChartDataEntry, x: 450.691955566406, y 6616.0
ChartDataEntry, x: 453.031677246094, y 6458.0
ChartDataEntry, x: 455.456634521484, y 6300.0
ChartDataEntry, x: 455.092803955078, y 6142.0
ChartDataEntry, x: 454.658813476562, y 5984.0
ChartDataEntry, x: 454.492492675781, y 5905.0
ChartDataEntry, x: 454.336486816406, y 5668.0
ChartDataEntry, x: 454.281585693359, y 5589.0
ChartDataEntry, x: 454.243194580078, y 5510.0`

Line can be drawn:
![img_0081](https://user-images.githubusercontent.com/1708171/27643552-1052f7cc-5c2a-11e7-8332-adf8ae5ee35a.PNG)

And for data set:
`Charts.LineChartDataSet, 13 entries:
ChartDataEntry, x: 450.359344482422, y 7182.85095214844
ChartDataEntry, x: 450.068511962891, y 7090.0
ChartDataEntry, x: 449.577758789062, y 6932.0
ChartDataEntry, x: 449.156005859375, y 6774.0
ChartDataEntry, x: 450.691955566406, y 6616.0
ChartDataEntry, x: 453.031677246094, y 6458.0
ChartDataEntry, x: 455.456634521484, y 6300.0
ChartDataEntry, x: 455.092803955078, y 6142.0
ChartDataEntry, x: 454.658813476562, y 5984.0
ChartDataEntry, x: 454.492492675781, y 5905.0
ChartDataEntry, x: 454.336486816406, y 5668.0
ChartDataEntry, x: 454.281585693359, y 5589.0
ChartDataEntry, x: 454.243194580078, y 5510.0`

It is no longer drawing:
![img_0082](https://user-images.githubusercontent.com/1708171/27643564-176205ee-5c2a-11e7-9564-6c23b51b7a37.PNG)

liuxuan30:
[...]

PR #3201, opened 2018-01-21
Fix lines not displaying when x-coordinates are unordered

barriserloth:
This commit fixes an issue brought up in #2567
(https://github.com/danielgindi/Charts/issues/2567), where
lines (and points) in LineCharts would not render correctly if the
x-values were not in order.

Let me know if this duplicates something that is already open, or if this is not a fix that users are interested in.

### DIFF
--- a/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
+++ b/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
@@ -73,17 +73,8 @@ open class BarLineScatterCandleBubbleRenderer: DataRenderer
                       dataSet: IBarLineScatterCandleBubbleChartDataSet,
                       animator: Animator?)
         {
-            let phaseX = Swift.max(0.0, Swift.min(1.0, animator?.phaseX ?? 1.0))
-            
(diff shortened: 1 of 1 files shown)


jjatie:
@liuxuan30 I don't understand why we use the old logic that was there, are you able to shed any light on it?

liuxuan30:
The old code seems starts from v2.x. And `lowestVisibleX` and the highest is changed after zooming. 
I doubt this PR could work correctly when chart zoomed? @barriserloth 
when zooming in, the lowest and highest vixible X values are a subset of the data set, so simply using`dataSet.entryCount` shouldn't work.

barriserloth:
I hadn't tested it with zooming in, sorry about that. I can investigate a better fix, but please feel free to close the PR if you'd rather take a different approach to solving this @liuxuan30

jjatie:
@barriserloth Please investigate further if you have time, and reopen this PR if you come up with a better solution.

liuxuan30:
yeah, you are welcome on this, but sorting is not that simple to remove for the moment

hungcaobss:
The line chart display correct now for unordered x-coordinates but can not highlight. Please support me on this

PR #2052, opened 2017-01-10
 #2047 BubbleChartRenderer colors from dataset issue (3.0.1)

chbachmann:
Fix for issue #2047 BubbleChartRenderer colors from dataset issue (3.0.1)

PR #3202, opened 2018-01-22
BubbleChart uses correct colour for index now.

jjatie:
Reimplements #2052


liuxuan30:
looks legit to merge. Can you double check? It makes sense with what I could remember.

PR #1712, opened 2016-10-25
Added Feature to adjust data label position.

hb2708:
Added one variable for offset and calculated radius depending on the given  to achieve the feature referenced in  #1687   Attaching the sample result.
The value is set as follows 
`pieChart.dataLabelOffset = 0.75`
screen shot 2016-10-25 at 11 17 10 am

PR #3214, opened 2018-01-26
Added Feature to adjust data label position.

hb2708:
This PR Contains the same code as the #1712.

Reason behind this is to remove the conflicts which may stop it from merging in the main CodeBase

### DIFF
--- a/Source/Charts/Charts/PieChartView.swift
+++ b/Source/Charts/Charts/PieChartView.swift
@@ -73,7 +73,7 @@ open class PieChartView: PieRadarChartViewBase
     private var _maxAngle: CGFloat = 360.0
     
     /// Extra offset value to be added in the radius to move data label outside the chart
-    fileprivate var _dataLabelOffset: CGFloat = 0.0
+    private var _dataLabelOffset: CGFloat = 0.0
     
     public override init(frame: CGRect)
     {
--- a/Source/Charts/Charts/PieChartView.swift
+++ b/Source/Charts/Charts/PieChartView.swift
@@ -71,7 +71,10 @@ open class PieChartView: PieRadarChartViewBase
     
     /// maximum angle for this pie
     private var _maxAngle: CGFloat = 360.0
-
+    
+    /// Extra offset value to be added in the radius to move data label outside the chart
+    fileprivate var _dataLabelOffset: CGFloat = 0.0
+    
     public override init(frame: CGRect)
     {
         super.init(frame: frame)
@@ -188,7 +191,7 @@ open class PieChartView: PieRadarChartViewBase
         _absoluteAngles = [CGFloat]()
         
         guard let data = _data else { return }
-
+        
         let entryCount = data.entryCount
         
         _drawAngles.reserveCapacity(entryCount)
@@ -197,20 +200,20 @@ open class PieChartView: PieRadarChartViewBase
         let yValueSum = (_data as! PieChartData).yValueSum
         
         var dataSets = data.dataSets
-
+        
         var cnt = 0
-
+        
         for i in 0 ..< data.dataSetCount
         {
             let set = dataSets[i]
             let entryCount = set.entryCount
-
+            
             for j in 0 ..< entryCount
             {
                 guard let e = set.entryForIndex(j) else { continue }
                 
                 _drawAngles.append(calcAngle(value: abs(e.y), yValueSum: yValueSum))
-
+                
                 if cnt == 0
                 {
                     _absoluteAngles.append(_drawAngles[cnt])
@@ -219,7 +222,7 @@ open class PieChartView: PieRadarChartViewBase
                 {
                     _absoluteAngles.append(_absoluteAngles[cnt - 1] + _drawAngles[cnt])
                 }
-
+                
                 cnt += 1
             }
         }
@@ -302,7 +305,7 @@ open class PieChartView: PieRadarChartViewBase
     {
         return _drawAngles
     }
-
+    
     /// - returns: The absolute angles of the different chart slices (where the
     /// slices end)
     @objc open var absoluteAngles: [CGFloat]
@@ -314,7 +317,7 @@ open class PieChartView: PieRadarChartViewBase
     /// 
     /// - note: Use holeTransparent with holeColor = nil to make the hole transparent.*
     @objc open var holeColor: NSUIColor?
-    {
+        {
         get
         {
             return _holeColor
@@ -330,7 +333,7 @@ open class PieChartView: PieRadarChartViewBase
     ///
     /// **default**: `false`
     @objc open var drawSlicesUnderHoleEnabled: Bool
-    {
+        {
         get
         {
             return _drawSlicesUnderHoleEnabled
@@ -350,7 +353,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// `true` if the hole in the center of the pie-chart is set to be visible, `false` ifnot
     @objc open var drawHoleEnabled: Bool
-    {
+        {
         get
         {
             return _drawHoleEnabled
@@ -364,7 +367,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// - returns: `true` if the hole in the center of the pie-chart is set to be visible, `false` ifnot
     @objc open var isDrawHoleEnabled: Bool
-    {
+        {
         get
         {
             return drawHoleEnabled
@@ -373,7 +376,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// the text that is displayed in the center of the pie-chart
     @objc open var centerText: String?
-    {
+        {
         get
         {
             return self.centerAttributedText?.string
@@ -409,7 +412,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// the text that is displayed in the center of the pie-chart
     @objc open var centerAttributedText: NSAttributedString?
-    {
+        {
         get
         {
             return _centerAttributedText
@@ -423,7 +426,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// Sets the offset the center text should have from it's original position in dp. Default x = 0, y = 0
     @objc open var centerTextOffset: CGPoint
-    {
+        {
         get
         {
             return _centerTextOffset
@@ -437,7 +440,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// `true` if drawing the center text is enabled
     @objc open var drawCenterTextEnabled: Bool
-    {
+        {
         get
         {
             return _drawCenterTextEnabled
@@ -451,7 +454,7 @@ open class PieChartView: PieRadarChartViewBase
     
     /// - returns: `true` if drawing the center text is enabled
     @objc open var isDrawCenterTextEnabled: Bool
-    {
+        {
         get
         {
             return drawCenterTextEnabled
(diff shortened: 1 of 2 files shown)


Source/Charts/Charts/PieChartView.swift
```diff
@@ -314,7 +317,7 @@ open class PieChartView: PieRadarChartViewBase
     /// 
     /// - note: Use holeTransparent with holeColor = nil to make the hole transparent.*
     @objc open var holeColor: NSUIColor?
-    {
+        {
```
liuxuan30: code style issue for aligning {

Source/Charts/Charts/PieChartView.swift
```diff
@@ -188,7 +191,7 @@ open class PieChartView: PieRadarChartViewBase
         _absoluteAngles = [CGFloat]()
         
         guard let data = _data else { return }
-
+
```
liuxuan30: maybe we can remove these kinds of diff. @jjatie seems indent with spaces, we should come to agreement, whether use indent for such cases.

Source/Charts/Charts/PieChartView.swift
```diff
@@ -639,4 +642,28 @@ open class PieChartView: PieRadarChartViewBase
...
+        {
+        get
+        {
+            return _dataLabelOffset
+        }
+        set
+        {
+            _dataLabelOffset = newValue
+            
+            if _dataLabelOffset > 1.0
```
liuxuan30: still, we need some justifications for max to 1.0. Why not suport > 1.0?

Source/Charts/Charts/PieChartView.swift
```diff
@@ -639,4 +642,28 @@ open class PieChartView: PieRadarChartViewBase
             }
         }
     }
+    
+    /// The offset value that is used for shifting the data label outside of the center of chart.
+    /// **default**: 0.0
+    open var dataLabelOffset: CGFloat
```
liuxuan30: and we should mention min/max in API documentation

Source/Charts/Charts/PieChartView.swift
hb2708: Max is kept at 1.0 To limit the distance of labels if we allow it to me more than it will go vary far form the visible chart view

Source/Charts/Charts/PieChartView.swift
hb2708: I didn't did anything special here just auto indented all the code

Source/Charts/Charts/PieChartView.swift
hb2708: I didn't did anything special here just auto indented all the code

Source/Charts/Charts/PieChartView.swift
hb2708: The min will be 0.0 and max will be 1.0

Source/Charts/Charts/PieChartView.swift
liuxuan30: ok, you can have different code style on your side, but you have to keep the code style consistent across the library. Here 
image
the `{` has more indent than the old one. Need to fix all code style issues before merge.

Source/Charts/Charts/PieChartView.swift
liuxuan30: @jjatie what you think the multiplier limit?

Source/Charts/Charts/PieChartView.swift
hb2708: ok will fix it

liuxuan30:
@jjatie add your review. I have finished, though some blank spaces for blank lines. If CI passed and you finished review, we can go ahead. 

I checked 3.1 items, almost done.

liuxuan30:
@jjatie waiting for your review. This is the last PR for 3.1

Source/Charts/Charts/PieChartView.swift
hb2708: My Idea behind this was it will be in the ration of the pie chart radius so it will be easy to move them according the chart size for the user.

Source/Charts/Charts/PieChartView.swift
hb2708: ok will fix it

Source/Charts/Charts/PieChartView.swift
```diff
@@ -532,7 +535,7 @@ open class PieChartView: PieRadarChartViewBase
             setNeedsDisplay()
         }
     }
-
```
hb2708: Here there was unnecessary space which i have removed

Source/Charts/Charts/PieChartView.swift
```diff
@@ -294,7 +297,7 @@ open class PieChartView: PieRadarChartViewBase
         
         return -1
     }
-
```
hb2708: Here there was unnecessary space which i have removed

Source/Charts/Charts/PieChartView.swift
```diff
@@ -294,7 +297,7 @@ open class PieChartView: PieRadarChartViewBase
         
         return -1
     }
-    
+
```
hb2708: Here there was unnecessary space which i have removed

Source/Charts/Charts/PieChartView.swift
```diff
@@ -532,7 +535,7 @@ open class PieChartView: PieRadarChartViewBase
             setNeedsDisplay()
         }
     }
-        
+
```
hb2708: Here there was unnecessary space which i have removed

liuxuan30: thanks for this.

Source/Charts/Charts/PieChartView.swift
liuxuan30: @hb2708 I see. However, as a generic feature that goes into the library, we need to make it flexible?
we are having 
```swift
 chart.radius + chart.radius * chart.dataLabelOffset
```
here so it implies the radius will be in range [radius, 2*radius], which would work in your scenario but for other users it becomes kind of restriction.

Changing to a simple absolute offset remove that restriction, and if you want to limite the range, you can just pass 2*radius for the offset.

hb2708: Ok,

Let me try to explain my approach again.
If we do not pass the dataLabelOffset then it will be 0 by default.
So data labels will be there at default position as per original library implementation.

So by using `chart.radius + chart.radius * chart.dataLabelOffset` will move the position of the data labels only SO it will move form the default position to exactly same Distance outside the chart. 

And it will limit the max distance so it will not go too away form the chart so it will be a good UX

hb2708:
@jjatie As we moved this to 3.1 release do we to close the review changes request

hb2708:
@jjatie in  3.1 Release The old PR Is added which is #1712 Which we closed due to conflicts.
so we need to replace #1712 with #3214

hb2708:
@jjatie whats the status now it's approved from you guys?

liuxuan30:
@hb2708 @jjatie is busy with family things. I'm currently holding the PRs and handling.
>Let me try to explain my approach again.
If we do not pass the dataLabelOffset then it will be 0 by default.
So data labels will be there at default position as per original library implementation.
So by using chart.radius + chart.radius * chart.dataLabelOffset will move the position of the data labels only SO it will move form the default position to exactly same Distance outside the chart.
And it will limit the max distance so it will not go too away form the chart so it will be a good UX

Last time we didn't have agreement on the limit and the way we have offset. I understand your point fully, but I'm going to take a decision not to set the restriction. Every dev using the offset should have the sense not to set a large offset to make their UI weird.

From my experience, if you set the max restriction and we merged them, one day there will be people asking to remove it because of some reasons. So it's really unnecessary.

I'd ask you to go with `chart.radius + chart.dataLabelOffset`, or I will make the change when merging this PR (still need some days, as I'm cleaning my notification inbox recently)

hb2708:
@liuxuan30 Okey your suggestions are more practical will do the change and again update it.

Thanks for your valuable time.

liuxuan30:
thank you for your understanding as well. We could review when you have updated.

hb2708:
@liuxuan30 Fixed it the way u suggested. Hope it pass the CI

liuxuan30:
Hi @hb2708 I was testing your PR and look deeper.
There is some code under the radius:
```swift
        // get whole the radius
        let radius = chart.radius
        let rotationAngle = chart.rotationAngle
        var drawAngles = chart.drawAngles
        var absoluteAngles = chart.absoluteAngles
        
        let phaseX = animator.phaseX
        let phaseY = animator.phaseY
        
        var labelRadiusOffset = radius / 10.0 * 3.0
        
        if chart.drawHoleEnabled
        {
            labelRadiusOffset = (radius - (radius * chart.holeRadiusPercent)) / 2.0
        }
        
        let labelRadius = radius - labelRadiusOffset
```
It already has `labelRadiusOffset` and `labelRadius` there. Are these what you are looking for?

I'm sorry I don't see above code before. But will a better solution to change `var labelRadiusOffset = radius / 10.0 * 3.0` to a more flexible one?

Can you check is `labelRadiusOffset` or `labelRadius` is deciding the label position you want to change?

hb2708:
Yes i just saw it it's very confusing.

hb2708:
not sure what to do now

liuxuan30:
I would start with `var labelRadiusOffset = radius / 10.0 * 3.0`, 
make it more flexible like
```swift
var labelRadiusOffset = radius * something
```
while `something` default is 0.3
but the name is decided yet, to keep consistent with your proposal.

liuxuan30:
@hb2708 I'm going to take the first approach based on your commits.

liuxuan30:
@hb2708 @jjatie have look at my new commits; I think this provides compatibility with default behavior and also provide a way that people can change. 
@hb2708 I'm sorry you might want to figure out another formula to generate the offset you need?

liuxuan30:
we also need to add one unit test for pie chart, which is currently missing. So the CI here only means no compile errors.

liuxuan30:
@hb2708 please pull master again, I have added pie chart unit test, let's give it a try.

liuxuan30:
@hb2708 have been heard from you yet.

hb2708:
Just checked your commits
Sorry for delayed response

hb2708:
@liuxuan30 As per the changes you made I can see that you have removed the hardcoded Calculation and allowing user to set the value of `valueLabelOffset`. It looks good.

But when user want to move the labels away form chart the value needs to be decreased from 0.3 to some lower value and to get it totally outside of chart it needs to be go in negative

liuxuan30:
yes, I'm sorry about that, but to keep both consistence and customizable, not many options to have.
Also, please pull master so we could run a pie chart ut.

hb2708:
Yes i did pull

liuxuan30:
Thanks, the piet chart UT works. @jjatie have a look, I think I'm ok

jjatie:
I still don't like that `valueLabelOffset` is a multiplier instead of a constant.

hb2708:
@liuxuan30 thanks for approval.
But @jjatie is not satisfied yet with the solution.

liuxuan30:
@jjatie the old implementation took a multiplier. 
How about we keep the old one, make it a variable and change later in 4.0?

jjatie:
@liuxuan30 I know, and I argued against it. I think we should just drop this for 3.1. If we add it now, and then drastically change it's implementation (multiplier -> constant) later that's not good.

liuxuan30:
removed from sprint. let's figure a better way

PR #3215, opened 2018-01-26
Support other bundle than main MarkerView.viewFromXib()

charlymr:
Hi there;

why this pull request:

If you use the Charts Framework from another Bundle, this method will not find the ressources as this is looking in the main bundle only.

I added a method to be able to load the ressources from a given bundle.

Thanks,

Denis

### DIFF
--- a/Source/Charts/Components/MarkerView.swift
+++ b/Source/Charts/Components/MarkerView.swift
@@ -72,12 +72,8 @@ open class MarkerView: NSUIView, IMarker
     }
     
     @objc
-    open class func viewFromXib() -> MarkerView? {
-        return viewFromXib(in: Bundle.main)
-    }
-    
-    @objc
-    open class func viewFromXib(in bundle: Bundle) -> MarkerView? {
+    open class func viewFromXib(in bundle: Bundle = .main) -> MarkerView?
+    {
         #if !os(OSX)
             return bundle.loadNibNamed(
                 String(describing: self),
--- a/Source/Charts/Components/MarkerView.swift
+++ b/Source/Charts/Components/MarkerView.swift
@@ -77,8 +77,7 @@ open class MarkerView: NSUIView, IMarker
     }
     
     @objc
-    open class func viewFromXib(in bundle: Bundle) -> MarkerView?
-    {
+    open class func viewFromXib(in bundle: Bundle) -> MarkerView? {
         #if !os(OSX)
             return bundle.loadNibNamed(
                 String(describing: self),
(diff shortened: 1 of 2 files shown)


Source/Charts/Components/MarkerView.swift
```diff
@@ -72,10 +72,15 @@ open class MarkerView: NSUIView, IMarker
     }
     
     @objc
-    open class func viewFromXib() -> MarkerView?
+    open class func viewFromXib() -> MarkerView? {
```
jjatie: please put the `{` on a new line.

Source/Charts/Components/MarkerView.swift
```diff
@@ -72,10 +72,15 @@ open class MarkerView: NSUIView, IMarker
...
+        return viewFromXib(in: Bundle.main)
+    }
+    
+    @objc
+    open class func viewFromXib(in bundle: Bundle) -> MarkerView?
```
jjatie: This can be one method: `open class func viewFromXib(in bundle: Bundle = .main) -> MarkerView?` instead of two.

charlymr: Yeah that's what I was thinking in the first place, however if someone is overriding that method it will break it's code. (@objc) is declared as well here. So even more important!

What do you think?

jjatie: The main reason to override this method is to change the bundle. If it breaks their code, this is a very minor change and we will have a few possible minor code breaking changes in 3.1.

Let's make this one method.

jjatie: And keep the `{` on a new line.

charlymr: All done. Thanks. Great job by the way :)

jjatie:
Please update the macOS demo for the new method signature.

jjatie:
@liuxuan30 I'll leave this for you to review. I know you're busy.

Source/Charts/Components/MarkerView.swift
```diff
@@ -72,10 +72,10 @@ open class MarkerView: NSUIView, IMarker
     }
     
     @objc
-    open class func viewFromXib() -> MarkerView?
+    open class func viewFromXib(in bundle: Bundle = .main) -> MarkerView?
```
liuxuan30: I don't realize @objc can be used for functions with default parameters.. :)

liuxuan30:
Seems fine. Though busy, I can afford some time every day in the morning.

Issue #3166, opened 2018-01-09
Realtime LineChart poor performance due to calcMinMax


samueleperricone:
I've found a possible performance bottleneck. In a realtime chart Adding many entries per second to the chart I noticed that automatically the calcMinMax of the dataset is called just after EACH `values.append(e)`

calcMinMax function determine the min and max values using .forEach that is causing the CPU over 60% spent looping the values, moreover in the main thread. 

I'm performing some experiment disabling min/max calculation or using native for loop instead of array .forEach

@danielgindi please have a look, is really calcMinMax on **all** values needed? I calculate min/max Y values manually in my code and only on **visible values**, so maybe you could expose some boolean to enable/disable min/max auto calculation and improving performance removing the forEach function and avoid function calls.

    ```
    open override func calcMinMax()
    {
        guard !values.isEmpty else { return }

        _yMax = -Double.greatestFiniteMagnitude
        _yMin = Double.greatestFiniteMagnitude
        _xMax = -Double.greatestFiniteMagnitude
        _xMin = Double.greatestFiniteMagnitude

        values.forEach { calcMinMax(entry: $0) }
    }
```

samueleperricone:
The DataSet implementation in Android is different.

in swift the didSet is invoked even when we add a new value, instead on Android we have separated addEntry, setValues methods.

when we add a new value Android updates the min/max comparing just the added entry not the whole array. when removing entries the min/max are recalculated on full-array.

Check
https://github.com/PhilJay/MPAndroidChart/blob/master/MPChartLib/src/main/java/com/github/mikephil/charting/data/DataSet.java

jjatie:
We have those methods too. I think we should be treating DataSet as a collection and hide it's backing store (aka `values`). All side affects should be baked into those methods and not be inside the setter for `values`. In the short term this is an easy fix, but we are currently unable to implement this using the standard Swift way because of conflicts with Objective-C compatibility.

I think there is some more we can do to improve performance with live data. I'll try to find my notes on it this week.

codingspark:
Using Instruments you can discover the full calls stack causing the CPU overload. 
I'm adding 30 entries per second and I have 10 minutes of data buffer that means 30 * 60 * 10 = 18000 entries to loop.

replacing the .forEach is a good starting point. 

[...]

PR #3216, opened 2018-01-27
Restored old performance in ChartDataSet

jjatie:
for https://github.com/danielgindi/Charts/issues/3166
Temporary fix for calcMinMax frequency.
Better implementation coming for Charts 4

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -58,11 +58,19 @@ open class ChartDataSet: ChartBaseDataSet
     /// - note: Calls `notifyDataSetChanged()` after setting a new value.
     /// - returns: The array of y-values that this DataSet represents.
     /// the entries that this dataset represents / holds together
-    @objc open var values: [ChartDataEntry] {
-        didSet {
+    @objc open var values: [ChartDataEntry]
+        {
+        didSet
+        {
+            if isIndirectValuesCall {
+                isIndirectValuesCall = false
+                return
+            }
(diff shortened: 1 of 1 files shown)


liuxuan30:
hmm.. what PR is before this invert? I don't remember you particularly have a performance enhance PR？

jjatie:
I don't know which one was the problem. I think it was when I moved to the `didSet`. I think it's important we have a fix for now, I have a good idea on how to improve it in 4.0.

liuxuan30:
Weird I didn't catch this performance drop.

liuxuan30:
wait a minute. I checked it's #3000 
```swift
    @objc open var values: [ChartDataEntry]
    {
        get
        {
            return _values
        }
        set
        {
            _values = newValue
            notifyDataSetChanged()
        }
    }
```
I didn't see `didSet` or `set` has any difference

liuxuan30:
eh, I kind of start to recall. `didSet` will be called for Array type every time an element is added in or removed, is that correct? That's the drop

liuxuan30:
@jjatie do you think it's the right behavior for `didSet` on collection types? I will keep an eye on if swift evolution would address this. Kind of confusing.

jjatie:
Yes, that’s the reason why. didSet is called whenever the underlying data mutates (excluding direct memory access). A computed variable doesn’t have underlying data, so it doesn’t make sense for it to be called on append/remove. 

There’s a couple of major architectural changes I’m going to make over the next couple weeks. After that I’ll work in making performance better for 4.0

liuxuan30:
@jjatie take some rest.

PR #3218, opened 2018-01-27
Add minimum slice angle for value labels to PieChartView

mgod:
Allow hiding labels for small slices in pie charts. Based on #944, also addresses #3182 and #2062, but with a slightly different implementation.

This adds a property `drawSliceTextMinimumAngle` to the pie chart. Any slices smaller than that angle will not be drawn.

Also tweaked a few values in the pie charts demos to make them a little easier to read.

### DIFF
--- a/Source/Charts/Charts/PieChartView.swift
+++ b/Source/Charts/Charts/PieChartView.swift
@@ -72,8 +72,6 @@ open class PieChartView: PieRadarChartViewBase
     /// maximum angle for this pie
     private var _maxAngle: CGFloat = 360.0
 
-    private var _drawSliceTextMinimumAngle: CGFloat = 0.0
-
     public override init(frame: CGRect)
     {
         super.init(frame: frame)
@@ -641,16 +639,10 @@ open class PieChartView: PieRadarChartViewBase
         }
     }
     
-    /// smallest pie slice angle that will have a label drawn
-    @objc open var drawSliceTextMinimumAngle: CGFloat
+    /// smallest pie slice angle that will have a label drawn in degrees, 0 by default
+    @objc open var sliceTextDrawingThreshold: CGFloat = 0.0
     {
-        get
-        {
-            return _drawSliceTextMinimumAngle
-        }
-        set
-        {
-            _drawSliceTextMinimumAngle = newValue
+        didSet {
             setNeedsDisplay()
         }
     }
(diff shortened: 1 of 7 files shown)


Source/Charts/Charts/PieChartView.swift
```diff
@@ -72,6 +72,8 @@ open class PieChartView: PieRadarChartViewBase
     /// maximum angle for this pie
     private var _maxAngle: CGFloat = 360.0
 
+    private var _drawSliceTextMinimumAngle: CGFloat = 0.0
```
jjatie: This should be one `open` variable with a `didSet`. The comment should indicate whether this is degrees or radians. The name is weird; it sounds like an angle value is going to be drawn. Perhaps something like `minimumAngleToDrawSliceText` or `sliceTextDrawingThreshold`. I'm open to suggestions.

Source/Charts/Renderers/PieChartRenderer.swift
```diff
@@ -358,10 +359,10 @@ open class PieChartRenderer: NSObject, DataRenderer
                 let sliceXBase = cos(transformedAngle.DEG2RAD)
                 let sliceYBase = sin(transformedAngle.DEG2RAD)
                 
-                let drawXOutside = drawEntryLabels && xValuePosition == .outsideSlice
-                let drawYOutside = drawValues && yValuePosition == .outsideSlice
-                let drawXInside = drawEntryLabels && xValuePosition == .insideSlice
-                let drawYInside = drawValues && yValuePosition == .insideSlice
+                let drawXOutside = sliceAngle > drawSliceTextMinimumAngle && drawEntryLabels && xValuePosition == .outsideSlice
```
jjatie: In an effort to make this a little cleaner, do 
```
guard sliceAngle > drawSliceTextMinimumAngle else { continue }
```

and then you don't need to add it to each line.

mgod: The way the core loop is written, `continue` breaks the `xIndex` handling later in the loop. Looking over the code, I can't tell if the `xIndex` actually works correctly in cases where there are multiple data sets, but I don't understand it well enough to tweak it. This loop feels like it's complicated enough that I don't want to do more than minimum necessary changes.

I'm also not sure if we want to disable drawing the icons, which this change would do as well.

Happy to do something like:

```
var drawXOutside = drawEntryLabels && xValuePosition == .outsideSlice
var drawYOutside = drawValues && yValuePosition == .outsideSlice
var drawXInside = drawEntryLabels && xValuePosition == .insideSlice
var drawYInside = drawValues && yValuePosition == .insideSlice
if sliceAngle <= sliceTextDrawingThreshold {
    drawXOutside = false
    drawYOutside = false
    drawXInside = false
    drawYInside = false
}
```

petester42:
Thanks rebased and merged

Issue #3251, opened 2018-02-08
It is not possible to check if chart is being dragged, scaled or decelerated


kzyryanov:
In some cases it is good to have possibility to check if chart is being dragged, scaled or decelerated (like for UIScrollView)

liuxuan30:
May I know what scenario it's helpful to know?
inside the library there are some private vars for them:
```swift
    private var _isDragging = false
    private var _isScaling = false
```
You can change the source code to make it public or add some vars/extensions if possible.

kzyryanov:
@liuxuan30 
It's not a really often case, but if you want to make some complicated scroll (like scroll several chart views or chart view + some scroll view) you need to know the state (like UIScrollView has these readonly variables)

ninjazoete:
@kzyryanov 
This is exactly what I need. I need to know if chart was scrolled and if it was the end of content size. Did you come up with idea how to get to this?

kzyryanov:
@ninjazoete Check my pull request https://github.com/danielgindi/Charts/pull/3252 🙃

PR #3252, opened 2018-02-08
Add dragging, scaling and decelerating variables to BarLineChartViewBase

kzyryanov:
In some cases it is good to have possibility to check if chart is being dragged, scaled or decelerated (like for UIScrollView)
#3251

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -34,6 +34,18 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     private var _scaleXEnabled = true
     private var _scaleYEnabled = true
     
+    @objc open var isDragging: Bool {
+        return _isDragging
+    }
+    
+    @objc open var isDecelerating: Bool {
+        return _decelerationDisplayLink != nil
+    }
+    
+    @objc open var isScaling: Bool {
+        return _isScaling
+    }
+    
     /// the color for the background of the chart-drawing area (everything behind the grid lines).
     @objc open var gridBackgroundColor = NSUIColor(red: 240/255.0, green: 240/255.0, blue: 240/255.0, alpha: 1.0)
     


liuxuan30:
@jjatie what you think about making them public directly?
It's the first time I see someone want to know the state.

kzyryanov:
@liuxuan30 @jjatie 
It's not a really often case, but if you want to make some complicated scroll (like scroll several chart views or chart view + some scroll view) you need to know the state (like UIScrollView has these readonly variables)

liuxuan30:
@kzyryanov this PR looks most likely will be delay to 4.0 release, which contains our swift refactoring, leading by @jjatie. I found your code style maybe is not aligned with the new swift code style in the library. May I ask you to switch the branch to 4.0 and later when @jjatie came back, he could offer more insights and get this merged.

jjatie:
As @liuxuan30 said, there is no reason to add new variables, just change the access control on the existing ones (and remove the `_` prefix). If we do accept this it will be in 4.x. I think the feature is something we should provide, but this is not how to do it.

Since you mentioned `UIScrollView`, I think it would make more sense for us to expose this information by providing a proper delegate similar to `UIScrollViewDelegate`. Now that I'm thinking about it, @liuxuan30 do you know why all of the charts aren't already subclasses of `UIScrollView`?

liuxuan30:
@jjatie I think it should not be UIScrollView after all, as the zooming and panning are done by CoreGraphics directly, involving UIScrollView seems unnecessary.
Implementing a UIScrollChartView would be an interesting topic, but what I guess the UIScrollView's nature might cause more troubles than leveraging the scroll?

sunnypurewal:
It should be noted that there is no private variable for isDecelerating so the only way to return that flag is to check if the display link is not null, just like in this PR.

jjatie:
This functionality needs to be implemented for ScatterCharts, `CandleChart`, `BubbleChart` and `CombinedChart`. Exposing the information through a property means the only way to respond to it is through KVO, which is undesirable. This should be a delegate callback.

Issue #3255, opened 2018-02-09
About `isDrawingValuesAllowed` function in `ChartDataRendererBase`:


Renford:
```
let arr1 = Array(1 ..< 100).map { x in return sin(Double(x) / 2.0 / 3.141 ) * 100.0 }
let ds1 = LineChartDataSet(values: arr1, label: PriceLineLabelKey)
ds1.drawValuesEnabled = false;

let arr2 = Array(1 ..< 100).map { x in return sin(Double(x) / 2.0 / 3.141 * 2) * 100.0 }
let ds2 = LineChartDataSet(values: arr2, label: PriceLineLabelKey)
ds1.drawValuesEnabled = false;

let arr3 = Array(1 ..< 10).map { x in return sin(Double(x) / 2.0 / 3.141 * 3) * 100.0 }
let ds3 = ScatterChartDataSet(values: arr3, label: PriceLineLabelKey)
ds1.drawValuesEnabled = true;

let lineData = LineChartData()
lineData.addDataSet(ds1)
lineData.addDataSet(ds2)

let scatterData = ScatterChartData()
scatterData.addDataSet(ds3)

let data = CombinedChartData()
data.lineData = lineData
data.scatterData = scatterData

chartView.data = data
chartView.notifyDataSetChanged()
```

when ds1、ds2 drawValuesEnabled == false, ds3 drawValuesEnabled == true, `data.entryCount` == 210。but `data.entryCount` should be 10。

```
open func isDrawingValuesAllowed(dataProvider: ChartDataProvider?) -> Bool
    {
        guard let data = dataProvider?.data
            else { return false }
        
        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * (viewPortHandler?.scaleX ?? 1.0))
    }
```

liuxuan30:
why this is an issue, if the data.entryCount is not "correct" in your terms?
the entry count reflects how many entries are there, but it does not mean it should be drawn on the screen, they are not a conflict.

Renford:
I want to show a part of dataset In `CombinedChartView`, and hide another part. For example, show values of ds3 and hide value of ds1,ds2. 

But, `ds1.entryCount` or `ds2.entryCount` too many, so `ds3.drawValuesEnabled = true` is invalid. See to line 115 in `ScatterChartRenderer` and line 56 in `DataRenderer`.

jjatie:
[...]

PR #3256, opened 2018-02-09
fix bug about drawValuesEnabled

Renford:
look at issues 3255 
https://github.com/danielgindi/Charts/issues/3255

### DIFF
--- a/Source/Charts/Renderers/ChartDataRendererBase.swift
+++ b/Source/Charts/Renderers/ChartDataRendererBase.swift
@@ -50,10 +50,19 @@ open class DataRenderer: Renderer
     /// An opportunity for initializing internal buffers used for rendering with a new size.
     /// Since this might do memory allocations, it should only be called if necessary.
     @objc open func initBuffers() { }
-    
+
     @objc open func isDrawingValuesAllowed(dataProvider: ChartDataProvider?) -> Bool
     {
-        guard let data = dataProvider?.data else { return false }
-        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * viewPortHandler.scaleX)
+        guard let data = dataProvider?.data
+            else { return false }
+
+        var showEntryCount = 0
+        for dataset in data.dataSets {
+            if dataset.drawValuesEnabled == true {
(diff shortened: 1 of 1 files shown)


Source/Charts/Renderers/ChartDataRendererBase.swift
```diff
@@ -50,10 +50,19 @@ open class DataRenderer: Renderer
...
     /// Since this might do memory allocations, it should only be called if necessary.
     @objc open func initBuffers() { }
-    
+
     @objc open func isDrawingValuesAllowed(dataProvider: ChartDataProvider?) -> Bool
     {
-        guard let data = dataProvider?.data else { return false }
-        return data.entryCount < Int(CGFloat(dataProvider?.maxVisibleCount ?? 0) * viewPortHandler.scaleX)
+        guard let data = dataProvider?.data
+            else { return false }
```
jjatie: Please revert to the previous style (one line guard statement)

Source/Charts/Renderers/ChartDataRendererBase.swift
```diff
@@ -50,10 +50,19 @@ open class DataRenderer: Renderer
...
+
+        var showEntryCount = 0
+        for dataset in data.dataSets {
+            if dataset.drawValuesEnabled == true {
+                showEntryCount += dataset.entryCount
+            }
+        }
```
jjatie:
```
let showEntryCount = data.dataSets.reduce(into: 0) { result, set in
    if set.drawValuesEnabled
    {
        result += set.entryCount
    }
}
```

jjatie:
I'll think about this one. I understand where you're coming from in #3255, but as @liuxuan30 pointed out this isn't a bug, it's a matter of preference. My gut tells me that this is the better default to have, but then the question becomes: what happens when ds1 (from #3255) is told to draw? Do we automatically draw so that all values can be seen or only leave 10?

jjatie:
I've looked through issues and this doesn't seem to have come up before. I am closing this because it seems to be an issue that stems from a misalignment between your mental model and the implementation. If you can come up with a better argument for it I'm happy to reconsider.

PR #3265, opened 2018-02-13
Min and Max reset when clearing ChartDataSet (Fixes #3260)

### LINKED ISSUE #3260: ChartDataSet min and max values not recalculated when calling clear()
The method `clear()` of `ChartDataSet` is described as:

> Removes all values from this DataSet and recalculates min and max value.

but the min and max values are never recalculated (or rather, reset).
Calling `clear()` _does_ trigger `calcMinMax()`, but the variables `_yMax`, `_yMin`, `_xMax`, `_xMin` are never reset to their initial values because of a guard statement:
```
open override func calcMinMax()
{
    guard !values.isEmpty else { return }

    _yMax = -Double.greatestFiniteMagnitude
    _yMin = Double.greatestFiniteMagnitude
    _xMax = -Double.greatestFiniteMagnitude
    _xMin = Double.greatestFiniteMagnitude

    values.forEach { calcMinMax(entry: $0) }
}
```
The  guard statement should come after the min max variables are reset (or even removed completely, since it wouldn't make any difference).

carlo-:
Fixes issue #3260 
The min and max values of `ChartDataSet` should always be reset to their default values when `clear()` is called. Since `clear()` removes all items from `values`, the guard statement prevents this reset from happening. Moving the guard statement won't have any impact on the rest of the project.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -86,22 +86,22 @@ open class ChartDataSet: ChartBaseDataSet
     
     open override func calcMinMax()
     {
-        guard !values.isEmpty else { return }
-
         _yMax = -Double.greatestFiniteMagnitude
         _yMin = Double.greatestFiniteMagnitude
         _xMax = -Double.greatestFiniteMagnitude
         _xMin = Double.greatestFiniteMagnitude
+
+        guard !values.isEmpty else { return }
 
         values.forEach { calcMinMax(entry: $0) }
     }
     
     open override func calcMinMaxY(fromX: Double, toX: Double)
     {
-        guard !values.isEmpty else { return }
-
         _yMax = -Double.greatestFiniteMagnitude
         _yMin = Double.greatestFiniteMagnitude
+
+        guard !values.isEmpty else { return }
         
         let indexFrom = entryIndex(x: fromX, closestToY: Double.nan, rounding: .down)
         let indexTo = entryIndex(x: toX, closestToY: Double.nan, rounding: .up)


liuxuan30:
Thanks, I will take a look at this PR and should be accepted in 3.1 release, after I clean out my notification box.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -86,22 +86,22 @@ open class ChartDataSet: ChartBaseDataSet
...
     }
     
     open override func calcMinMaxY(fromX: Double, toX: Double)
     {
-        guard !values.isEmpty else { return }
-
         _yMax = -Double.greatestFiniteMagnitude
         _yMin = Double.greatestFiniteMagnitude
+
+        guard !values.isEmpty else { return }
```
liuxuan30: The change here seems legit. But how you find `calcMinMaxY` needs fix as well? I don't see `clear()` will eventually call `calcMinMaxY`?
I just want to make sure we don't missed other similar issues. Can you help do a double check?

carlo-: You're right, `calcMinMaxY` isn't triggered when calling `clear()`, and I haven't had any trouble from its current implementation, but I thought it would still be better to reset `_yMax` and `_yMin` to their defaults rather than leaving them with random numbers from the previous set. Feel free to edit my changes if you'd rather leave that alone.

liuxuan30: alright, got you. seems fine to me.

jjatie: It's weird that this issue is only coming up now, as the implementation hasn't changed in a very long time. @carlo I think you're right and that the guard statement should be removed entirely.

liuxuan30:
@jjatie I tested a few seems `inf` won't trigger crash though `singular matrix` or weird axis rendering would happen, but it's enough signs. Have your review. I'm ok.

PR #3290, opened 2018-02-22
Implemented a simple animation for highlighting

kalmurzayev:
Hi!
I implemented an animated highlighting for `PieChartView`.
- I extended IPieChartDataSet protocol to include `isSelectionAnimated` and `selectionShiftDuration`
- I added these properties in PieChartDataSet by setting their default values to `false` and `0.5` by default
- overriden `func highlightValue(_ highlight: Highlight?, callDelegate: Bool)` to include checking `isSelectionAnimated` and animating slice if needed
- added unit tests checking if animation occurs only when `isSelectionAnimated` is enabled

Will be happy to hear your feedback and make necessary changes


liuxuan30:
Thanks! It's a big feature, so it might take longer to start review, as we are busy with releasing 3.1 and plan for 4.0, and swift refactoring.

I'm thinking this feature would be nice to have in 4.0 or future. Our 4.0 branch now is focusing on swift refactoring, so I guess you have to switch to 4.0 branch and follow the new code style there.

@jjatie may provide more help regarding the new swift trend and  code style.

kalmurzayev:
Great. So should I probably rebase my branch onto branch 4.0?

liuxuan30:
Yes you could. if you unsure the new swift code style, just keep old ones, @jjatie could help review

liuxuan30:
please switch this PR's target to 4.0 branch as well

jjatie:
@kalmurzayev Why was the changelog updated with so many items?

Source/Charts/Animation/Animator.swift
```diff
@@ -29,15 +29,65 @@ public protocol AnimatorDelegate
 @objc(ChartAnimator)
 open class Animator: NSObject
 {
+    public enum Dimension: Int {
+        case X, Y, H
```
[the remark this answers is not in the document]
jjatie: These should be lowercase. Why are they backed by an `Int`?

Source/Charts/Animation/Animator.swift
```diff
@@ -29,15 +29,65 @@ public protocol AnimatorDelegate
 @objc(ChartAnimator)
 open class Animator: NSObject
```
jjatie: If these changes only apply to the PieChart, then you should subclass `Animator` to be `PieChartAnimator`. If some things still belong in the base class, of course keep them here.

Source/Charts/Animation/Animator.swift
```diff
@@ -29,15 +29,65 @@ public protocol AnimatorDelegate
...
+        var phase: Double = 1.0
+        var duration: TimeInterval = 0.0
+        var startTime: TimeInterval = 0.0
+        var endTime: TimeInterval = 0.0
+        var enabled: Bool = false
+        var easing: ChartEasingFunctionBlock?
+        
+        mutating func updatePhase(currentTime: TimeInterval) {
+            var elapsed = currentTime - startTime
+            if elapsed > duration
```
jjatie: `elapsed = min(duration, elapsed)`

Source/Charts/Animation/Animator.swift
```diff
@@ -29,15 +29,65 @@ public protocol AnimatorDelegate
...
+    }
+    private var animatedDimensions: [Dimension: State] = [:]
+    
     @objc open weak var delegate: AnimatorDelegate?
     @objc open var updateBlock: (() -> Void)?
     @objc open var stopBlock: (() -> Void)?
     
     /// the phase that is animated and influences the drawn values on the x-axis
-    @objc open var phaseX: Double = 1.0
+    @objc open var phaseX: Double {
```
jjatie: The API you're replacing allowed you to set `phaseX`. It's not obvious what the replacement is. Does it make sense to declare three properties like `var xDimension: State` or something of the like? What are your thoughts?

You're already making a fairly large change to the API, so I'm okay making some more. Let's make sure that it makes the most sense before we merge it. I've been wanting to do something the like `State` struct in a lot of different areas of the framework.

Source/Charts/Animation/Animator.swift
```diff
@@ -29,15 +29,65 @@ public protocol AnimatorDelegate
...
+        }
+    }
+    fileprivate var hasUnfinishedAnimations: Bool {
+        return animatedDimensions.reduce(false) { (has, animation) in
+            return has || animation.1.phase != 1.0
+        }
+    }
+    
+    fileprivate var hasEnabledAnimations: Bool {
+        return animatedDimensions.reduce(true) { (has, animation) in
```
jjatie: I think you meant to start with `false` because `(true || x) == true` always.

Source/Charts/Animation/Animator.swift
```diff
@@ -69,53 +118,56 @@ open class Animator: NSObject
...
-
+        
         _displayLink?.remove(from: .main, forMode: .commonModes)
         _displayLink = nil
-
+        
         _enabledX = false
         _enabledY = false
-
+        disableAllAnimations();
```
jjatie: no semicolon

Source/Charts/Animation/Animator.swift
```diff
@@ -69,53 +118,56 @@ open class Animator: NSObject
...
-            if elapsed > duration
+    }
+    
+    fileprivate func finishAllAnimations() {
+        for (dim, _) in animatedDimensions {
+            animatedDimensions[dim]?.enabled = false
+            animatedDimensions[dim]?.phase = 1.0
+        }
+    }
+
```
jjatie: extra whitespace

Source/Charts/Animation/Animator.swift
```diff
@@ -144,29 +197,39 @@ open class Animator: NSObject
...
+                _displayLink?.add(to: RunLoop.main, forMode: .commonModes)
+            }
         }
     }
     
+    open func animate(dimension: Dimension, duration: TimeInterval, easingOption: ChartEasingOption) {
+        animate(dimension: dimension, duration: duration, easing: easingFunctionFromOption(easingOption))
+    }
+    
+    open func animate(dimension: Dimension, duration: TimeInterval) {
```
jjatie: this can be replaced by a default value on the previous method

Source/Charts/Charts/PieChartView.swift
```diff
@@ -264,6 +264,21 @@ open class PieChartView: PieRadarChartViewBase
...
+    @objc open func animateHighlight(duration: TimeInterval) {
+        chartAnimator.animate(dimension: .H, duration: duration)
+    }
+    
+    @objc override open func highlightValue(_ highlight: Highlight?, callDelegate: Bool) {
+        super.highlightValue(highlight, callDelegate: callDelegate);
+        guard let pieData = self.data as? PieChartData,
+            let duration = pieData.dataSet?.selectionShiftDuration,
+            let isEnabled = pieData.dataSet?.isSelectionAnimated,
+            isEnabled else { return }
```
jjatie:
```
/// ...
isEnabled
else { return }
```

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -243,7 +243,7 @@ open class PieRadarChartViewBase: ChartViewBase
 
         viewPortHandler.restrainViewPort(offsetLeft: offsetLeft, offsetTop: offsetTop, offsetRight: offsetRight, offsetBottom: offsetBottom)
     }
-
+
```
jjatie: You're adding in whitespace. Make sure you have the two boxes checked in Xcode Text Editing > White Editing prefs to avoid this.

VatanaChhorn:
@kalmurzayev any updates?

Issue #620, opened 2015-12-21
StackedBar value label overlapped


luomuwen:
i use stackedBar to show two data, but when one value is zero , it seems  two label is overlapped.
![e998a5b0-5e3f-49d6-800c-464d95934ef9](https://cloud.githubusercontent.com/assets/1125682/11922984/a9e691c2-a7d7-11e5-9b10-bd0362e44aaf.png)

liuxuan30:
your chart seems weird, could you provide the sample code using ChartsDemo code?

I can see many 0 on left, and two 0 on right, and overlap with 1?

luomuwen:
This is my code
```Objectivec
    NSMutableArray *xVals = [[NSMutableArray alloc] init];
    for (int i = 0; i < 12; i++){
        [xVals addObject:self.months[i]];
    }
    
    NSMutableArray *yVals = [[NSMutableArray alloc] init];
    if(self.statisticsModel.monthStatisticsArray){
        for (int i = 0; i < [self.statisticsModel.monthStatisticsArray count]; i++){
            CVMonthStatisticsModel* monthModel = self.statisticsModel.monthStatisticsArray[i];
            
            CVLog(@"success:%d, fail:%d", monthModel.success, monthModel.fail);
            [yVals addObject:[[BarChartDataEntry alloc] initWithValues:@[@(monthModel.success), @(monthModel.fail)] xIndex:i]];
        }
    }
    
    BarChartDataSet *set1 = [[BarChartDataSet alloc] initWithYVals:yVals label:@""];
    set1.barSpace = 0.3;
    set1.colors = @[UIColorFromRGB(0x00b26d), UIColorFromRGB(0xfac955)];
    set1.stackLabels = @[@"成功", @"失败"];
    set1.valueTextColor = UIColorFromRGB(0x2CC9CA);
    
    NSNumberFormatter* formatter = [[NSNumberFormatter alloc] init];
    formatter.numberStyle =  kCFNumberFormatterDecimalStyle;
    
    set1.valueFormatter = formatter;
    
    NSMutableArray *dataSets = [[NSMutableArray alloc] init];
    [dataSets addObject:set1];
    
    BarChartData *data = [[BarChartData alloc] initWithXVals:xVals dataSets:dataSets];
    [data setValueFont:[UIFont fontWithName:@"HelveticaNeue-Light" size:7.f]];
    self.statisticView.data = data;
```
this is normal 
[...]

PR #3316, opened 2018-03-04
StackedBar value label overlapped #620

jonasfugedi:
Suggested changes to improve label rendering for stacked bar graphs where labels may overlap or not fit within the bar segment. Also does not show label for zero size stack.


liuxuan30:
could you add images before/after the fix so we get a rough idea?

chetanpanchal94:
I tried this solution, but still values in stack bar chart still overlaps

Issue #3327, opened 2018-03-11
ChartsDemo-Swift crashes


valeriyvan:
ChartsDemo-Swift crashes when in Line Chart in option choose Toggle Data option and then choose, Toggle Cubic option.

liuxuan30:
That's because you niled out the chartData and then try to unwrap it:
```swift
        case .toggleCubic:
            for set in chartView.data!.dataSets as! [LineChartDataSet] {
                set.mode = (set.mode == .cubicBezier) ? .linear : .cubicBezier
            }
```
@jjatie this is no big deal. But I checked many of the view controllers uses `chartView.data!.dataSets`,  do you have any smart fix for this? Or we have to fix it one by one

jjatie:
Looks like we'll need to guard each `handleOption` implementation.

jjatie:
@liuxuan30 @valeriyvan I've thought about this some more. It's easy to imagine a case where a user wants to temporarily hide a given `ChartData`. It seems to me that `ChartData` should have an `isEnabled` property or something. If `isEnabled == false` then it's data sets are not drawn.z

Once this implemented, the demos should not be setting the data to nil, but simply toggling `chartData.isEnabled`

valeriyvan:
Does it sound like data should’t be optional anymore?

jjatie:
That's a different discussion, but I've been thinking yes.

jjatie:
This seems like a corner case and optionality of many properties are changing in 4.0. As we have limited resources to keep making changes for non-critical areas, I am closing this. After 4.0 is released, we will have more time to reevaluate issues like this.

PR #3329, opened 2018-03-12
Refactors optionTapped func, adds guard to avoid crash.

valeriyvan:
Solves #3327.
As for me it doesn't solve origin of a problem.
For charts with the only dataset option to toggle dataset should't be shown at all.

### DIFF
--- a/ChartsDemo/Swift/Demos/CombinedChartViewController.swift
+++ b/ChartsDemo/Swift/Demos/CombinedChartViewController.swift
@@ -92,28 +92,27 @@ class CombinedChartViewController: DemoBaseViewController {
     }
     
     override func optionTapped(_ option: Option) {
+        guard let data = chartView.data else
+        {
+            super.handleOption(option, forChartView: chartView)
+            return
+        }
+
         switch option {
         case .toggleLineValues:
-            for set in chartView.data!.dataSets {
-                if let set = set as? LineChartDataSet {
-                    set.drawValuesEnabled = !set .drawValuesEnabled
-                    
-                }
-            }
+            data.dataSets.filter { $0 is LineChartDataSet }
+                         .forEach { $0.drawValuesEnabled = !$0.drawValuesEnabled }
             chartView.setNeedsDisplay()
             
         case .toggleBarValues:
-            for set in chartView.data!.dataSets {
-                if let set = set as? BarChartDataSet {
-                    set.drawValuesEnabled = !set .drawValuesEnabled
-                }
-            }
+            data.dataSets.filter { $0 is BarChartDataSet }
+                         .forEach { $0.drawValuesEnabled = !$0.drawValuesEnabled }
             chartView.setNeedsDisplay()
             
         case .removeDataSet:
-            let rnd = Int(arc4random_uniform(UInt32(chartView.data!.dataSetCount)))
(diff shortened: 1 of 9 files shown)


ChartsDemo/Swift/Demos/LineChart1ViewController.swift
```diff
@@ -142,35 +142,30 @@ class LineChart1ViewController: DemoBaseViewController {
     }
     
     override func optionTapped(_ option: Option) {
+        guard let sets = chartView.data?.dataSets as? [LineChartDataSet] else {
```
jjatie: "{" on a new line please.

ChartsDemo/Swift/Demos/LineChart1ViewController.swift
```diff
@@ -142,35 +142,30 @@ class LineChart1ViewController: DemoBaseViewController {
...
+            super.handleOption(option, forChartView: chartView)
+            return
+        }
+
         switch option {
         case .toggleFilled:
-            for set in chartView.data!.dataSets as! [LineChartDataSet] {
-                set.drawFilledEnabled = !set.drawFilledEnabled
-            }
+            sets.forEach{ $0.drawFilledEnabled = !$0.drawFilledEnabled }
```
jjatie: a space between `forEach` and `{`

ChartsDemo/Swift/Demos/CombinedChartViewController.swift
```diff
@@ -92,28 +92,27 @@ class CombinedChartViewController: DemoBaseViewController {
...
         switch option {
         case .toggleLineValues:
-            for set in chartView.data!.dataSets {
-                if let set = set as? LineChartDataSet {
-                    set.drawValuesEnabled = !set .drawValuesEnabled
-                    
-                }
-            }
+            data.dataSets.filter { $0 is LineChartDataSet }
+                         .forEach { $0.drawValuesEnabled = !$0.drawValuesEnabled }
```
jjatie: This is fine here, but if you plan on contributing to the Charts framework avoid this. Here we have to loop through `dataSets` twice to enable toggle `drawValuesEnabled`, whereas something like :
```
for case let set as LineChartDataSet in data.dataSets {
    set.drawValuesEnabled = !set.drawValuesEnabled
}
```
only has to loop once.
I am curious what the performance characteristics of:
```
data.dataSets.lazy
    .flter { ... }
    .forEach { ... }
```
are like.

valeriyvan:
done

ChartsDemo/Swift/Demos/CombinedChartViewController.swift
valeriyvan: I don't know an answer for your last question. 
Definitely, `sequence.filter{ $0.condition }.forEach{ $0.doSomething }` is slower than `sequence.forEach{ if $0.condition { $0.doSomething } }`. One is faster, another is readable. I guess, here, where we loop over data sets, minor performance loss could be sacrificed for readability.

valeriyvan:
done done

jjatie:
@valeriyvan take a look at my comment on #3327

valeriyvan:
After putting so much effort into fixing bug the way you suggested, you change your mind(

jjatie:
I know it sucks (it happens with my own PRs a lot). Sometimes you just need to see the outcome of a decision before you see it's wrong.

PR #3330, opened 2018-03-12
Uses guard instead of if

valeriyvan:


ChartsDemo/Swift/Demos/LineChart1ViewController.swift
```diff
@@ -99,7 +99,7 @@ class LineChart1ViewController: DemoBaseViewController {
     }
 
     override func updateChartData() {
-        if self.shouldHideData {
+        guard !self.shouldHideData else {
```
jjatie: if makes sense here, but should be 
```
if shouldHideData {
    // Hide the data...
} else {
    // Show the data
}
```

jjatie:
@valeriyvan take a look at my comment on #3327

PR #3355, opened 2018-03-21
Fix crash when dataSet for adding is nil

eshumilov:


liuxuan30:
What is the crash log?
```swift
    @objc open func addDataSet(_ dataSet: IChartDataSet!)
    {
        calcMinMax(dataSet: dataSet)
        
        _dataSets.append(dataSet)
    }
```
the parameter `dataSet: IChartDataSet!` won't take a nil in the first place, why bother adding a useless check?

eshumilov:
Sorry, I'm not sure that I understand what do you mean by "won't take a nil in the first place".
Using this declaration allows to send `nil` as argument.

To reproduce just place next line of code and execute it:
```ChartData().addDataSet(nil)```

screen shot 2018-03-22 at 11 28 11 am

This fix is similar how it was done in the next func `removeDataSet`.

I think that more appropriate way to solve this issue is remove exclamation mark (!) from such functions, but it will broke the current api as optional values could not be passed.

liuxuan30:
yes, the Fatal error in your screenshot already explained: You passed a nil to a function who doesn't take nil technically, because it will be unwrapped without checking nil first. it might a little confusing here: if it does not take nil, why uses `!` at beginning, which I don't have a good answer yet. 

You have to avoid passing nil. I agree we could remove `!` here.

@jjatie do we addressed this in our 4.0 branch?

liuxuan30:
Closing, as this is the not right way to fix it and 4.0 branch already take a different way.

jjatie:
@liuxuan30 `removeDataSet` will be updated. My mistake

Issue #2963, opened 2017-11-06
Line chart lines are clipped


bdorfman:
If you, for example, on a LineChartView set `leftAxis.axisMinimum` to 0, then the line itself will be clipped at 0 instead of being full width. Setting `clipValuesToContentEnabled` to true doesn't fix this, nor does altering the `minOffset` of the view. And there is no easy way that I can think of to offset axisMinimum to compensate for lineWidth (since the width is in screen points which will not translate neatly to an axis value).

Example screenshot of this behavior:
<img width="203" alt="simulator006" src="https://user-images.githubusercontent.com/2694018/32467727-84fd1b88-c300-11e7-8b7b-a31a47ce8439.png">

liuxuan30:
that's a little tricky, as you specifically have line segment from 0 to 0, while you set clip on 0 line. That's done by CoreGraphics. One workaround is you modify `contentRect.height` to include half line width. 
`clipValuesToContentEnabled` is for the value labels clipping, so it will not work in your case for sure.

wtmoose:
>One workaround is you modify contentRect.height to include half line width.

I don't see how this is a workaround. The clipping area needs to be larger than the drawing area, but both are defined by the `contentRect`. So in this case, increasing `contentRect` draws the path larger and it is still clipped.

It seems that the solution would require having the following line of code apply an additional offset to `contentRect`:

````swift
context.clip(to: _viewPortHandler.contentRect)
````

Am I missing something?

bdorfman:
Yes, the suggested contectRect clipping doesn't work. I believe the solution I came up with was to do a somewhat complicated conversion between screen points/pixels and the graphs coordinate space and then shift the axis miminum and maximums so that they wouldn't clip. It's annoying and you have to recalculate whenever the pixel size of the graph changes. I still think this is a bug and should be handled for you by the library.

wtmoose:
@bdorfman I can see why it works the way it does: clipping is needed for the case where values fall outside of the viewport. If the clipping region were extended, a line that crosses the viewport would get clipped outside of the axes. But for data sets that are wholly within the viewport, it would be nice to be able to disable the clipping entirely.

@danielgindi would a PR that adds an option to disable viewport clipping have a chance of being accepted?

liuxuan30:
my apology if workaround not work. But the idea is similar, you have to give more space for the lines to avoid clipping.
For PR - Yes we definitely accept any meaningful PR and you are more than welcome.

PR #3360, opened 2018-03-22
Add option to disable clipping data to contentRect

wtmoose:
### Issue Link :link:
#2963 

### Goals :soccer:
Support cases where the clipping of data produces unwanted visual issues by adding an option to disable it.

screen shot 2018-03-22 at 9 13 54 am

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -50,6 +50,11 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
     /// When enabled, the values will be clipped to contentRect, otherwise they can bleed outside the content rect.
     @objc open var clipValuesToContentEnabled: Bool = false
 
+    /// When disabled, the data and/or highlights will not be clipped to contentRect. Disabling this option can
+    /// be useful, when the data lies fully within the content rect, but is drawn in such a way (such as thick lines)
+    /// that there is unwanted clipping.
+    @objc open var clipDataToContentEnabled: Bool = true
+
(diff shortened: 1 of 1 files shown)


jjatie:
@liuxuan30 looks good to me. We are going to have to release a 3.1.1 for Swift 4.1, so I'm fine with this going to master.

liuxuan30:
sorry to dig this up. I generated the change log and find this PR merged

My concern is `renderer.drawHighlighted(context: context, indices: _indicesToHighlight)`, the highlighted shapes won't be clipped at all. Not sure the impact.

jjatie:
@liuxuan30 I'm not sure what your concern is? The default behaviour is the same as before.

liuxuan30:
if we turn off `clipDataToContentEnabled`, `drawHighlighted` will not be clipped as well. This could introduce some side effects. However I'm not sure if people are really drawing out of bound

jjatie:
I’m fine with reversing the commit

wtmoose:
Hey guys, as the documentation says, this option is specifically for scenarios where the data lies fully within the content rect, so I don't see what the concern is:

````
/// When disabled, the data and/or highlights will not be clipped to contentRect. Disabling this option can
/// be useful, when the data lies fully within the content rect, but is drawn in such a way (such as thick lines)
/// that there is unwanted clipping.
````

For what it's worth, I ended up implementing my own charts because I couldn't achieve the look I wanted. Consider breaking some of the larger methods into smaller pieces with custom rendering in mind.

liuxuan30:
I'm ok to keep. Just remember we might have an issue :)

danielgindi:
Just bumped into this, a better solution for this would *probably* be inflating the content rect with half the stroke width.
Will not have side-effects with line charts etc.

PR #3421, opened 2018-04-25
Add Objective-c compatible for turning off drag in X and Y Axis separately

lennonhe:
### Goals :soccer:
Currently, turning off drag in X and Y Axis separately is only available for Swift.
This PR adds objective-c compatible for enabling/disabling drag in either axis separately.

Add `@objc` tag for `dragXEnabled`, `dragYEnabled` properties.


liuxuan30:
@jjatie is this a regression that we forgot to add @objc for these two?

jjatie:
@liuxuan30 no, not a regression. Originally you couldn't turn of x & y independently. When that feature was added, it was never exposed to objective-c

PR #3435, opened 2018-04-30
Added clamping function for `Comparable`

jjatie:
Adding convenience function. There is discussion in the Swift forums about adding this to the language, so we may be able to remove it in the future.

### DIFF
--- a/Source/Charts/Components/ChartLimitLine.swift
+++ b/Source/Charts/Components/ChartLimitLine.swift
@@ -68,18 +68,7 @@ open class ChartLimitLine: ComponentBase
         }
         set
         {
-            if newValue < 0.2
-            {
-                _lineWidth = 0.2
-            }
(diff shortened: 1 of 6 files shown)


Source/Charts/Components/ChartLimitLine.swift
```diff
@@ -68,18 +68,7 @@ open class ChartLimitLine: ComponentBase
...
-            }
-            else if newValue > 12.0
-            {
-                _lineWidth = 12.0
-            }
-            else
-            {
-                _lineWidth = newValue
-            }
+            _lineWidth = newValue.clamped(to: 0.2...12)
```
liuxuan30: here 12 is cast to Double by default? is this new swift feature?

jjatie: This has been the case since Swift 1

PR #3445, opened 2018-05-07
Make legendRenderer property public in order to be externally customizable

nagykatalin:
### Goals :soccer:
The main idea behind this pull request is to make the legend renderer externally customizable and overridable.

Our customer's request was to detect when the user's clicks on a legend and to highlight corresponding line on the chart.

By making public the legendRenderer property (similarly to the renderer property) we would like to use our own legend renderer class which stores the exact location of the labels and in this way we can detect if the user clicked on a label or not.

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -102,7 +102,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     open var noDataTextAlignment: NSTextAlignment = .left
 
     /// The renderer object responsible for rendering / drawing the Legend.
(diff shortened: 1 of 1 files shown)


jjatie:
@nagykatalin I'm okay with this, however please target 4.0.0 instead of master.
@liuxuan30 Any comments?

liuxuan30:
I'm ok too, giving the fact other major renderers are already public. But I recommend be careful about the parameters, as ` LegendRenderer(viewPortHandler: _viewPortHandler, legend: _legend)`,  `viewPortHandler` may not be available outside.

petester42:
I think legend still being readonly is fine.

petester42:
Thanks 👍

liuxuan30:
good to have you back:)

PR #1917, opened 2016-12-05
Add rounded corners on the bar chart #1066

Neral:
Added ability to draw rounded corners on each bar.

By default rounded corners is disabled and to enable it `isDrawRoundedBarEnabled` should be used.
Also using `barRoundingCorners` could be decided which corners should be rounded ([reference](https://developer.apple.com/reference/uikit/uirectcorner)).

Feature: #1066 

charts

PR #3504, opened 2018-06-11
Add flag to render Bar chart with rounded Bars

philfi:
### Goals :soccer:
Wanted to be able to render a vertical bar chart with a nice rounded bar look as opposed to a simple rectangle.

 - Add a flag (`isDrawRoundedBarsEnabled`) to existing `BarChartDataProvider` (and associated classes) to denote whether bars will be rendered as rounded
 - Update `BarChartRenderer` to render a rounded rectangle (using UIBezierPath) when the `dataProvider` has its `isDrawRoundedBarsEnabled` property set to `true`
- Also updated rendering of borders to ensure they match the filled space.

### Todo:
 - [ ] Add support for macOS (need to use NSBezierPath or similar)

### DIFF
--- a/Source/Charts/Charts/BarChartView.swift
+++ b/Source/Charts/Charts/BarChartView.swift
@@ -20,6 +20,9 @@ open class BarChartView: BarLineChartViewBase, BarChartDataProvider
 
     /// if set to true, a grey area is drawn behind each bar that indicates the maximum value
     private var _drawBarShadowEnabled = false
+
+    /// if set to true, bars will be drawn with rounded ends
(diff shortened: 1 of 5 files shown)


liuxuan30:
#1917 is this a dup or different approach to implement?

philfi:
@liuxuan30 hadn't seen that before, was mostly just sharing what I had done in my fork (which in retrospect was waaaayyy too specific to my use case). Looks like #1917 implementation is further along though albeit stalled, will get acquainted with what's going on there to see if I can contribute something and close this.

Issue #1060, opened 2016-05-21
Add Accessibility support for charts.


madnik:
Currently graphs are not accessible to VoiceOver and other accessible technologies.

danielgindi:
A PR is welcome :-)

liuxuan30:
What does the support looks like?

graesser-tiaa:
See this example for alt tags to add accessibility support to a web page  - http://accessibilityresources.org/joomla/index.php/blog/alt-attributes-it-s-not-that-hard
The blog post shows what to do for a bar chart.

pauljadam:
If you want to see what an accessible chart looks like on iOS just view the Stocks app from Apple it's got VoiceOver Accessibility for data points.

https://pbs.twimg.com/media/DOR8G6lVQAEQ7Hi.jpg:large


 I found this chart library because they're using it at work and it's not accessible. Can you copy the Apple stocks implementation?

mathewa6:
Hello!

Thanks for the great library. I'd be happy to try and tackle this issue for as many chart types as possible, so feel free to assign it to me. 

While I'm working on it, does anyone have any resources or recommendations for what the Scatter and Combined chart should sound like to a VO user? Right now I'm leaning toward a axis based range summarization (For eg, the sample scatter plot would sound like "0 to 10, DS1: 9 points from 81 to 99, DS2: 5 points between 50 and 87... etc"), however as you can imagine in these two cases, it can be an auditory overload given enough categories.

The other option is to simply let the creator describe the visual. But in that case should the description be a required property? This might not work as well for charts with dynamic data.

There is also the option for simply narrating a table of values, but usability wise, I think that is a poor choice, since the point of a visualization is to make the information contained palatable.

As of now these (and variants) are what I'm working on:
- [ ] Line
- [ ] - Bar
- [x] - Pie
- [ ] - Bubble
- [ ] - Radar

petester42:
Thanks for looking into this @mathewa6. This resource might help you. It’s for the web but the same ideas apply https://www.w3.org/WAI/tutorials/images/complex/

mathewa6:
@petester42 Thanks! I hadn't considered splitting descriptions so that is very helpful.

[iOS/tvOS](https://developer.apple.com/documentation/uikit/accessibility/uiaccessibilitycontainer) allows for a little more interactivity and that way, most folks who use this library will get accessibility "for free" based on titles and legends already provided, without needing to manually add descriptions.

[...]

PR #3520, opened 2018-06-20
Accessibility Support for (most) Chart types

mathewa6:
### Issue Link :link:

[Accessibility Support (#1060) link](https://github.com/danielgindi/Charts/issues/1060)

### Goals :soccer:

This PR adds VoiceOver (VO) support to the following chart types:

- Line Chart
- Bar Chart
- Pie Chart
- Bubble Chart
- Radar Chart
- CandleStick Chart

This PR aimed to minimize intrusion with rendering or drawing of Charts as possible while still remaining usable. I'm leaving out support for Combined charts and Scatter charts for now for the same reason, as I think they need a bit more work to be made accessible in a usable manner (i.e. without an information overload to a VO user).

Accessibility is added as extensions to NSUIView, which is done in Platforms+Accessibility.swift to abstract differences between macOS/iOS. With that done, all individual ChartRenderers need to do is populate `accessibleChartElements` property of NSUIView during drawing. Some chart types like Bar Charts which can have multiple datasets/stacks use an `accessibilityOrderedElements` array internally to reorder elements to make more sense when using VO, rather than in drawing order. There are some minor optional properties in ChartData.swift to help make data sound better when presented over VO such as prefixes, suffixes to each data point.

### Testing Details :mag:

No changes to drawing, so none added. Manual on-device testing & audited for usability at WWDC.


petester42:
Do you have any idea why voice over is saying "padding" when going through the datasets?

petester42:
Great work btw. Seems to work good with the demos.

Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -603,13 +609,31 @@ open class LineChartRenderer: LineRadarRenderer
...
+
+        // Make the chart header the first element in the accessible elements array
+        if let chart = dataProvider as? LineChartView {
+            let chartDescriptionText = chart.chartDescription?.text ?? ""
+            let dataSetDescriptions = lineData.dataSets.map { $0.label ?? "" }
+            let dataSetDescriptionText = dataSetDescriptions.joined(separator: ", ")
+            let dataSetCount = lineData.dataSets.count
+            let
+            element = NSUIAccessibilityElement(accessibilityContainer: chart)
+            element.accessibilityLabel = chartDescriptionText + ". \(dataSetCount) dataset\(dataSetCount == 1 ? "" : "s"). \(dataSetDescriptionText)"
```
petester42: The screen reader reads "padding" at the end of the label.

Source/Charts/Renderers/PieChartRenderer.swift
```diff
@@ -131,27 +133,43 @@ open class PieChartRenderer: DataRenderer
...
+        // NOTE: - Since we want to summarize the total count of slices/portions/elements, use a default string here
+        // This is unlike when we are naming individual slices, wherein it's alright to not use a prefix as descriptor.
+        // i.e. We want to VO to say "3 Elements" even if the developer didn't specify an accessibility prefix
+        // If prefix is unspecified it is safe to assume they did not want to use "Element 1", so that uses a default empty string
+        let prefix: String = chart.data?.accessibilityEntryLabelPrefix ?? "Element"
+        let description = chart.chartDescription?.text ?? dataSet.label ?? chart.centerText ??  ""
+
+        let
+        element = NSUIAccessibilityElement(accessibilityContainer: chart)
+        element.accessibilityLabel = description + ". \(entryCount) \(prefix + (entryCount == 1 ? "" : "s"))"
```
petester42: The screen reader reads "padding" at the end of the label.

Source/Charts/Renderers/RadarChartRenderer.swift
```diff
@@ -37,7 +52,21 @@ open class RadarChartRenderer: LineRadarRenderer
...
+
+            // If we redraw the data, remove and repopulate accessible elements to update label values and frames
+            self.accessibleChartElements.removeAll()
+
+            // Make the chart header the first element in the accessible elements array
+            if let chartDescriptionText: String = chart.chartDescription?.text {
+                let dataSetCount = radarData!.dataSets.count
+                let
+                element = NSUIAccessibilityElement(accessibilityContainer: chart)
+                element.accessibilityLabel = chartDescriptionText + ". \(dataSetCount) dataset\(dataSetCount == 1 ? "" : "s")"
```
petester42: The screen reader reads "padding" at the end of the label.

Source/Charts/Renderers/RadarChartRenderer.swift
```diff
@@ -37,7 +52,21 @@ open class RadarChartRenderer: LineRadarRenderer
         if radarData != nil
         {
             let mostEntries = radarData?.maxEntryCountSet?.entryCount ?? 0
-            
+
+            // If we redraw the data, remove and repopulate accessible elements to update label values and frames
+            self.accessibleChartElements.removeAll()
+
+            // Make the chart header the first element in the accessible elements array
+            if let chartDescriptionText: String = chart.chartDescription?.text {
```
petester42: Why does this chart not say the datasets in the chart description is present?

Source/Charts/Renderers/RadarChartRenderer.swift
```diff
@@ -71,7 +100,20 @@ open class RadarChartRenderer: LineRadarRenderer
...
+
+        let prefix: String = chart.data?.accessibilityEntryLabelPrefix ?? "Item"
+        let description = dataSet.label ?? ""
+
+        // Make a tuple of (xLabels, value, originalIndex) then sort it
+        // This is done, so that the labels are narrated in decreasing order of their corresponding value
+        // Otherwise, there is no non-visual logic to the data presented
+        let accessibilityEntryValues =  Array(0 ..< entryCount).map { (dataSet.entryForIndex($0)?.y ?? 0, $0) }
+        let accessibilityAxisLabelValueTuples = zip(accessibilityXLabels, accessibilityEntryValues).map { ($0, $1.0, $1.1) }.sorted { $0.1 > $1.1 }
+        let accessibilityDataSetDescription: String = description + ". \(entryCount) \(prefix + (entryCount == 1 ? "" : "s")). "
```
petester42: The screen reader reads "padding" at the end of the label.

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -18,6 +18,28 @@ import CoreGraphics
 
 open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
```
petester42: In the demo project the "Another Bar Chart" demo has a frame that is too large around the bars.

mathewa6:
Thanks! 

Regarding omission of Scatter & Combined charts: the same method using `accessibleChartElements` can be used, but almost all the folks we tested with found it challenging to use (We tested most charts by showing a simple example and asking questions like "Can you find the day when x was smallest?"). So I left that for another PR: since a "summarizing" method will probably need fiddling with the rendering to allow parsing by axis values rather than by chart data itself.
Now that you mention it: I will add the chart itself as an accessibilityElement, so that at least a label/description can be added in the case of Scatter & combined.

I will look into the "padding" issue: not sure why that occurs.

Source/Charts/Renderers/BarChartRenderer.swift
mathewa6: I see it! Will fix tomorrow.

Source/Charts/Renderers/RadarChartRenderer.swift
mathewa6: I believe this was because, for Radar chart, each individual dataSet is a heading element, so it's the next item in VO scroll order. Can definitely be added here, but I thought it was redundant, so let me know if you think I should add it.

mathewa6:
@petester42 

Alright! I've addressed the rects in the AnotherBarChart sample and made the headers/titles more consistent overall.

I cannot reproduce the "padding" issue. The word padding doesn't even appear in the project (excluding comments/variables) so I'm a little lost.
Is there a specific config you are using? Perhaps traverse the element by character using VO and see if indeed it is saying "padding" and I can look over what string it is saying?

petester42:
Thanks for the changes. I still have the padding issue. I don't have any special settings, I just turned on voiceover without changing anything (iOS 11.4) and was trying out navigating by swiping through elements. If you can't find out why we can merge and then open an issue if other people have this problem.

petester42:
Is it possible that it is reading the CGRect for the frame?

mathewa6:
Interesting. I will play with it again and see if it could be the frame or perhaps a superclass' label. It doesn't occur on 11.4.1 beta or 12 beta. Will test on 11.4 as well to be sure. 

If you or another reviewer can try on another device and it still occurs, probably needs to be fixed before merge.
I'm wondering though, if it's somehow distorting the "Heading" text it should be saying. Most of these charts have the top level or multiple elements (Radar ch) as heading, which causes VO to explicitly append "heading" to the end of the label. I'm imagining it is messing it up somehow and saying "padding". That's a funny bug to have 😄

petester42:
It is a weird one for sure. I tried debugging it for a bit and my only thought would be that it is reading the whole element rather than just the label. I have an older phone with maybe an older version I’ll try it out there when I have time.

mathewa6:
Alright, in that case, it might be worth trying to see what happens if you remove any attributes like heading and set the label to some static text constant. If it still shows then perhaps see if its there in other apps too? FYI it doesn't occur on 11.4 for me either.

Thanks for looking into it and sorry about this!

petester42:
Alright. Thanks. For looking into this. I’ll check stuff out tonight and if it’s only my phone then I’ll merge. We can fix it later if it is an issue.

mathewa6:
That's fine. I couldn't make it happen on another device either so I'm not too concerned. 

I'll be opening another PR for panning support with VO and the remaining chart types, so if it does become an issue I'd be willing to look again.

Thanks for the merge!

acegreen:
@petester42  hey, any chance we can override the elements on a bar selection?

Matthew1874:
hello, does combined chart support accessibility?
Hope for reply
Thank you

PR #3528, opened 2018-06-24
fix the error title for demo

yangasahi:


ChartsDemo-iOS/Objective-C/Demos/PieChartViewController.m
```diff
@@ -28,7 +28,7 @@ - (void)viewDidLoad
 {
     [super viewDidLoad];
     
-    self.title = @"Pie Bar Chart";
```
liuxuan30: @petester42 @jjatie need some advise on the name. Pie Bar Chart is ok or kind of typo ?

yangasahi: IMO
"Pie Chart" is better.

liuxuan30: I would agree. It's more clear.

ChartsDemo-iOS/Swift/Demos/PieChartViewController.swift
```diff
@@ -21,7 +21,7 @@ class PieChartViewController: DemoBaseViewController {
         super.viewDidLoad()
         
         // Do any additional setup after loading the view.
-        self.title = "Half Pie Bar Chart"
+        self.title = "Pie Chart"
```
liuxuan30: I think we should keep the 'half' word?

yangasahi: I think that the "Pie Chart" is correct since the controller name is "PieChartViewController".
BTW, I have fixed the title to "Half Pie Chart" from  "Half Pie Bar Chart" for "HalfPieChartViewController.swift"

liuxuan30: I see.

liuxuan30:
Thanks!

Issue #2060, opened 2017-01-10
Problem selecting bubbles on the same xIndex


chuynadamas:
Hi @danielgindi and @liuxuan30 again, thank you for this amazing job!

I found a little issue here, when you have a chart with some bubbles at the same xIndex, something like that, you are not able to select all of them.

![screen shot 2017-01-10 at 12 35 33 pm](https://cloud.githubusercontent.com/assets/5713880/21819455/5c85f9e2-d731-11e6-8ad9-00472fa78f4d.png)

I did a little research and i discover that the following comparison on **ChartDataSet.swift** the problem

```swift
                    if abs(value.y - yValue) < abs(closestYValue - yValue)
                    {
                        closestYValue = yValue
                        closestYIndex = closest
                    }
 ```

If we add an equal ```=``` now we are able to select every bubble.

```swfit
                    if abs(value.y - yValue) <= abs(closestYValue - yValue)
                    {
                        closestYValue = yValue
                        closestYIndex = closest
                    }
```

![ezgif com-video-to-gif](https://cloud.githubusercontent.com/assets/5713880/21820718/7bc2f59e-d736-11e6-81b1-28a8751b0101.gif)

liuxuan30:
sounds like you solved it? PR welcome :)

PR #3548, opened 2018-07-05
Support inlune bubble viz selection

chuynadamas:
### Issue Link :link:
This is for the inline bubble selection #2060

### Goals :soccer:
Have the ability to select multiple bubbles in the same index

Just take in consideration the less than or equal ```abs(closestYValue - yValue)```

### Testing Details :mag:
Implementing in the my project test it in the demo also


liuxuan30:
wow, an ancient issue with a modern fix! (joking) thanks!

PR #2640, opened 2017-07-25
added color for each legendEntry(if not set, using default legend color)

SeRG1k17:
Simple way add color for each legendEntry.
Usage:

```
    NSMutableArray *legendEntries = [[NSMutableArray alloc] init];
    
    for (int i = 0; i < count; i++)
    {
        UIColor *entryColor = colors[i];
        ChartLegendEntry *entry = [[ChartLegendEntry alloc] initWithLabel:parties[i % parties.count]
                                                               labelColor:entryColor // <-- New property
                                                                     form:ChartLegendFormCircle
                                                                 formSize:8
                                                            formLineWidth:0
                                                        formLineDashPhase:0
                                                      formLineDashLengths:0
                                                                formColor:entryColor];
        
        [legendEntries addObject:entry];
    }
    
    [_chartView.legend setCustomWithEntries: legendEntries];
```

PR #3558, opened 2018-07-14
Add label colors to legend entries

petester42:
Adds a color setting to legend entries to customize labels. Addresses the feedback in #2640.

### DIFF
--- a/Source/Charts/Renderers/LegendRenderer.swift
+++ b/Source/Charts/Renderers/LegendRenderer.swift
@@ -59,6 +59,7 @@ open class LegendRenderer: NSObject, Renderer
                         entries.append(
                             LegendEntry(
                                 label: sLabels[j % sLabels.count],
+                                labelColor: nil,
                                 form: dataSet.form,
                                 formSize: dataSet.formSize,
                                 formLineWidth: dataSet.formLineWidth,
@@ -76,6 +77,7 @@ open class LegendRenderer: NSObject, Renderer
                         entries.append(
                             LegendEntry(
                                 label: dataSet.label,
+                                labelColor: nil,
                                 form: .none,
                                 formSize: CGFloat.nan,
                                 formLineWidth: CGFloat.nan,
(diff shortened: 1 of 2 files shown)


Source/Charts/Renderers/LegendRenderer.swift
```diff
@@ -59,6 +59,7 @@ open class LegendRenderer: NSObject, Renderer
                         entries.append(
                             LegendEntry(
                                 label: sLabels[j % sLabels.count],
+                                labelColor: nil,
```
liuxuan30: is it possible to use default value? seeing `labelColor: nil` repeat

petester42: We could but the rest of that function doesn’t. Seems weird to have a default at the start of the function. That’s my preference though

petester42: We could but the rest of that function doesn’t. Seems weird to have a default at the start of the function. That’s my preference though

liuxuan30: ok. if @jjatie has no question or no response for a while, we could merge then.

petester42: Cool. Trying to finish up 4.0

jjatie: I think the issue is that we are putting too many things in the initializer. The only thing we need to create a valid `LegendEntry` is the `label`, everything else is optional. The `label` should be a constant and the rest of the properties remain vars.

jjatie: It might even make sense to only allow the entry group to be created through the empty initializer, i.e.:
```
@objc(ChartLegendEntry)
open class LegendEntry: NSObject
{
@objc(ChartLegendEntry)
open class LegendEntry: NSObject
{
    public override init()
    {
        super.init()
    }
    
    /// A label to start a group.
    @objc public init(label: String?)
    {
        self.label = label
    }
    public override init()
    {
        super.init()
    }
    
    /// - parameter label:                  The legend entry text.
    @objc public init(label: String)
    {
        self.label = label
    }
    
    // ...
}
```

petester42: Don’t know if the empty initializer is useful but I agree with your points. I’ll change some stuff

petester42:
@jjatie I have addressed your comments

petester42:
I was thinking the same thing but with all the variables. I’d rather they all be immutable.

petester42:
We can go through the whole API in maybe v5 to address having too many mutable variables

Issue #3550, opened 2018-07-06
horizontalBezier bad performance


piotrros:
## What did you do?

Create a simple line chart with `.horizontalBezier` mode enabled.

## What did you expect to happen?

As usual: draw a chart.

## What happened instead?

With, for example, 2000 values it takes about one minute to render this chart. On Android (using MPAndroidChart) it can even handle more than 10000 values easily. Removing `set.mode = .horizontalBezier` helps, but it I think this is a real issue as 2000 values aren't that much. Just a funny note: it the real app user had 30000 values on his chart (actually there are 3 charts 30000 values each). I tested on his account and it took 90 minutes (!) to render this chart. From the other hand on Android, it took about a 2 minutes. I know 30000 is just too much so I'll try to reduce the number of values, but still, 2000 taking so much time is really bad. That forces me to disable `.horizontalBezier` when there are more than 1000 values.

## Charts Environment

**Charts version/Branch/Commit Number:**
3.1.1
**Xcode version:**
9.4.1
**Swift version:**
4.1
**Platform(s) running Charts:**
iOS 11.4 (tried on iPhone 6 and on iPad Pro 10.5')
**macOS version running Xcode:**
10.13.5
## Demo Project

This is a demo project with modified LineChart1ViewController.swift file. Just run the app and tap the first option (Line Chart). My project is much more complex so I've created this simple demo as removed as much code as possible.

Just in case I'm also putting here source code directly:

```
import UIKit
import Charts

class LineChart1ViewController: DemoBaseViewController {

    @IBOutlet var chartView: LineChartView!
    @IBOutlet var sliderX: UISlider!
    @IBOutlet var sliderY: UISlider!
    @IBOutlet var sliderTextX: UITextField!
    @IBOutlet var sliderTextY: UITextField!

    override func viewDidLoad() {
        super.viewDidLoad()
        self.options = []
        self.setDataCount(2000, range: 100)
    }
    
    func setDataCount(_ count: Int, range: UInt32) {
        let values = (0..<count).map { (i) -> ChartDataEntry in
            let val = Double(arc4random_uniform(range) + 3)
            return ChartDataEntry(x: Double(i), y: val)
        }
        
        let set = LineChartDataSet(values: values, label: "DataSet 1")
        set.mode = .horizontalBezier
        
        let data = LineChartData(dataSet: set)
        chartView.data = data
    }
    
    override func optionTapped(_ option: Option) {}

[...]

PR #3577, opened 2018-07-23
Improved performance drawHorizontalBezier

Banannzza:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/3550

### Goals :soccer:
Reduce execution time of drawHorizontalBezier

Explanation of the decision on the link: https://stackoverflow.com/a/6353597

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -246,6 +246,10 @@ open class LineChartRenderer: LineRadarRenderer
         context.beginPath()
         context.addPath(cubicPath)
         context.setStrokeColor(drawingColor.cgColor)
+        context.setLineCap(CGLineCap.butt)
+        context.setLineJoin(CGLineJoin.round)
+        context.setMiterLimit(10.0)
+        context.setFlatness(0.6)
         context.strokePath()
         
         context.restoreGState()


liuxuan30:
hi there, thanks for investigating on this. Have you tested on your side? Can you post the comparison results?

Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -246,6 +246,10 @@ open class LineChartRenderer: LineRadarRenderer
         context.beginPath()
         context.addPath(cubicPath)
         context.setStrokeColor(drawingColor.cgColor)
+        context.setLineCap(CGLineCap.butt)
```
liuxuan30: we have a var for line cap
```
    /// Line cap type, default is CGLineCap.Butt
    open var lineCapType = CGLineCap.butt
```
in line chart data set.

Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -246,6 +246,10 @@ open class LineChartRenderer: LineRadarRenderer
...
+        context.setLineJoin(CGLineJoin.round)
```
liuxuan30: If configuring line cap and line join, miter limit, flatness will improve the performance, please create a new var for each of them, and maybe help to test the performance on your side.

I'm not sure if it has to be hardcoded params.

liuxuan30:
@petester42 @jjatie @danielgindi are you aware of this subtle trick?

petester42:
Nope

liuxuan30:
I have given some thoughts on this, I'm not sure if setting those properties could improve our performance. The difference is the context of our `context`.

@Banannzza can you post the performance data, e.g. instruments time cost before and after you apply these changes?

jjatie:
@liuxuan30 @Banannzza I read the SO post differently. Why are we not just using a `UIBezierPath`? If even in their best implementations they are equal in performance, we should be using the higher level API. If Apple is able to make further optimizations, `UIBezierPath` will get it for free.

petester42:
Sure it would be a better API for adding points to draw but it won’t address the core performance issue. The issue is that we are using core graphics uses too much cpu when drawing.

jjatie:
@petester42 Since we're only supporting iOS 8+, shouldn't these CG calls be hitting the GPU more than the CPU? Since Metal was introduced Core Graphics is supposed to be backed by Metal. I've not looked into this too much.

petester42:
No idea. Historically it has not been the case. I believe the actual drawing is done on the gpu but all the setup code is not.

liuxuan30:
what's the status of this? Should we merge or hold?

Issue #3590, opened 2018-07-31
CandleStickChart bars have not equal width


lazik234:
## What did you do?

Draw CandleStick chart with 30+ bars, using filled and not filled bars

## What did you expect to happen?

All bars should have same width and equal spacing

## What happened instead?

Not filled bars are 1 pixel wider then filled, spacing is not equal

## Charts Environment

**Charts version/Branch/Commit Number: Charts (3.0.4)**
**Xcode version: 9.4**
**Swift version: 4.0**
**Platform(s) running Charts: iOS8+**
**macOS version running Xcode: 10.13.5**

## Demo Project
Bug is easy to reproduce on any iPhone, just draw a graph with 30+ bars, using filled and not filled bars. See screenshot:
![screen shot 2018-07-31 at 19 45 56](https://user-images.githubusercontent.com/1306465/43474795-82fbff2e-94fc-11e8-88af-0c9d6be2ab5b.png)

PR #3593, opened 2018-08-01
Fix for candle chart bar width (#3590)

lazik234:
### Issue Link :link:
Fix for #3590

### Goals :soccer:
It will make width of candle chart bars equal

According to documentation of "stroke" method: https://developer.apple.com/documentation/coregraphics/cgcontext/1454679-stroke
"The line straddles the path, with half of the total width on either side." It makes stroked bars 1 pixel (half of pixel on both sides) wider, and it is visible on iPhone if bars count is 30+.

### Testing Details :mag:
Screenshot after fix applied: https://user-images.githubusercontent.com/1306465/43476381-229e4916-9501-11e8-966f-79f676822e3e.png

### DIFF
--- a/Source/Charts/Renderers/CandleStickChartRenderer.swift
+++ b/Source/Charts/Renderers/CandleStickChartRenderer.swift
@@ -169,7 +169,10 @@ open class CandleStickChartRenderer: LineScatterCandleRadarRenderer
                 trans.rectValueToPixel(&_bodyRect)
                 
                 // draw body differently for increasing and decreasing entry
-
+                let strokeRect = CGRect(x:_bodyRect.origin.x+0.5,
+                                        y:_bodyRect.origin.y+0.5,
+                                        width:_bodyRect.size.width-1,
+                                        height:_bodyRect.size.height-1)
                 if open > close
                 {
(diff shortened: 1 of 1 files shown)


liuxuan30:
this may take longer. We need to prioritize some big PRs first.

anton-filimonov:
IMHO, unfortunately, this PR does not actually fix the problem (if there is one). 
First reason is the `_bodyRect` coordinates initially are not integer values because they are calculated with transformation matrix and not aligned by any way after that. (so we first have to round origin and size and than apply additional changes)
The second reason is this solution does not take into account the `scale` of the current screen and the `line width` ([the documentation](https://developer.apple.com/library/archive/documentation/2DDrawing/Conceptual/DrawingPrintingiOS/GraphicsDrawingOverview/GraphicsDrawingOverview.html) part  "Points Versus Pixels" gives details)
Also, if we want to align candle to physical pixels, we have to take into account shadows, not just candle bodies
And if we do the physical pixel alignment it could make scrolling/scaling not so smooth (not sure about that, need to test), and if so, probably this alignment would better be optional.

liuxuan30:
@lazik234 what's your feedback to @anton-filimonov?

PR #3599, opened 2018-08-10
Know when user is done interacting with graph

dtroupe18:
### Goals :soccer:

Easily be able to determine when a user stops interacting (highlighting) with a chart.

I made the delegate method optional because it shouldn't have to be implemented. It's just an additional feature that some developers might find useful.

When a user stops a panGesture on a chart the delegate method will call back. This allows for the developer to know when the user stopped highlighting a graph. They can then update the chart accordingly. I used the feature to transition between a cubic (smooth) and linear graph as the user starts and stops highlighting a graph. Similar to the charts in the RobinHood App.

### Testing Details :mag:

It's just an optinal callback so didn't add any tests.


liuxuan30:
can you check if latest master fixed yours?

liuxuan30:
fixed by #3612

PR #3612, opened 2018-08-27
Feature - ChartView Pan Ended Delegate Call

AntonTheDev:
### Issue Link :link:

Over the years, every I have come to implement this library, i have found that 1 feature for me has always been missing which is the ability to know when the user stopped panning across a chart to display values on the fly, and have it reset once the user lets go. I've yet to find a clean solution other than modifying the library, and I have added some callback logic to the ChartViewDelegate

### Goals :soccer:
To be able to install this from the master repo, and not always have to fork it to add this exact feature./

Added and optional method to the  ChartViewDelegate to tell the delegate that the pan gesture has reached a cancelled / ended state

```
public protocol ChartViewDelegate
{
     ...

    /// Called when a user stop highlighting values while panning
    @objc optional func chartViewDidEndPanning(_ chartView: ChartViewBase)
}
```

### Testing Details :mag:
None, there are no specific tests. This should not effect the snapshotting at all, and has no effect on the existing codebase other than an optional callback.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -779,6 +779,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 }
                 
                 _isDragging = false
+            
+                delegate?.chartViewDidEndPanning?(self)
             }
             
             if _outerScrollView !== nil
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -779,8 +779,6 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 }
                 
                 _isDragging = false
-            
-                delegate?.chartViewDidEndPanning?(self)
             }
             
             if _outerScrollView !== nil
@@ -788,6 +786,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 _outerScrollView?.nsuiIsScrollEnabled = true
                 _outerScrollView = nil
             }
+            
+            delegate?.chartViewDidEndPanning?(self)
         }
     }
     
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -25,6 +25,9 @@ public protocol ChartViewDelegate
     /// - parameter highlight: The corresponding highlight object that contains information about the highlighted position such as dataSetIndex etc.
     @objc optional func chartValueSelected(_ chartView: ChartViewBase, entry: ChartDataEntry, highlight: Highlight)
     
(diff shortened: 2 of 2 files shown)


jjatie:
@liuxuan30 @petester42 This looks good to me!

proped:
I need this. Could u merged it ?

AntonTheDev:
I updated the comment accordingly in the code, hope this helps :)

natan:
@liuxuan30 Any update on this?

liuxuan30:
I'm open to merge.but as @petester42 said, if we add this, maybe we should add more?

natan:
@liuxuan30 @AntonTheDev Great, thank you for moving this along!

liuxuan30:
shall we add at least `didStartPanning` to pair for this PR?

AntonTheDev:
I believe the whole "didStartPanning" can be be assumed when "chartValueSelected" is called? Since you can't really start panning without making a callback

anton-filimonov:
I believe it'd be useful to make the callbacks similar to what we have for scrollview. And actually call them in pretty much the same situations as ScrollView does.

jjatie:
I second making the delegate behave and be named as closely to UIScrollViewDelegate as possible.

liuxuan30:
@AntonTheDev it's just strange to me that we had didEnd without didStart. Though you might be able to know it get started through chartValueSelected, I think it's still different events.

Arkezis:
Thanks for your awesome library ! It is really helping us :)
This PR would be awesome for our application, we are looking forward to merge it :)

liuxuan30:
@petester42 @jjatie how about this PR? shall we add a few more or just merge? I'm looking for a few PRs that can be merged soon and release a new version next week.

jcooperation0137:
@AntonTheDev thanks for your improvement. 
Have you checked chartViewDidEndPanning for a LineChartView inside of a tableviewcell? 
I've checked, the delegate doesn't work. Do you mind check it again on your side?

jlubeck:
> @AntonTheDev I'm ok to merge this, but it's just strange to me that we had didEnd without didStart. Though you might be able to know it get started through chartValueSelected, I think it's still different events.
> 
> How about we append the `didStartPanning` in `recognizer.state == NSUIGestureRecognizerState.began` closure?

Yes please! I'm running into this situation myself right now. chartValueSelected get called both on drag AND tap, but chartViewDidEndPanning only gets called when the user stops dragging. Definitely different use cases...

Specifically, when this chart gets embedded inside a UIScrollView, we want to stop the UIScrollView from dragging when we are dragging on the chart. So one would think we can stop the UIScrollView from dragging on chartValueSelected, and restart it on chartViewDidEndPanning

But the problem with this solution is that chartValueSelected is also called on a single tap on the chart. So when this happens, the UIScrollView stops dragging, but then we have no callback to restart it... And the user is stuck with an unscrollable UIScrollView until get actually drags on the Chart.

Hope this was clear. I'm about to leave for the day, but if I don't hear from anyone back I'll probably make a pull request myself tomorrow.

PR #3587, opened 2018-07-27
update barRect.size.height calculation

shiwangwang:
when bottom equals top, offset is not necessary
fix  #3586

PR #3650, opened 2018-09-19
improvements in barRect height calculation 

potato04:
###  issue
My last PR #3587 fixed bar height calculation when all y values equals 0.
Recently,  I found bars will displayed in an unexpected size when all y values less than 0.
And the black border will be oversized too when voice over is enable. 
img_0333

### how to reproduce
replace those codes
https://github.com/danielgindi/Charts/blob/5087a04bbefc9f15995113954ce217bbfb445577/ChartsDemo-iOS/Swift/Demos/AnotherBarChartViewController.swift#L64-L68
with 
``` swift
        let yVals = [
            BarChartDataEntry(x: 0, y: -10),
            BarChartDataEntry(x: 1, y: -20),
            BarChartDataEntry(x: 2, y: -30),
            BarChartDataEntry(x: 3, y: -15)
        ]
```

### other
I moved the offset calculation code out of the loop, calculate once is enough
https://github.com/danielgindi/Charts/blob/5087a04bbefc9f15995113954ce217bbfb445577/Source/Charts/Renderers/BarChartRenderer.swift#L112

### DIFF
--- a/Source/Charts/Components/YAxis.swift
+++ b/Source/Charts/Components/YAxis.swift
@@ -150,6 +150,9 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
+        min = max < min ? max : min
+        max = min > max ? min : max
+        
         // temporary range (before calculations)
         let range = abs(max - min)
         
--- a/Source/Charts/Components/YAxis.swift
+++ b/Source/Charts/Components/YAxis.swift
@@ -150,8 +150,19 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
-        min = max < min ? max : min
-        max = min > max ? min : max
+        if min > max {
+            if _customAxisMax && _customAxisMin {
+                //both max and min are manullay setted
+                min = max < min ? max : min
+                max = min > max ? min : max
+            } else if _customAxisMax && !_customAxisMin {
+                min = max - 1
+            } else if !_customAxisMax && _customAxisMin {
+                max = min + 1
+            }
+        }
+        
+        
         
         // temporary range (before calculations)
         let range = abs(max - min)
--- a/Source/Charts/Components/YAxis.swift
+++ b/Source/Charts/Components/YAxis.swift
@@ -150,8 +150,16 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
-        min = max < min ? max : min
-        max = min > max ? min : max
+        // Make sure max is greater than min
+        if min > max {
+            if _customAxisMax && _customAxisMin {
+                (min, max) = (max, min)
+            } else if _customAxisMax && !_customAxisMin {
+                min = max - 1
+            } else if !_customAxisMax && _customAxisMin {
+                max = min + 1
+            }
+        }
         
         // temporary range (before calculations)
         let range = abs(max - min)
--- a/Source/Charts/Components/YAxis.swift
+++ b/Source/Charts/Components/YAxis.swift
@@ -150,19 +150,8 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
-        if min > max {
-            if _customAxisMax && _customAxisMin {
-                //both max and min are manullay setted
-                min = max < min ? max : min
-                max = min > max ? min : max
-            } else if _customAxisMax && !_customAxisMin {
-                min = max - 1
-            } else if !_customAxisMax && _customAxisMin {
-                max = min + 1
-            }
-        }
-        
-        
+        min = max < min ? max : min
+        max = min > max ? min : max
         
         // temporary range (before calculations)
         let range = abs(max - min)
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -108,6 +108,28 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
         var barRect = CGRect()
         var x: Double
         var y: Double
+
+
+
+        // When drawing with an auto calculated y-axis minimum, the renderer actually draws each bar from 0
+        // to the required value. This drawn bar is then clipped to the visible chart rect in BarLineChartViewBase's draw(rect:) using clipDataToContent.
+        // While this works fine when calculating the bar rects for drawing, it causes the accessibilityFrames to be oversized in some cases.
+        // This offset attempts to undo that unnecessary drawing when calculating barRects, particularly when not using custom axis minima.
+        // This allows the minimum to still be visually non zero, but the rects are only drawn where necessary.
+        // This offset calculation is not necessary when bars won't be clipped (when y-axis minimum <= 0  and y-axis maximum  >= 0)
+        var heightOffset: CGFloat = 0.0
+        var originYOffset: CGFloat = 0.0
+        if let offsetView = dataProvider as? BarChartView {
+            let offsetAxis = offsetView.leftAxis.isEnabled ? offsetView.leftAxis : offsetView.rightAxis
+            if !offsetAxis._customAxisMin {
+                if barData.yMin > 0 {
+                    heightOffset = CGFloat(offsetAxis.axisMinimum)
+                }
+                if barData.yMax < 0 {
+                    originYOffset = CGFloat(offsetAxis.axisMaximum)
+                }         
+            }
+        }
         
         for i in stride(from: 0, to: min(Int(ceil(Double(dataSet.entryCount) * animator.phaseX)), dataSet.entryCount), by: 1)
         {
@@ -139,28 +161,10 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
                     bottom *= CGFloat(phaseY)
                 }
 
-                // When drawing with an auto calculated y-axis minimum, the renderer actually draws each bar from 0
-                // to the required value. This drawn bar is then clipped to the visible chart rect in BarLineChartViewBase's draw(rect:) using clipDataToContent.
-                // While this works fine when calculating the bar rects for drawing, it causes the accessibilityFrames to be oversized in some cases.
-                // This offset attempts to undo that unnecessary drawing when calculating barRects, particularly when not using custom axis minima.
-                // This allows the minimum to still be visually non zero, but the rects are only drawn where necessary.
-                // This offset calculation also avoids cases where there are positive/negative values mixed, since those won't need this offset.
-                var offset: CGFloat = 0.0
-                if let offsetView = dataProvider as? BarChartView {
-
-                    let offsetAxis = offsetView.leftAxis.isEnabled ? offsetView.leftAxis : offsetView.rightAxis
-
-                    if barData.yMin.sign != barData.yMax.sign { offset = 0.0 }
-                    else if !offsetAxis._customAxisMin {
-                        offset = CGFloat(offsetAxis.axisMinimum)
-                    }
-                }
-
                 barRect.origin.x = left
+                barRect.origin.y = top + originYOffset
                 barRect.size.width = right - left
-                barRect.origin.y = top
-                barRect.size.height = bottom == top ? 0 : bottom - top + offset
-
+                barRect.size.height = bottom - top + heightOffset - originYOffset
                 buffer.rects[bufferIndex] = barRect
                 bufferIndex += 1
             }
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -125,9 +125,11 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
                 if barData.yMin > 0 {
                     heightOffset = CGFloat(offsetAxis.axisMinimum)
                 }
+            }
+            if !offsetAxis._customAxisMax {
                 if barData.yMax < 0 {
                     originYOffset = CGFloat(offsetAxis.axisMaximum)
-                }         
+                }
             }
         }
         


potato04:
@petester42

liuxuan30:
@potato04 could this be a new issue of your #3587? If you revert #3587, is this still happening? 
I'm start thinking if #3587 is the right fix. checking `top == 0` seems not enough, combining the context of `prepareBuffer`.

liuxuan30:
I checked the old commits and your PR. However I found in most cases your fix seems correct, but if you try to set 
```swift
chartView.leftAxis.axisMaximum = -40.0
```
and with your test data, it shouldn't draw any bars, but instead, it draws 4 bars and wrong. seems a new issue to me. Not having a fix yet. It seems even I revert all of the changes, it still draw only one bar. Would mind take a look?

BTW, it's time to add UT for negative bars, and +/- bars, do you know how to add UT?

potato04:
I think I solved these problems.
1. `chartView.leftAxis.axisMaximum = -40.0`  This problem is due to the wrong calculation of originYOffset.
2. `It seems even I revert all of the changes, it still draw only one bar, and left axis labels are all gone.`  When max or min was set manually, sometimes it will cause min>max, so the y-axis will not work properly.
Please review my new commits.

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -125,9 +125,11 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
                 if barData.yMin > 0 {
                     heightOffset = CGFloat(offsetAxis.axisMinimum)
                 }
+            }
+            if !offsetAxis._customAxisMax {
```
liuxuan30: at first I was thinking a fix like this:
```swift
if !offsetAxis._customAxisMin || !offsetAxis._customAxisMax {
                if barData.yMin > 0 {
                    heightOffset = CGFloat(offsetAxis.axisMinimum)
                }
                if barData.yMax < 0 {
                    originYOffset = CGFloat(offsetAxis.axisMaximum)
                }         
            }
```
I think there is a chance when `offsetAxis._customAxisMin` is true, and offsetAxis._customAxisMax is false, you still need to set originYOffset. Like all negative values?

Source/Charts/Components/YAxis.swift
```diff
@@ -150,6 +150,17 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
+        // Make sure max is greater than min
+        if min > max {
```
liuxuan30: will take a look if this is the right place to fit. I wonder why there is min > max scenario, it should be solved earlier. but thanks for finding the cause.

liuxuan30: BTW, please change the code style, we put `{` a separate line in this library.

liuxuan30:
One more issue found:
data set: 
```swift
        let yVals = [
            BarChartDataEntry(x: 0, y: 10),
            BarChartDataEntry(x: 1, y: 20),
            BarChartDataEntry(x: 2, y: 30),
            BarChartDataEntry(x: 3, y: 15)
        ]
```
and enable `chartView.leftAxis.inverted = true`, it draws wrong:
image

only uses 
```swift
barRect.origin.y = top
barRect.size.height = bottom - top
``` solve the issue.

We also need to check highlight code as well. The highlight seems right, as forgot to consider the offset.

The accessibility has more bug then we think :(  I think we need add enough UTs for bar chart

File: Source/Charts/Components/YAxis.swift (hunk 2 above)
Comment (potato04) [reply Source/Charts/Components/YAxis.swift]:
``` swift
var min = _customAxisMin ? _axisMinimum : dataMin
var max = _customAxisMax ? _axisMaximum : dataMax
```
These lines of code caused min>max
Look example below：
```swift
chartView.leftAxis.axisMaximum = -40.0
let yVals = [
            BarChartDataEntry(x: 0, y: -10),
            BarChartDataEntry(x: 1, y: -20),
            BarChartDataEntry(x: 2, y: -30),
            BarChartDataEntry(x: 3, y: -15)
]
.....
var min = dataMin // dataMin = -30
var max = _axisMaxium // Manually set to 40
```

liuxuan30:
I start thinking if any better solution for the accessibility feature. The offset introduces more complexity, and seems destroyed the beauty of using `_matrixOffset`.

liuxuan30:
It seems the inverted bug starts from `barRect.origin.y = top + originYOffset`

I think we have to think about this.

Source/Charts/Renderers/BarChartRenderer.swift
potato04: After some debugging, it seems that the value of  _customAxisMax/_customAxisMin has nothing to do with calculating offset.

potato04:
Please review the new commits and advise your comments

Source/Charts/Components/YAxis.swift
```diff
@@ -150,6 +150,23 @@ open class YAxis: AxisBase
...
+        // Make sure max is greater than min
+        if min > max
+        {
+            if _customAxisMax && _customAxisMin
+            {
+                (min, max) = (max, min)
+            }
+            else if _customAxisMax && !_customAxisMin
+            {
+                min = max - 1
```
liuxuan30: what you think if we use `min = 0.5 * max` and `max = 1.5 * min`? @potato04 @jjatie @petester42 ?
clearly if min > max, and user only set one custom value of both, +/- 1 is not always the good value, for example with large values and decimals

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -108,6 +108,32 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
...
+            let offsetAxis = dataSet.axisDependency == .left ? offsetView.leftAxis : offsetView.rightAxis
+            if offsetAxis.axisMinimum > 0
+            {
+                heightOffset = CGFloat(offsetAxis.axisMinimum)
+            }
+            if offsetAxis.axisMaximum < 0
+            {
+                originYOffset = CGFloat(offsetAxis.axisMaximum)
+            }
+            if isInverted
```
liuxuan30: wow my skill is a little rusty, didn't realize this is a solution. I will test more

Source/Charts/Components/YAxis.swift
potato04: Couldn't agree more, may be we should update those codes too 
https://github.com/danielgindi/Charts/blob/a77a358fb0d8a89b6121d4cdcbc122cc42cce395/Source/Charts/Components/YAxis.swift#L170-L178

liuxuan30:
@potato04 Could you grant me push to your branch? Seems I cannot push a commit directly to your fix-barRectCaculation branch. So I use Github editor to add one more commit. Please have a review if my change is valid. The idea is to keep barRect calculation the same, move the offset earlier.

liuxuan30:
Items to act:
1. Check animation is fine (filed #3666 while testing, seems irrelevant to this PR). Partially done. Need double check
2. Check if highlight code need to change as well
3. Add UT for every case: + bars, +/- bars, - bars

@potato04 do you know how to add UT? if not, I will add it next week; but need be able to push to your branch.

potato04:
> Items to act:
> 
> 1. Check animation is fine (filed #3666 while testing, seems irrelevant to this PR). Partially done. Need double check
> 2. Check if highlight code need to change as well
> 3. Check Horizontal bar chart
> 4. Add UT for every case: + bars, +/- bars, - bars
> 
> @potato04 do you know how to add UT? if not, I will add it next week; but need be able to push to your branch.

i don't know how to grant this permission to you , but i have send a collaborator invitation to you.
do you mean that we should add more test case for `BarChartTests.swift` ? 
 i have not used FBSnapshotTestCase before, but i will try it

liuxuan30:
@potato04 it should be fine you invited me, I think I could push then. 
You just need to use carthage to install that or manually. FBSnapshotTestCase is used to generate images for testing.

No worry. I would add next week, if I didn't see you add. Take your time. You have helped a lot. Thx

Source/Charts/Components/YAxis.swift
potato04: When I was adding unit tests, I found that the correction for the maximum and minimum values were wrong.
Because we just modified `min` and `max` instead of `_axisMinimum` and `_axisMaximum`.
I think we should do that in the setter function of `axisMinimum`/`axisMaximum` if it is necessary.
https://github.com/danielgindi/Charts/blob/86ae25384960e764641650bd76ef20fb9148aa96/Source/Charts/Components/AxisBase.swift#L307-L319
@liuxuan30

Source/Charts/Components/YAxis.swift
```diff
@@ -150,6 +150,24 @@ open class YAxis: AxisBase
         var min = _customAxisMin ? _axisMinimum : dataMin
         var max = _customAxisMax ? _axisMaximum : dataMax
         
+        // Make sure max is greater than min
+        // Discussion: https://github.com/danielgindi/Charts/pull/3650#discussion_r221409991
+        if min > max
+        {
+            if _customAxisMax && _customAxisMin
```
jjatie: I think this would be easier to read as a switch statement:
```
switch (customAxisMax, customAxisMin) {
case (true, true):
    (min, max) = (max, min)

case (true, false):
    min = max * 0.5

case (false, true):
    max = min * 1.5

case (false, false):
    break
}
```

Source/Charts/Components/YAxis.swift
```diff
@@ -177,6 +178,9 @@ open class YAxis: AxisBase
             min = min - 1.0
         }
         
+        _axisMinimum = min
```
liuxuan30: I don't think we should change this? Because user set this specifically

potato04: If someone sets the max and the min value causes the min to be greater than max,  then we exchange their values.
So we should change _axisMinimum and _axisMaximum  too

potato04: Or we just warn the user that max must be greater than min instead of exchange their values

Source/Charts/Components/YAxis.swift
potato04: thanks!

Source/Charts/Components/YAxis.swift
potato04: OK, Then we have no need to change min and max
those codes can be removed
https://github.com/danielgindi/Charts/blob/9a9edcb44d0d0c00aaaaa371c03393fc624aeb4e/Source/Charts/Components/YAxis.swift#L155-L169

potato04: The reason the y-axis label disappearing in your previous comment https://github.com/danielgindi/Charts/pull/3650#issuecomment-424949606 is that you set chartView.leftAxis.axisMaximum = -40.0 specifically (the min of the data in this example is -30, so min>max ), what should we do about this situation?

liuxuan30:
BTW, I‘m looking at your UT images, how did you add them? It's a little weird to see line chart and pie chart test images being updated.

Source/Charts/Components/YAxis.swift
liuxuan30: oh my bad. min = max * 0.5  and max = min * 1.5 is wrong for negative values. 

If we use min = max - 1 and max = min + 1 scenario, the min max would be correct.

so either we stick to `min = max - 1 and max = min + 1` , or we use `min = max < 0 ? max * 1.5 : max * 0.5 and max = min < 0 ? min * 0.5 : min * 1.5`.

The idea is the same, make sure display properly, like in this case, .axisMaximum = -40.0, and min is -40 * 1.5 = -60, to display:
image

scripts/multi_devices_test.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# Script configuration
+
+ios_devices=(
```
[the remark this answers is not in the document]
potato04: I added this script to make it easier to generate images of all resolutions in record mode.

potato04:
> BTW, I‘m looking at your UT images, how did you add them? It's a little weird to see line chart and pie chart test images being updated.

e... i introduced some new device resolutions, then I turn on the record mode and start generating images...

Source/Charts/Components/YAxis.swift
potato04: Got it!

potato04:
Actually i am a bit skeptical about the need to test on devices with different resolutions.
Because the OS is the same, size of the chart is the same (480*350)
https://github.com/danielgindi/Charts/blob/fbec767b7dafacddc05eba1dd2563aa6dfe97e55/Tests/Charts/BarChartTests.swift#L37


In addition, I found that travis-ci only use iphone 7 (iOS 12, 375*667@2x) to build and test our framework 
So , for test case `testDefaultValues`, only image `testDefaultValues_iOS_375.0_667.0@2x.png` will be tested and others will not.
tim 20181009135754

scripts/multi_devices_test.sh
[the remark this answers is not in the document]
potato04: after some research, i think this file is no need any more, i will delete it

potato04:
1. remove unnecessary files
2. update code based on the review feedback

liuxuan30:
@potato04 I push one commit please have a look, and check if any other edge testing should be added, especially for issues found in this thread (exclude the animation issue)

@petester42 @jjatie please have a review about the unit test as well: some test images diff shows slight pixel shift, not sure why, but both old and new image seems drawing correctly

potato04:
@liuxuan30 Maybe we should add test case for all values are non-zero and equal?
btw, the script that automatically generate images is still a bit useful. ^_^

Tests/Charts/BarChartTests.swift
```diff
@@ -4,88 +4,341 @@ import FBSnapshotTestCase
...
+    {
+        let values: [Double] = [-8, -104, -81, -93, -52, -44, -97, -101, -75, -28,
+                                -76, -25, -20, -13, -52, -44, -57, -23, -45, -91,
+                                -99, -14, -84, -48, -40, -71, -106, -41, -45, -61]
+        return setupCustomValuesDataEntries(values: values)
+    }
+
+    func setupZeroValuesDataEntries() -> [ChartDataEntry]
+    {
+        let values: [Double] = [0,0,0,0,0,0,0,0,0,0,
```
jjatie: `let values = [Double](repeating: 0.0, count: 15)`

liuxuan30: @jjatie I found a weird issue, the old values count is 30, if I use your 15 or even count: 1, the UT still passes, which should fail. Can you run `testZeroValues` on your side to double check? 

I manually generated images to compare, the zero value labels count should have obvious difference, but the UT still pass.

potato04: tolerance?

liuxuan30: @potato04 could you try on your side? I'm not sure if it's tolerance issue, but it has obvious diff, I don't think so

potato04: Same result here,  but i found it's a tolerance issue, the diff of two images calculated by FBSnapshot is 17205/67200≈0.0256(<0.03)

potato04: below is the comparison code
https://github.com/facebookarchive/ios-snapshot-test-case/blob/3efc829601128f7d9306e8e58db04d32594b62ce/FBSnapshotTestCase/Categories/UIImage%2BCompare.m#L104-L121

liuxuan30: but the zero value labels is a lot from human eyes.. can't believe it will be very small difference..

Anyway, for zero value tests,  I think we could update the tolerance to make it more strict.

liuxuan30:
> > @potato04
> > > Check if highlight code need to change as well
> > > Check Horizontal bar chart
> > 
> > 
> > I suppose highlight code need to align, right?
> 
> Horizontal bar and highlight seem to be working fine

highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?

> Do you mean that we shall add unit tests?

I don't think we need to add so far.

potato04:
oh, i got you
 > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
no, we don't need to do that. Because it makes no difference to the UI

potato04:
Some issues that were fundamental to the calculation of bar rectangle had not yet been adequately addressed. I had to create a commit again...😓
In this commit, First, I moved the offset calculation code back into the loop because the offset of each bar may be different.
Second, replaced originYOffset and heightOffset with topOffset and bottomOffset and added comments to make it easier to understand the calculation process.

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -128,9 +129,84 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
...
+                 * |                       |//|         } bottomOffset = min - y   |                |  |   |//|          } bottomOffset = -max     |
+                 * |        y ->           +--+        <- bottom                   |        0 -> ---+--+---+--+-------  <- bottom                  |
+                 * |                                                               |                                                               |
+                 * +---------------------------------------------------------------+---------------------------------------------------------------+
+                 */
+                var topOffset: CGFloat = 0.0
+                var bottomOffset: CGFloat = 0.0
+                if let offsetView = dataProvider as? BarChartView
+                {
+                    let offsetAxis = dataSet.axisDependency == .left ? offsetView.leftAxis : offsetView.rightAxis
```
jjatie:
```suggestion
                    let offsetAxis = offsetView.getAxis(dataSet.axisDependency)
```

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
@potato04 more barRects to check:
1. bar border feature;
2. bar shadow feature; 

they are involved with rect cuculation. We have to check it as well.

liuxuan30:
@potato04 more barRects to check:
1. bar border feature;
2. bar shadow feature; 

they are involved with rect cuculation. We have to check it as well.

liuxuan30:
@potato04 more barRects to check:
1. bar border feature;
2. bar shadow feature; 

they are involved with rect cuculation. We have to check it as well.

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
@potato04 more barRects to check:
1. bar border feature;
2. bar shadow feature; 

they are involved with rect cuculation. We have to check it as well.

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

liuxuan30:
> oh, i got you
> 
> > highlight code is still drawing from zero line I suppose? Shall we need to align the barRect?
> 
> no, we don't need to do that. Because it makes no difference to the UI

I don't think so, because, I remember the old behavior is drawing from zero line, and we change it to draw from the `offset`. The old behavior is also not affecting UI. It's like an improvement, so I think we should align both rendering and highlight with the same logic. 

I checked the highlight code:
```swift
                    y1 = e.y
                    y2 = 0.0
                }
                
                prepareBarHighlight(x: e.x, y1: y1, y2: y2, barWidthHalf: barData.barWidth / 2.0, trans: trans, rect: &barRect)
```
it is similar, drawing from y2 = 0.0, using the clip trick to cut off unnecessary part of the highlight. Though it's working ok, but I'm kind of concerned if their barRect calculations are not aligned, there would be problem in the future?

@jjatie what you think?

potato04:
> The old behavior is also not affecting UI.
No, If you turn on voice over on your iPhone(guide: https://help.apple.com/iphone/10/#/iph3e2e2329), open chart-demo app and select one of the bars in the chart view, then a corresponding black border( VoiceOver Cursor) will appear, before i created this PR, the VoiceOver Cursor was oversized in some cases, because its size depends on the size of the bar being drawn instead of the size being clipped. please take a look at the image I posted at the top of this thread. The reason is the previous offset calculation did not cover all cases. So the main purpose of my PR is improve offset calculation and fix other problems caused by it.

jjatie:
@liuxuan30 @potato04 The bar border is fine (it's just a stroke setting, no calculation involved). The highlight seems fine. I'll need to take a closer look at the bar shadows later.

This PR is making me really want to try out Core Animation for our bar Charts.

liuxuan30:
@jjatie I would support try out CA as well. What you think that do we need to change highlight barRect calculation to align with this PR? I'm kind of unsure. I would agree what @potato04 said above. Highlight code works fine, but the way we calculate the barRect has changed significantly.

liuxuan30:
@jjatie this means you are ok to leave highlight untouched? then we could merge then

PR #3654, opened 2018-09-19
Make NSUIAccessibilityElement initializer public.

417-72KI:
### Issue Link :link:

When I create custom renderer with overriding LineChartRenderer, I can't create `NSUIAccessibilityElement` instance because its initializer is not public!
image

Furthermore, although `NSUIAccessibilityElement` is defined as open class, its subclass can create no instance.
image

### Goals :soccer:

To be able to create `NSUIAccessibilityElement` instance outside Charts.framework.

- Added `public` modifier to `NSUIAccessibilityElement.init(accessibilityContainer:)`.

### Testing Details :mag:

None, there are no specific tests.
This should not effect the snapshotting at all, and has no effect on the existing codebase.

### DIFF
--- a/Source/Charts/Utils/Platform+Accessibility.swift
+++ b/Source/Charts/Utils/Platform+Accessibility.swift
@@ -33,7 +33,7 @@ open class NSUIAccessibilityElement: UIAccessibilityElement
         }
     }
 
-    override init(accessibilityContainer container: Any)
+    override public init(accessibilityContainer container: Any)
     {
         // We can force unwrap since all chart views are subclasses of UIView
         containerView = container as! UIView
@@ -164,7 +164,7 @@ open class NSUIAccessibilityElement: NSAccessibilityElement
         }
     }
 
-    init(accessibilityContainer container: Any)
+    public init(accessibilityContainer container: Any)
     {
         // We can force unwrap since all chart views are subclasses of NSView
         containerView = container as! NSView


Source/Charts/Utils/Platform+Accessibility.swift
```diff
@@ -33,7 +33,7 @@ open class NSUIAccessibilityElement: UIAccessibilityElement
         }
     }
 
-    override init(accessibilityContainer container: Any)
```
liuxuan30: @jjatie not familiar with Accessibility. Is it best practice to make it public? If so, why the super class didn't do that...

liuxuan30:
@jjatie @petester42 seems good to me

Source/Charts/Utils/Platform+Accessibility.swift
```diff
@@ -33,7 +33,7 @@ open class NSUIAccessibilityElement: UIAccessibilityElement
...
+    override public init(accessibilityContainer container: Any)
```
liuxuan30: shall we change it to open? @petester42 @jjatie

jjatie: yes!

jjatie: Yep!

Source/Charts/Utils/Platform+Accessibility.swift
```diff
@@ -164,7 +164,7 @@ open class NSUIAccessibilityElement: NSAccessibilityElement
         }
     }
 
-    init(accessibilityContainer container: Any)
+    public init(accessibilityContainer container: Any)
```
jjatie: `open`

petester42:
I’ll look into it. Anything that is nsui prefixed is usually just for internal use since it’s platform dependent switch to have to same API. I’m hesitant to merge because it should be internal only.

Issue #3662, opened 2018-09-25
CombinedChartView's  candleData don't display in v3.2.0


Michael-Du:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

ℹ I use CombinedChartView, but candleData don't display

## What did you expect to happen?

ℹ CombinedChartView's  candleData display

## Charts Environment

**Charts version Number:*3.2.0*
**Xcode version:*10.0*
**Swift version:*4.2*
**Platform(s) running Charts:*iOS*
**macOS version running Xcode:*10.14*

## Demo Project

ℹI find this problem in https://github.com/danielgindi/Charts/tree/master/ChartsDemo-iOS/Swift

liuxuan30:
Are you able to reproduce with ChartsDemo?

Michael-Du:
> Are you able to reproduce with ChartsDemo?

You can find this bug in your ChartsDemo
![screen shot 2018-09-27 at 10 31 14 am](https://user-images.githubusercontent.com/17307261/46120176-0c4f8180-c241-11e8-842f-d1ef84256f12.png)

vcmilani:
I have same problem here, with example and my code!

boatt:
I have same problem here

petester42:
fixed by #3664

PR #3664, opened 2018-09-27
fix issue #3662

Michael-Du:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/3662
screen shot 2018-09-27 at 4 22 14 pm

When i used CombinedChartView, CandleStickChartRenderer never draw dataSet

screen shot 2018-09-27 at 4 25 51 pm

### Testing Details :mag:
I tested in example demo and my project

### DIFF
--- a/Source/Charts/Renderers/CandleStickChartRenderer.swift
+++ b/Source/Charts/Renderers/CandleStickChartRenderer.swift
@@ -62,8 +62,7 @@ open class CandleStickChartRenderer: LineScatterCandleRadarRenderer
     @objc open func drawDataSet(context: CGContext, dataSet: ICandleChartDataSet)
     {
         guard
-            let dataProvider = dataProvider,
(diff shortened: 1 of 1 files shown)


liuxuan30:
I remember it's used to be working. I need to check old commits what broke it.

liuxuan30:
@mathewa6 can you see this? Need you double check, we are good to merge this fix. But not sure why you cast dataProvider to candleView in the first place.

mathewa6:
@liuxuan30  Sorry about this: it is a mistake on my side. And it is safe to merge with regards to accessibility.

The reason the cast existed is that at one point, I was attempting to use a single inherited createAccessibilityElement() method. However, for some charts like Bar Charts, we needed to access the valueFormatter for axes labels, hence I used a cast on the container. There was no reason for it to be a guard or for it to exist in CandleStick charts.

PR #3681, opened 2018-10-07
Update ViewPortHandler.swift

no33jou:
fix y precision .

### DIFF
--- a/Source/Charts/Utils/ViewPortHandler.swift
+++ b/Source/Charts/Utils/ViewPortHandler.swift
@@ -434,7 +434,7 @@ open class ViewPortHandler: NSObject
     
     @objc open func isInBoundsTop(_ y: CGFloat) -> Bool
     {
-        return _contentRect.origin.y <= y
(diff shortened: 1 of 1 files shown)


Source/Charts/Utils/ViewPortHandler.swift
```diff
@@ -434,7 +434,7 @@ open class ViewPortHandler: NSObject
     
     @objc open func isInBoundsTop(_ y: CGFloat) -> Bool
     {
-        return _contentRect.origin.y <= y
+        return _contentRect.origin.y <= y + 1.0
```
jjatie: Why?

no33jou: when iOS9 ，_contentRect.origin.y is 30.0，y is 2.999999986，so The y value error is plus or minus 1

liuxuan30: I don't get it. what's the problem again?

liuxuan30:
no update since then. closing.

Issue #3694, opened 2018-10-17
No Support for Multiple Line Chart CircleHole colors


jhatin94:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## Feature Request

I tried to update the circleHoles of a Line Chart to have different colors similarly to how the actual circles can have different colors. However, it seems as if circleHole color can only be defined on a per data set basis.

## Charts Environment

**Charts version/Branch/Commit Number:**
v3.2.1 master branch
**Xcode version:**
Xcode 10.0
**Swift version:**
Swift 4.2

## Demo Project
The current ChartsDemo-iOS-Swift workspace

liuxuan30:
circleHoleColor is on dataSet basis meaning one line has a consistent circleHoleColor.

PurpleShrek:
You can make a point with the desired "circleHoleColor" as an image, and insert it into an array with yValues. Just don't forget to allow images to be shown.
in the example below, the last point is replaced with an image

        private var myChart: LineChartDataSet {
            let setChart = LineChartDataSet(entries: yValues, label: "")
            setChart.drawIconsEnabled = true //<------------------ this line allowed show images
            return setChart
        }

        private let yValues: [ChartDataEntry] = [
           ChartDataEntry(x: 0, y: 2200),
           ChartDataEntry(x: 1, y: 2200),
           ChartDataEntry(x: 2, y: 2100),
           ChartDataEntry(x: 3, y: 2200),
           ChartDataEntry(x: 4, y: 2300),
           ChartDataEntry(x: 5, y: 2400),
           ChartDataEntry(x: 6, y: 2300),
           ChartDataEntry(x: 7, y: 2200),
           ChartDataEntry(x: 8, y: 2500, icon: UIImage(named: "chartIcon")) //<----- append image
    ]

![IMG_9271](https://user-images.githubusercontent.com/63758848/125100222-8515e480-e0e1-11eb-89b8-79bc2f6bde4f.jpg)

PR #3695, opened 2018-10-17
Updated LineChart to Support an Array of Colors for CircleHoles - Issue #3694

jhatin94:
Fixes Issue #3694 

Changed the LineChartDataSet to support an array of colors for CircleHoles, similar to how CircleColors is done.

* Updated LineChartDataSet Interface to contain an array of colors for circleHoles and a function to get a circleHole color at a given index.
* Updated LineChartDataSet Implementation file to instantiate an empty NSUIColor array for circleHoleColors.
* Added implementation to get a color out of the new circleHoleColors array by index, returning the existing circleHoleColor property if the array is empty or the index is out of bounds.
* Updated the call for drawing a circleHole in the LineChartRenderer to use the new function to get the circleHoleColor by index instead of referencing the property

### Issue Link [:link:](https://github.com/danielgindi/Charts/issues/3694)
Fixes Issue #3694

### Goals :soccer:
The goal of this PR is to allow developers to customize their LineChart's individual circle holes

No new architectural changes
Now an array of circleHoleColors can be defined in the same way circleColors can be.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/LineChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/LineChartDataSet.swift
@@ -89,6 +89,19 @@ open class LineChartDataSet: LineRadarChartDataSet, ILineChartDataSet
         return circleColors[index]
     }
     
+    open var circleHoleColors = [NSUIColor]()
+    
+    /// - returns: The color at the given index of the DataSet's circle-hole-color array.
+    /// Performs a IndexOutOfBounds check by modulus.
+    open func getCircleHoleColor(atIndex index: Int) -> NSUIColor? {
+        let size = circleHoleColors.count
+        let index = index % size
+        guard size > 0, index < size else {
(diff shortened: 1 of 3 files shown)


liuxuan30:
may I ask the popularity of this? circleHoleColor is on data set basis, so each line has a consistent circleHoleColor. You want to one line have different circleHoleColor for each dot?

jhatin94:
So in my use-case, I would like the circle hole to appear empty (i.e. white/clear) but then be filled in to the color of the line at certain data points. Thus, I need at least 2 different colors for holes. I'm not sure how popular this is, but I figured I'd add this to be totally customizable since the lines between points and the points themselves in the data set can be changed individually.

liuxuan30:
@jjatie @petester42 what you think

jjatie:
This is the first time I've seen this request. I worry that adding this to the library encourages things like a different hole color for every hole on a line.

liuxuan30:
this is exactly a different hole color per hole for the same line (data set)

Issue #3578, opened 2018-07-24
A lot of properties not copies in `copyWithZone`


aelam:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

I copied dataSets 

## What did you expect to happen?

the properties should be the same with old dataSets

## What happened instead?

Some properties are not copied. the chart displays different 
For example 
`LineRadarChartDataSet` :  `lineWidth`
`ChartBaseDataSet`: `drawValuesEnabled`, `visible`

## Charts Environment

**Charts version/Branch/Commit Number:**
**Xcode version:*9.3*
**Swift version:*4.0*
**Platform(s) running Charts:**
**macOS version running Xcode:**

liuxuan30:
would you mind to find a PR?

aelam:
I don't get it , Find a PR?

I mean, the dataSet is not fully copied, here is my subclass implementation 

Take `BarChartDataSet` as an example 
There are two properties  `barBorderWidth`,  `barBorderColor`
they are not set in `copyWithZone` ,  
so my code is like this. that mean I would lose the origin  barBorderWidth
```
barDataSetA.barBorderWidth = 0.5
barDataSetB = barDataSetA.copy()
barDataSetA.barBorderWidth is not equal to  barDataSetB.barBorderWidth

```
here is my subclass of BarChartDataSet. when I copy it, the values would be same with the origin one 


```swift
public class MACDBarChartDataSet: PosNegBarChartDataSet, ISKBarChartDataSet {
    
    public func barBorderColor(atIndex index: Int) -> NSUIColor {
        let entry = self.entryForIndex(index)!
        if entry.y > 0 {
            return self.colors.first!
        } else {
            return self.colors[1]
        }
    }
    
    public override func copyWithZone(_ zone: NSZone?) -> AnyObject {
        let copy = super.copyWithZone(zone) as! PosNegBarChartDataSet
        copy.drawValuesEnabled = drawValuesEnabled
        copy.colors = colors
        copy.barBorderWidth = barBorderWidth
        copy.barBorderColor = barBorderColor
        
        return copy
    }

}

```

petester42:
I think he meant, send a pull request fixing this issue. This probably happened when new properties were added to the class but the person who added them forgot to put them in the copy method.

liuxuan30:
@aelam sorry, typo. should be 'fire a PR'

PR #3715, opened 2018-10-26
Add missing properties to copy(with:) methods

dstranz:
## Issue Link :link:
#3578 

### Goals :soccer:
* Add missing properties to copy(with:) methods

Add any missing property to copy(with:) methods

### Testing Details :mag:
I didn't add any tests cases, but if you think it will be valuable let me know.

#hacktoberfest #coinpaprika

### DIFF
--- a/Source/Charts/Data/Implementations/ChartBaseDataSet.swift
+++ b/Source/Charts/Data/Implementations/ChartBaseDataSet.swift
@@ -13,7 +13,7 @@ import Foundation
 import CoreGraphics
 
 
-open class ChartBaseDataSet: NSObject, IChartDataSet
+open class ChartBaseDataSet: NSObject, IChartDataSet, NSCopying
 {
     public required override init()
     {
@@ -411,7 +411,7 @@ open class ChartBaseDataSet: NSObject, IChartDataSet
     
     // MARK: - NSCopying
     
-    @objc open func copyWithZone(_ zone: NSZone?) -> AnyObject
+    open func copy(with zone: NSZone? = nil) -> Any 
     {
         let copy = type(of: self).init()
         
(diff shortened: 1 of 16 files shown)


liuxuan30:
thanks for this PR. I wonder if this is a complete list for now?

btw, I wonder if there is better ideas/solutions, because people would add new properties without remembering to add in copy(). Either could we tell whether we must add in copy(), or any smart way to auto add, or just by reviewers have to keep it in mind

@jjatie any thoughts

jjatie:
Really, the only way to keep on top of this is to do it by hand. One solution would be to implement the data model as Swift structs, but only expose them through classes. We get copy semantics and only have to add one property to the `copy` override i.e:

```
internal struct DataEntry: Equatable {
    let x: Int
    let y: Int
}

@objc
public class ChartDataEntry: NSObject {

    internal private(set) var entry: DataEntry

    var x: Int {
        get { return entry.x }
        set(x) { entry.x = x }
    }
    
    var y: Int {
        get { return entry.y }
        set(y) { entry.y = y }
    }

    func copy(with zone: NSZone? = nil) -> Any {
        let copy = type(of: self).init()
        copy.entry = entry
    }
}
```

It might be possible too, to create a protocol for this i.e.
```
protocol DataEntryProtocol {
    var x: { get set }
    var y: { get set }
}

extension DataEntry: DataEntryProtocol { }
extension ChartDataEntry: DataEntryProtocol { }
```

Then we could update our API to accept the protocol instead of the concrete type. This would allow us to start developing a nicer Swift API, and move objc compatibility into its own library for Charts 4.

dstranz:
@jjatie I've just add `NSCopying` conformance to base classes to have compiler validation & change method names.

liuxuan30:
I can't believe how `func copyWithZone(_:) -> AnyObject` is added, lol

Issue #3656, opened 2018-09-19
Incorrect behavior for AxisBase.axisMaxLabels


UweMeier:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

```
chartView.rightAxis.axisMinLabels = 10
chartView.rightAxis.axisMaxLabels = 15
```

## What did you expect to happen?

`axisMinLabels` to be 10
`axisMaxLabels` to be 15

## What happened instead?

`axisMinLabels == 15`
`axisMaxLabels == 15`

## Charts Environment

**Charts version/Branch/Commit Number:** 3.2.0
**Xcode version:** 10.0 (10A255)
**Swift version:** 4.2
**Platform(s) running Charts:** iOS 11

## Solution

looks like axisMaxLabels' didSet has some erroneous code inside:

https://github.com/danielgindi/Charts/blob/5087a04bbefc9f15995113954ce217bbfb445577/Source/Charts/Components/AxisBase.swift#L222-L224

Shouldn't it be `axisMaxLabels = axisMaxLabels > 0 ? axisMaxLabels : oldValue` in Line 223?

liuxuan30:
Seems a bug. Would you mind a fix and file a PR?

PR #3721, opened 2018-10-28
Fix wrong assignment to axisMaxLabels property

ggirotto:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/3656

### Goals :soccer:
Ensure that the max number of labels in the Y Axis is correct.

I'm just changing the wrong assignment.

### Testing Details :mag:
I didn't added any tests, but if necessary please let me know


liuxuan30:
good catch! Thanks

But why I remember there used to be a similar fix somewhere..

PR #3736, opened 2018-11-07
Fix legend offset bug for horizontal bar chart (Fixes #3301)

### LINKED ISSUE #3301: Horizontal Bar Chart legend issue
Have just updated to v3.0.5 (from 3.0.3) and one of my charts now has a slight overlap with the legend. Is this a known issue (I can see similar though different issues have been raised) and is there a setting that I could use to apply some padding/margin?

img_1672

SvenMuc:
### Issue Link :link:
Horizontal Bar Chart legend issue #3301

### Goals :soccer:
This PR fixes the legend offset issue in case of horizontal bar charts.

The guard conditions in the method `calculateLegendOffsets` in the file `HorizontalBarChartView.swift` were wrong.

### Testing Details :mag:
Tested in own app.

### DIFF
--- a/Source/Charts/Charts/HorizontalBarChartView.swift
+++ b/Source/Charts/Charts/HorizontalBarChartView.swift
@@ -39,7 +39,7 @@ open class HorizontalBarChartView: BarChartView
         guard
             let legend = _legend,
             legend.isEnabled,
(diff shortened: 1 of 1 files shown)


liuxuan30:
Thanks!

PR #3747, opened 2018-11-16
Implemented horizontal gradient

dmitryvirych:
Just implemented a horizontal gradient to the line chart.
Using: 
dataSet.gradientType = .horizontal

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -881,7 +881,15 @@ open class LineChartRenderer: LineRadarRenderer
         context.addPath(spline)
         context.replacePathWithStrokedPath()
         context.clip()
-        context.drawLinearGradient(gradient, start: gradientStart, end: gradientEnd, options: [])
+        
+        switch dataSet.gradientType {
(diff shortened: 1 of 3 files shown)


Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -881,7 +881,15 @@ open class LineChartRenderer: LineRadarRenderer
         context.addPath(spline)
         context.replacePathWithStrokedPath()
         context.clip()
-        context.drawLinearGradient(gradient, start: gradientStart, end: gradientEnd, options: [])
+        
+        switch dataSet.gradientType {
```
jjatie: Since the only thing changing is the `start` and `end`, the switch statement should be there i.e.
```
let start, end: CGPoint
switch dataSet.gradientType {
    case .horizontal: 
        start = ...
        end = ...

    case .vertical: ...
}
```

jjatie:
@liuxuan30 I'm not sure what you mean, this applies the line chart only.

liuxuan30:
oops, I guessed the feature from the title, I was thinking horizontal bar chart..

Issue #2180, opened 2017-02-21
Sharp edge on line graph


vsujan92:
Hello there, 
Actually I am using linegraph to plot around 144 points without zoom and scroll. Everything looks good, however in some cases there is a sharp line at some points.

Please refer the screenshot below.
<img width="723" alt="screen shot 2017-02-21 at 15 07 12" src="https://cloud.githubusercontent.com/assets/19152943/23158773/77733674-f848-11e6-8dfe-62f3b97dddd9.png">

Is there a way to avoid this?

liuxuan30:
can change line cap to rounded solve your issue?
```swift
    /// Line cap type, default is CGLineCap.Butt
    open var lineCapType = CGLineCap.butt
```

vsujan92:
Sorry, but that didn't work.

liuxuan30:
I can't be sure what's the issue right now.
From your screenshot, some has small, while some didn't:
![image](https://cloud.githubusercontent.com/assets/4375169/23387799/c64a1c4e-fd99-11e6-9fe0-45aa88f98e9e.png)
![image](https://cloud.githubusercontent.com/assets/4375169/23387801/c9301850-fd99-11e6-8d48-9705eb2774a9.png)
Can I reproduce it with ChartsDemo?
BTW, if you don't mind, horizontal bezier may solve the issue.

liuxuan30:
so before zoom(scaleX: 1, scaleY: 0, x: 0, y: 0), will it draw correctly?

vsujan92:
no the same issue occurs.

liuxuan30:
alright.. this is odd. but I have no idea right now.

dedenis:
Omg, Demo Project has this issue too 😨  And I have the same issue on my project... I dont know how to resolve it. 
See my attaches, please. I just run your ChartsDemo xcode-project, open "Multiple Line Chart", press Options->Toggle Filled and try zoom in/out vertically (you can see 3 orange sharp end). Can you fix it ? Seems like drawing algorithm bug (bcs on big zoom values this sharp parts of lines disappear)
![demo](https://cloud.githubusercontent.com/assets/18325968/24364154/114b7a64-133c-11e7-952d-8f153d3686ef.gif)

On my project it seems so:
![simulator screen shot 27 mar 2017 22 34 49](https://cloud.githubusercontent.com/assets/18325968/24364323/9a648d68-133c-11e7-8d00-9636308cc5fe.png)

[...]

PR #3764, opened 2018-12-01
Fixes sharp edges on the line chart

stokatyan:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/2180

### Goals :soccer:
This fixes the sharpness to some of the connected data points for the line charts.

When a line is drawn, vertices need to be extended to account for sharp angles, to avoid the long vertices the app needs to draw each line 1 at a time (the way it already does for line charts with more than 1 color for their lines).

**Note that existing users may have to update their line cap type to rounded** because when each line is drawn 1 at a time there can be very small open space between connected lines (this already exists for line charts with more than 1 color)

### Testing Details :mag:
**-----Before-----**
before fix

**-----After-----**
after fix

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -316,123 +316,63 @@ open class LineChartRenderer: LineRadarRenderer
         
         context.saveGState()
 
-        // more than 1 color
-        if dataSet.colors.count > 1
-        {
             if _lineSegments.count != pointsPerEntryPair
             {
                 // Allocate once in correct size
                 _lineSegments = [CGPoint](repeating: CGPoint(), count: pointsPerEntryPair)
             }
             
-            for j in stride(from: _xBounds.min, through: _xBounds.range + _xBounds.min, by: 1)
+        for j in stride(from: _xBounds.min, through: _xBounds.range + _xBounds.min, by: 1)
+        {
+            var e: ChartDataEntry! = dataSet.entryForIndex(j)
+            
+            if e == nil { continue }
+            
+            _lineSegments[0].x = CGFloat(e.x)
+            _lineSegments[0].y = CGFloat(e.y * phaseY)
+            
+            if j < _xBounds.max
             {
-                var e: ChartDataEntry! = dataSet.entryForIndex(j)
+                e = dataSet.entryForIndex(j + 1)
                 
-                if e == nil { continue }
+                if e == nil { break }
                 
-                _lineSegments[0].x = CGFloat(e.x)
-                _lineSegments[0].y = CGFloat(e.y * phaseY)
-                
-                if j < _xBounds.max
+                if isDrawSteppedEnabled
                 {
-                    e = dataSet.entryForIndex(j + 1)
-                    
-                    if e == nil { break }
-                    
-                    if isDrawSteppedEnabled
-                    {
-                        _lineSegments[1] = CGPoint(x: CGFloat(e.x), y: _lineSegments[0].y)
-                        _lineSegments[2] = _lineSegments[1]
-                        _lineSegments[3] = CGPoint(x: CGFloat(e.x), y: CGFloat(e.y * phaseY))
-                    }
-                    else
-                    {
-                        _lineSegments[1] = CGPoint(x: CGFloat(e.x), y: CGFloat(e.y * phaseY))
-                    }
+                    _lineSegments[1] = CGPoint(x: CGFloat(e.x), y: _lineSegments[0].y)
+                    _lineSegments[2] = _lineSegments[1]
+                    _lineSegments[3] = CGPoint(x: CGFloat(e.x), y: CGFloat(e.y * phaseY))
                 }
                 else
                 {
-                    _lineSegments[1] = _lineSegments[0]
+                    _lineSegments[1] = CGPoint(x: CGFloat(e.x), y: CGFloat(e.y * phaseY))
                 }
-
-                for i in 0..<_lineSegments.count
-                {
-                    _lineSegments[i] = _lineSegments[i].applying(valueToPixelMatrix)
-                }
-                
-                if (!viewPortHandler.isInBoundsRight(_lineSegments[0].x))
-                {
-                    break
-                }
-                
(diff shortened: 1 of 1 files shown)


liuxuan30:
wow thanks for solving the old issue. the change is much bigger than I thought. 
We need to spend time to review this.. please allow us to take a longer time than you could think.

and, could explain a little more? I'm confused why you said it's about single color or more colors?

stokatyan:
At the actual drawing level, the sharp edges are caused because there is only one line connecting the points. Charts uses only one line when there is only one color, and that causes the sharp edges.

The reason for sharp edges when there is one line is because the line is completely connected at every point, and has the same width at every point.  That makes sense for line segments and smooth (differentiable) curves.  However, you can’t really make sense of the width of a line at a vertex.

The picture attached shows what is going on.  The red lines are line segments, and the gray lines are continuous (not differentiable / not smooth) lines.

Line 1(c) uses 2 line segments with a rounded line cap to make it look like the lines are connected.

screen shot 2018-12-05 at 9 27 48 am

anton-filimonov:
@stokatyan @liuxuan30 seems like absolutely the same result could be reached by just calling `context.setLineJoin(.round)` and it seems to be less tricky. Another solution could be to set line join to .bevel, or set miterLimit to less then circleRadius/lineWidth. Actually, I suppose the best way to solve this is to provide some more properties to users to control these values (I mean add lineJoin and miterLimit properties to LineChartDataSet). 
P.S.: Can make pull request with this solution if you agree with me.

stokatyan:
@anton-filimonov I think that is a good suggestion.  Although, I think we should make that a separate PR and still get this one merged. 

The reason I would like that change to be a separate PR is because it will change functionality for existing users (having to consider how lines are joined).  The current PR didn't introduce any tricks, it just treats lines with one color the same way it was already treating lines with two colors.

liuxuan30:
@stokatyan so this PR implement like what 1(b), 2(b), and 1(c) did? My understanding is you use round line cap so 1(b) and 2(b) become 1(c)?

liuxuan30:
@petester42 @jjatie I would like to add this PR in next release soon. But Need your help to review this. This is a big change. Do you have time to do it?

liuxuan30:
@anton-filimonov are you saying your approach could solve the same issue as this PR did?
@stokatyan If it's true, I might consider not to merge this, as we only need one neat solution. If we merged this and later @anton-filimonov add more control and solve the same issue simultaneously, it's over kill then.

anton-filimonov:
@stokatyan you are right that this PR does not introduce any new tricks. My english is not good enough so probably I used incorrect sentences. I'm sorry if so. I think that the way the renderer handles case with multiple colors is tricky. Unfortunately this looks good only when `lineCap` is set to `.round`. For example, it looks like this for `lineCap == .butt`:
image
That's why I don't think that it's a good idea to use that same trick also for case with one color.
If we want to fix this concrete issue without adding changes for users - we can just set `lineJoin` to `.round` but probably not all the users will like this. What if there are people, who like current behaviour. We also have an option to add new properties and set default value for them to `.miter` because this way those people, for whom current behaviour is ok will not feel the difference, and those who want to change it - can do it with new properties

anton-filimonov:
@liuxuan30 Yes, such approach could fix the same issue by letting the user to decide how the line joins should be drawn. But this change will work only for one-color line.

jjatie:
@lixuan30 I’ll take a look today or tomorrow

stokatyan:
> @stokatyan so this PR implement like what 1(b), 2(b), and 1(c) did? My understanding is you use round line cap so 1(b) and 2(b) become 1(c)?

1(b) becomes 1(c), and 2(b) would behave similar to 1(c), but would have a different path.

@anton-filimonov @jjatie Got it, I will add the default `lineCap` parameter and include a performance test when I get a chance.  If I can't get it done during the week I should be able to make the updates over the weekend.

anton-filimonov:
@jjatie added another PR for this issue. It uses Core Graphics lineJoin parameter to solve the issue that's why to me it seems to be more natural. But that PR is exclusive with this one because this PR leaves no more line joins for single color line (for multicolor line this already is an issue as I showed in previous comments)

jjatie:
I disagree with #3797. It adds surface area to the API that doesn't completely solve the issue.

This PR unifies the drawing logic for single/multi-colour lines. Fixing the line cap to `.round` provides the only reasonable line to expect and resolves the issue. Assuming there is no meaningful performance hit in large data sets, this is definitely the preferred option.

liuxuan30:
any update on this? @stokatyan
I'm checking if we can merge this within this week or early next week

stokatyan:
@liuxuan30 I'm not sure what else is needed for this PR.  I don't see any existing performance tests to reference for the test that @jjatie is requesting.  If you want a performance test please be more specific about what is tested and what should be defined as a pass or fail.

I agree with jjatie regarding @anton-filimonov suggestions, so this PR looks done to me.

liuxuan30:
not sure why but this had a conflict now. please have a look

liuxuan30:
@jjatie 
>My only requested change is to implement a performance test to measure see if there is a meaningful difference between the two calculations.

Can you add the details for what @stokatyan asked?

stokatyan:
@liuxuan30 @jjatie is there anything left to do for this PR?

liuxuan30:
I have none. @jjatie

liuxuan30:
I have none. @jjatie

jjatie:
None

liuxuan30:
@stokatyan sorry to bring this up again. I was tracking an issue for line x axis animation in #4093, and after fix I found that the linear animation will have an unexpected dash line like below:
image

you can see at first it already have a dash line, with circle on the left while missing the right circle, indicating the animation is not reaching the second point, which is a bug.

I compared the linear code, it looks you choose to use
>context.strokeLineSegments(between: _lineSegments)

rather than addLine + strokePath way to draw linear. Do you remember why you do so? In old code base, strokeLineSegments is used for multiple colors only.

The old code is using addLine + strokePath  so no additional dash line during animating.

I suspect it's a bug in strokeLineSegments code path, but not sure if addLine + strokePath is causing the sharp edge.

Issue #3742, opened 2018-11-11
Default colors should adapt to macOS Mojave dark mode


sbenitezb:
## What did you do?

Rendered a chart under dark mode in Mojave.

## What did you expect to happen?

Labels, grids, etc, to show in white color.

## What happened instead?

Labels, grids, etc, show in black.

## Charts Environment

**Charts version/Branch/Commit Number:master**
**Xcode version:10.1**
**Swift version:4.2**
**Platform(s) running Charts:macOS**
**macOS version running Xcode:10.14.1 Mojave**

liuxuan30:
welcome a PR. We don't have the man power for the fancy features.

sbenitezb:
> welcome a PR. We don't have the man power for the fancy features.

Fancy? A chart library has to look good and legible. Black text over a dark background makes it useless for presenting information. You can’t read numbers or labels. I agree that you don’t have the man power, but this is no fancy feature, it’s just how things are now that we have a dark mode, like it or not. Mojave brought semantic colors. Charts should use that.

Anyway, I’ll see what I can do.

liuxuan30:
OK. Maybe not the right word.. cause this is for iOS mainly, not sure how often it's used in macOS.

Jean-Daniel:
Replacing NSUIColor.black by NSUIColor.labelColor for default text color should solve most of the color issues.

PR #3766, opened 2018-12-03
Switch from dark color to labelColor to fix darkmode issues.

Jean-Daniel:
### Issue Link :link:

#3742 

### Goals :soccer:

This change does not fix all Dark mode issues, but is (hopefully) a step in the right direction.

### DIFF
--- a/Source/Charts/Utils/Platform.swift
+++ b/Source/Charts/Utils/Platform.swift
@@ -139,6 +139,14 @@ public typealias NSUIGestureRecognizerState = UIGestureRecognizer.State
 		}
 	}
 
+	extension UIColor
+	{
+		static var labelColor : UIColor
+		{
+			return UIColor.black
+		}
+	}
+
 	extension UIView
     {
 		@objc final var nsuiGestureRecognizers: [NSUIGestureRecognizer]?
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -88,7 +88,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     @objc open var noDataFont: NSUIFont! = NSUIFont(name: "HelveticaNeue", size: 12.0)
     
     /// color of the no data text
-    @objc open var noDataTextColor: NSUIColor = NSUIColor.black
+    @objc open var noDataTextColor: NSUIColor = NSUIColor.labelColor
 
     /// alignment of the no data text
     open var noDataTextAlignment: NSTextAlignment = .left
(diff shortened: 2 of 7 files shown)


thierryH91200:
it's not exactly correct

I did not look at the problem for ios

```
extension NSColor
{
    static var labelColor : NSColor
    {
        var labelColor : NSColor
        {
            return NSColor.black
        }
        if #available(OSXApplicationExtension 10.13, *) {
            return NSColor(named: "Color") ?? NSColor.black
        } else {
            return NSColor.black
        }

    }
}
```

and also create "Assets.xcassets"

capture d ecran 2018-12-08 14 31 00


capture d ecran 2018-12-08 14 31 20

capture d ecran 2018-12-08 14 31 28

thierryH91200:
in fact the problem is the assignment of the kind

    @objc open var gridBackgroundColor = NSUIColor(red: 240/255.0, green: 240/255.0, blue: 240/255.0, alpha: 1.0)

rather
    @objc open var gridBackgroundColor = NSUIColor.gridColor
see
    open class var placeholderTextColor: NSColor { get }


AppKit
    open class var windowFrameTextColor: NSColor { get }
    open class var selectedMenuItemTextColor: NSColor { get }
    open class var alternateSelectedControlTextColor: NSColor { get }
    open class var headerTextColor: NSColor { get }

etc .....

UIKit is poor for the Color

liuxuan30:
@Jean-Daniel feedback to @thierryH91200?

Jean-Daniel:
If you want to use custom color and an asset catalog, you should use NSImage(named:bundle:), and place the asset catalog in the framework.

jjatie:
@thierryH91200 I'm not sure what your issue with this PR is. macOS does provide a default label color for standard and dark modes. For UIKit, this happens to be black, but it's reasonable to expect the API to exist in UIKit in the near future as well.

@Jean-Daniel please resolve conflicts

Source/Charts/Utils/Platform.swift
```diff
@@ -139,6 +139,14 @@ public typealias NSUIGestureRecognizerState = UIGestureRecognizer.State
 		}
 	}
 
+	extension UIColor
+	{
+		static var labelColor : UIColor
+		{
+			return UIColor.black
```
ChristianVinterly: I get that this is an old PR, but if UIColor.label is returned here, we can get Dark Mode support for labels on iOS as well, which would be nice.

jjatie: Simply removing the declaration will solve the issue for iOS 13+.
```
extension UIColor {
    @available(iOS, introduced: 9.0, obsoleted: 13.0)
    @available(tvOS, introduced: 9.0, obsoleted: 13.0)
    static var label: UIColor {
        return .black
    }
}
```

liuxuan30:
I'm going to look at the dark mode stuff recently. I know this is an old pr and there is a new one:https://github.com/danielgindi/Charts/issues/3742, which seems doing the same thing. 

from my comment https://github.com/danielgindi/Charts/pull/4171#issuecomment-540864782, we need to talk about a few things before. Can anyone reply? Thanks.

PR #3770, opened 2018-12-05
add chartScaled() call after double tap in BarLineChartViewBase

artemiusmk:
### Goals :soccer:

chartScaled() func from ChartViewDelegate should be called after performing double tap

I use chartScaled() func from ChartViewDelegate and found that it doesn't called after performing double tap. I see explanation before chartScaled() which says that it callbacks when the chart is scaled / zoomed via pinch zoom gesture, but zoom() func is called after double tap in BarLineChartViewBase, so I think it's logical to add this functionality. If you don't like this solution then maybe other possible way is to add chartZoomed() or chartDoubleTapped() to ChartViewDelegate


liuxuan30:
@jjatie thoughts? I'm ok

jjatie:
@liuxuan30 Yes, this is a good change.

On another note, we misnamed the delegate method. Keeping `scale` it should be `chartViewDidScale`, but I think really it should be `chartViewDidZoom`. I'll file a PR when I get a chance.

PR #3775, opened 2018-12-08
Turned gradient components and locations into constants

jjatie:
This turns the gradient components and locations into constants and keeps their initialization with their declarations


liuxuan30:
do we have UT for this? if not, I'm considering we can merge now

PR #3784, opened 2018-12-18
Corresponds to a problem that abnormally ends at the time of crack ex…

Kerubito0:
…pansion when data exists outside the graph display.

### DIFF
--- a/Source/Charts/Renderers/AxisRendererBase.swift
+++ b/Source/Charts/Renderers/AxisRendererBase.swift
@@ -90,6 +90,7 @@ open class AxisRendererBase: Renderer
     @objc open func computeAxisValues(min: Double, max: Double)
     {
         guard let axis = self.axis else { return }
+        if min.isNaN || max.isNaN { return }
         
         let yMin = min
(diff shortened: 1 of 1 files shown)


liuxuan30:
before merge, why there is NaN?

Source/Charts/Renderers/AxisRendererBase.swift
```diff
@@ -90,6 +90,7 @@ open class AxisRendererBase: Renderer
     @objc open func computeAxisValues(min: Double, max: Double)
     {
         guard let axis = self.axis else { return }
+        if min.isNaN || max.isNaN { return }
```
jjatie: Why are you not adding to the list of guarded conditions?

jjatie:
Is the NaN coming from your dataSet or an issue internal to the library?

liuxuan30:
usually we guarantee that no NaN will come into computeAxisValues. so we have to know where is the NaN generated.

jjatie:
Abandoned PR. Can be reopened when questions are answered.

PR #3797, opened 2018-12-21
Add lineJoinType and miterLimit parameters to LineChartDataSet (fixes #2180 )

### LINKED ISSUE #2180: Sharp edge on line graph

anton-filimonov:
let users fix sharp edges the way they want, keeping default behaviour same as current.

### Issue Link :link:
https://github.com/danielgindi/Charts/issues/2180

There is another PR that is made to solve same issue but it changes the default behaviour and actually also breaks line joins when `lineCapType` is not `.round`. This PR is another way to solve that problem, but it also not ideal - multicolored line will not use introduced `lineJoinType` property because of tricky way to draw that line (actually that tricky way leads to bad line when `lineCapType` is not `.round`)

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/LineChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/LineChartDataSet.swift
@@ -135,6 +135,12 @@ open class LineChartDataSet: LineRadarChartDataSet, ILineChartDataSet
     /// Line cap type, default is CGLineCap.Butt
     open var lineCapType = CGLineCap.butt
     
+    /// Line cap type, default is CGLineJoin.miter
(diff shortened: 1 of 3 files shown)


liuxuan30:
eh, is this PR exclusive to #3764? meaning we should only merge one of the two?

liuxuan30:
OK. Let's discuss in #3764 then to find a proper one.

PR #3799, opened 2018-12-25
Replaced Buffer class with equivalent type alia

jjatie:
### Goals :soccer:
The Buffer reference semantics were not taken advantage of, and seems only to exist for the name `buffer`
Needlessly added the extra `.rects` call when using it.

Makes usage a little cleaner


liuxuan30:
I remember `Buffer` is like a place holder. If we need add more vars we can add it into Buffer. like one day we might add colors into Buffer or CGPath array?
if we change it to alias, seems we are loosing the flexibility? Though I don't know if we will add something into Buffer :(

jjatie:
Looks like we were relying on the reference semantics too. I'll close the PR.

Issue #3803, opened 2018-12-29
chartTranslated is called when no translation happened


anton-filimonov:
[ * ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

 - Open sample app.
 - Go to Line Chart 1
 - Scale up a bit 
 - scroll to corner
 - try to scroll in direction of the same corner

## What did you expect to happen?

No scroll happens and no delegate methods are called

## What happened instead?

No scroll happens, but `chartTranslated` delegate method is called 

## Charts Environment

**Charts version:** 3.2.1
**Xcode version:** 10.1 (10B61)
**Swift version:** 4.2
**Platform(s) running Charts:** iOS
**macOS version running Xcode:** 10.14.2

## Demo Project

The library example project

PR #3804, opened 2018-12-29
Fix condition that is checked before `chartTranslated` delegate method call

anton-filimonov:
### Issue Link :link:
#3803 

### Goals :soccer:
eliminate not needed `chartTranslated` delegate method calls

Removed check if delegate is not nil because it's already done in next line by `?` operator
Added check that transformation matrix has really been changed before `chartTranslated` call

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -816,7 +816,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         
         matrix = _viewPortHandler.refresh(newMatrix: matrix, chart: self, invalidate: true)
         
-        if delegate !== nil
+        if matrix != originalMatrix
         {
             delegate?.chartTranslated?(self, dX: translation.x, dY: translation.y)
         }


Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -816,7 +816,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
         
         matrix = _viewPortHandler.refresh(newMatrix: matrix, chart: self, invalidate: true)
         
-        if delegate !== nil
+        if matrix != originalMatrix
```
liuxuan30: @jjatie struct in swift 4.2 is still copied on assigning right?
`refresh` will override `_touchMatrix` with `matrix`

@anton-filimonov in `refresh`->`limitTransAndScale()`, so `matrix` will be different anyway and makes `matrix != originalMatrix` always true to me?

anton-filimonov: @liuxuan30 yes, value types are copied on assigning, but that doesn't mean that comparing those values is useless because they are compared by value not by address. And since we are on the edge of available region we always get equal matrixes back from `refresh()->limitTransAndScale()`. And test shows that this change prevents calls of `chartTranslated` when no translation really happens.
The thing I really forgot to do is to do the same check when scale is performed. So I'll update this PR with that change soon

anton-filimonov: @liuxuan30 well I checked the situation with scale and same change there is not needed this much because delegate method calls there are limited with some other conditions. False calls are still possible there but it's only one call, for example when we zoom out the chart when it's already zoomed out to minimum (on the first call object think that we zoom in because no scale was made before)
So I'm not sure this change is needed. What do you think about it?

liuxuan30: @jjatie 
>A value type is a type whose value is copied when it’s assigned to a variable or constant, or when it’s passed to a function.

from https://docs.swift.org/swift-book/LanguageGuide/ClassesAndStructures.html

liuxuan30: @anton-filimonov sorry I made a mistake that I thought this is a func inside ChartsTransformer. Actually it's in `performPanChange()`, which is a pan gesture handler..

and in this case, when would matrix equals to originalMatrix for you to prevent the delegate get called? 

I checked the issue you posted:
```
scroll to edge
try to scroll in direction of the same edge
```
you are trying to scroll, which means a delegate should be called?

anton-filimonov: @liuxuan30 I'm trying to scroll, but no scroll happens because chart is on the edge. So chart is not translated, nothing changes and seems like `chartTranslated` shouldn't be called in such a case

liuxuan30: I see, so it's on the edge. But, if you are on the edge, would matrix equals to originalMatrix in this case? I just wonder they wouldn't be equal at any time.

anton-filimonov: Yes, matrix stays unchanged in this case. Actually we get new matrix, created in `limitTransAndScale` but that matrix is equal to previously used value

liuxuan30: I have tested on the edge, there are some `matrix == originalMatrix` cases, but there are also multiple false cases. from my understanding, only a few can be filtered out, is that what you are saying `So I'm not sure this change is needed.`?

anton-filimonov: `So I'm not sure this change is needed.` was about adding the same change to pinch gesture handling before calling `chartScaled`. There are two checks before all the work is done (`canZoomMoreX`, `canZoomMoreY`) and they eliminate almost all false calls, that's why I'm not sure this change is needed for that case.

anton-filimonov: It's much easier to reproduce the issue with false calls of `chartTranslated` when you scroll to the corner (I edited two last steps to reproduce the issue in #3803 ) so that translation is blocked in two directions and every scrolling step will cause no change to the transformation matrix

anton-filimonov: `but there are also multiple false cases mixing up at the same time` Probably some translation still occurs in this case. Maybe for example you reached the right edge but scrolled right-up.

liuxuan30: Yes I mean, adding this fix can filter a few calls, but generally, it won't prevent the delegate get called no matter where you are. You cannot rely on it to judge if you are on edge or something.

PR #3815, opened 2019-01-13
Add Collection conformances to ChartDataSet types

jjatie:
### Goals :soccer:
Allow more familiar API for collection types, the ability to implement more generic algorithms and simplify many existing algorithms by taking advantage of Collection conformances.

ChartDataSet adopts conformances:
MutableCollection
RandomAccessCollection
RangeReplaceableCollection

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -90,9 +90,9 @@ open class ChartDataSet: ChartBaseDataSet
         _xMax = -Double.greatestFiniteMagnitude
         _xMin = Double.greatestFiniteMagnitude
 
-        guard !values.isEmpty else { return }
+        guard !isEmpty else { return }
 
-        values.forEach { calcMinMax(entry: $0) }
+        forEach(calcMinMax)
     }
     
     open override func calcMinMaxY(fromX: Double, toX: Double)
@@ -100,17 +100,14 @@ open class ChartDataSet: ChartBaseDataSet
         _yMax = -Double.greatestFiniteMagnitude
         _yMin = Double.greatestFiniteMagnitude
 
-        guard !values.isEmpty else { return }
+        guard !isEmpty else { return }
         
         let indexFrom = entryIndex(x: fromX, closestToY: Double.nan, rounding: .down)
         let indexTo = entryIndex(x: toX, closestToY: Double.nan, rounding: .up)
         
         guard !(indexTo < indexFrom) else { return }
-        
-        (indexFrom...indexTo).forEach {
-            // only recalculate y
-            calcMinMaxY(entry: values[$0])
-        }
+        // only recalculate y
+        self[indexFrom...indexTo].forEach(calcMinMaxY)
     }
     
     @objc open func calcMinMaxX(entry e: ChartDataEntry)
@@ -160,17 +157,19 @@ open class ChartDataSet: ChartBaseDataSet
     open override var xMax: Double { return _xMax }
     
     /// The number of y-values this DataSet represents
-    open override var entryCount: Int { return values.count }
+    @available(*, deprecated, message: "Use `count` instead")
+    open override var entryCount: Int { return count }
     
     /// - Throws: out of bounds
     /// if `i` is out of bounds, it may throw an out-of-bounds exception
     /// - Returns: The entry object found at the given index (not x-value!)
+    @available(*, deprecated, message: "Use `subscript(index:)` instead.")
     open override func entryForIndex(_ i: Int) -> ChartDataEntry?
     {
-        guard i >= values.startIndex, i < values.endIndex else {
+        guard i >= startIndex, i < endIndex else {
             return nil
         }
-        return values[i]
+        return self[i]
     }
     
     /// - Parameters:
@@ -188,7 +187,7 @@ open class ChartDataSet: ChartBaseDataSet
         let index = entryIndex(x: xValue, closestToY: yValue, rounding: rounding)
         if index > -1
         {
-            return values[index]
+            return self[index]
         }
         return nil
     }
@@ -375,17 +374,10 @@ open class ChartDataSet: ChartBaseDataSet
     /// - Parameters:
     ///   - e: the entry to search for
     /// - Returns: The array-index of the specified entry
+    @available(*, deprecated, message: "Use `firstIndex(of:)` or `lastIndex(of:)`")
     open override func entryIndex(entry e: ChartDataEntry) -> Int
     {
-        for i in 0 ..< values.count
-        {
-            if values[i] === e
-            {
-                return i
-            }
-        }
-        
-        return -1
+        return firstIndex(of: e) ?? -1
     }
     
     /// Adds an Entry to the DataSet dynamically.
@@ -395,12 +387,13 @@ open class ChartDataSet: ChartBaseDataSet
     /// - Parameters:
     ///   - e: the entry to add
     /// - Returns: True
+    @available(*, deprecated, message: "Use `append(_:)` instead")
     open override func addEntry(_ e: ChartDataEntry) -> Bool
     {
         calcMinMax(entry: e)
 
         isIndirectValuesCall = true
-        values.append(e)
+        append(e)
         
         return true
     }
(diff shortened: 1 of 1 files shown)


liuxuan30:
should we use 4.0 branch? I suppose there would be conflicts later.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -517,3 +490,82 @@ open class ChartDataSet: ChartBaseDataSet
         return copy
     }
 }
+
+// MARK: MutableCollection
```
liuxuan30: if all above benefits come from all the protocols below, should we move the extension to top lines?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -517,3 +490,82 @@ open class ChartDataSet: ChartBaseDataSet
...
+    public func index(before: Index) -> Index {
+        return values.index(before: before)
+    }
+}
+
+// MARK: RangeReplaceableCollection
+extension ChartDataSet: RangeReplaceableCollection {
+    public func append(_ newElement: Element) {
+        isIndirectValuesCall = true
+        calcMinMax(entry: newElement)
```
liuxuan30: is it duplicated calls in `addEntry(_ e: ChartDataEntry) -> Bool`? There it already had a `calcMinMax`

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -517,3 +490,82 @@ open class ChartDataSet: ChartBaseDataSet
...
+    }
+
+    public subscript(position: Index) -> Element {
+        get {
+            // This is intentionally not a safe subscript to mirror
+            // the behaviour of the built in Swift Collection Types
+            return values[position]
+        }
+        set {
+            isIndirectValuesCall = true
```
liuxuan30: in `entryForIndex()` there is no `isIndirectValuesCall` and `calcMinMax`, why we need it here?
I'm a little concerned about `isIndirectValuesCall` here, forgetting to set it would be a trouble.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: This doesn't use `addEntry(_:)` so there is no duplicate call

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: `entryForIndex(_:)` is a get only method, note that the getter doesn't call `calcMinMax`. The setter allows arbitrary insertion of another element. If we insert an item we need to `calcMinMax`

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: No I mean in addEntry(), it will call `append(e)` I assume `append(e)` is this function call?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: so this is a new implementation right? do you remember what's the old function equivalent to this setter?

jjatie: We never had one. It would have been something like `insert(element:atIndex:)`

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: Correct, I will update addEntry to only call this and return `true`

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: It's convention to add conformance in extensions.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -417,87 +406,67 @@ open class ChartDataSet: ChartBaseDataSet
...
-                removed = true
-                break
-            }
-        }
-
-        notifyDataSetChanged()
-
-        return removed
+        guard let index = firstIndex(of: entry) else { return false }
+        _ = remove(at: index)
```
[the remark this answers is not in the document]
jjatie: This was on purpose. We always want to call `notifiyDataSetChanged` in our current implementation, so I go rid of it and let the property observer take car of it.

jjatie:
I think the deprecations should come in before Chart 4. There are already a ton of conflicts, so it doesn't matter to me

liuxuan30:
@jjatie while I'm pushing pod trunk, got several warnings from pod:
```
Validating podspec
 -> Charts (3.2.2)
    - NOTE  | [Charts/Core] xcodebuild:  note: Using new build system
    - NOTE  | [Charts/Core] xcodebuild:  note: Planning build
    - NOTE  | [Charts/Core] xcodebuild:  note: Constructing build description
    - NOTE  | [Charts/Core] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file. (in target 'App')
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift:22:37: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift:23:53: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:29:9: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:36:9: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:43:14: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:215:19: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:216:20: warning: 'values' is deprecated: Use `subscript(position:)` instead.
    - WARN  | [Charts/Core] xcodebuild:  Charts/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift:221:25: warning: 'values' is deprecated: Use `subscript(position:)` instead.
```

I'm using `--allow-warnings` for trunk push instead. Do you think we should fix those warnings?

jjatie:
Those are warnings intentionally generated by us. We won't be removing `values` but removing it's public exposure, so it's fine.

liuxuan30:
but if we don't have to remove `values`, we need find a way to get rid of the warnings

jjatie:
```
    @available(*, deprecated, message: "Use `subscript(position:)` instead.")
    @objc open var values: [ChartDataEntry]
        {
            get { return _values }
            set { _values = newValue }
        }
    internal var _values: [ChartDataEntry]
        {
        didSet
        {
            if isIndirectValuesCall {
                isIndirectValuesCall = false
                return
            }
            notifyDataSetChanged()
        }
    }
```
and then internally we only reference the internal var.

liuxuan30:
sure. let's do this with another PR

liuxuan30:
@jjatie I'm trying to fix the warnings, but I found there are demos using `values` like 
```objective-c
    BarChartDataSet *set = nil;
    if (_chartView.data.dataSetCount > 0)
    {
        set = (BarChartDataSet *)_chartView.data.dataSets[0];
        set.values = yValues;
        [_chartView.data notifyDataChanged];
        [_chartView notifyDataSetChanged];
    }
```
what you suggest? btw, why do we want to deprecate `values`?

liuxuan30:
@jjatie I didn't have a perfect idea yet, I'm thinking disable deprecate until we find a good way.
Either we add a new API or we don't encourage people to do so anymore. But I'm thinking keeping `values` public is fine

PR #3830, opened 2019-02-03
Refactored PieRadarChartViewBase.sampleVelocity

RomanPodymov:
### Goals :soccer:
As for me `PieRadarChartViewBase.sampleVelocity` is too cumbersome. Let's use more Swift standard library functions.

There is no new functionality, just refactoring. I hope that I used the best combination of Swift standard library functions of Xcode 10.1 to refactor `PieRadarChartViewBase.sampleVelocity`.

### DIFF
--- a/Source/Charts/Charts/PieRadarChartViewBase.swift
+++ b/Source/Charts/Charts/PieRadarChartViewBase.swift
@@ -673,7 +673,11 @@ open class PieRadarChartViewBase: ChartViewBase
         _velocitySamples.append(AngularVelocitySample(time: currentTime, angle: angleForPoint(x: touchLocation.x, y: touchLocation.y)))
         
         // Remove samples older than our sample time - 1 seconds        
-        _velocitySamples = Array(_velocitySamples.dropLast(2).drop { currentTime - $0.time > 1.0 })
+        _velocitySamples = Array(
+            _velocitySamples
+                .dropLast(2)
+                .drop { currentTime - $0.time > 1.0 }
+        )
     }
     
     private func calculateVelocity() -> CGFloat
--- a/Source/Charts/Charts/PieRadarChartViewBase.swift
+++ b/Source/Charts/Charts/PieRadarChartViewBase.swift
@@ -672,12 +672,10 @@ open class PieRadarChartViewBase: ChartViewBase
         
         _velocitySamples.append(AngularVelocitySample(time: currentTime, angle: angleForPoint(x: touchLocation.x, y: touchLocation.y)))
         
-        // Remove samples older than our sample time - 1 seconds        
-        _velocitySamples = Array(
-            _velocitySamples
-                .dropLast(2)
-                .drop { currentTime - $0.time > 1.0 }
-        )
+        // Remove samples older than our sample time - 1 seconds
+        if let indexOfNewerSample = _velocitySamples.dropLast(2).firstIndex(where: { currentTime - $0.time <= 1.0 }) {
+            _velocitySamples.removeSubrange(_velocitySamples.startIndex ..< indexOfNewerSample)
+        }
     }
     
     private func calculateVelocity() -> CGFloat
--- a/Source/Charts/Charts/PieRadarChartViewBase.swift
+++ b/Source/Charts/Charts/PieRadarChartViewBase.swift
@@ -672,23 +672,8 @@ open class PieRadarChartViewBase: ChartViewBase
         
         _velocitySamples.append(AngularVelocitySample(time: currentTime, angle: angleForPoint(x: touchLocation.x, y: touchLocation.y)))
         
-        // Remove samples older than our sample time - 1 seconds
-        var i = 0, count = _velocitySamples.count
-        while (i < count - 2)
-        {
-            if currentTime - _velocitySamples[i].time > 1.0
-            {
-                _velocitySamples.remove(at: 0)
-                i -= 1
-                count -= 1
-            }
-            else
-            {
-                break
-            }
-            
-            i += 1
-        }
+        // Remove samples older than our sample time - 1 seconds        
+        _velocitySamples = Array(_velocitySamples.dropLast(2).drop { currentTime - $0.time > 1.0 })
     }
     
     private func calculateVelocity() -> CGFloat


Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -672,23 +672,8 @@ open class PieRadarChartViewBase: ChartViewBase
...
-            }
-            else
-            {
-                break
-            }
-            
-            i += 1
-        }
+        // Remove samples older than our sample time - 1 seconds        
+        _velocitySamples = Array(_velocitySamples.dropLast(2).drop { currentTime - $0.time > 1.0 })
```
jjatie:
```
_velocitySamples = Array(
    _velocitySamples
        .dropLast(2)
        .drop { currentTime - $0.time > 1.0 }
)
```

RomanPodymov: Done.

RomanPodymov: @jjatie I found a bug in my pull request and I want to fix it.

jjatie: Go ahead. I don’t see the bug though.

jjatie: Oh, we don’t need to `dropLast(2)` only `drop(while:)`

jjatie: I'd prefer to use `removeAll(where:)` in this case. They are both O(n) operations, but we don't need to allocate a new `Array`. i.e.:
```
_velocitySamples.removeAll { currentTime - $0.time > 1.0 }
```
To be more true to the original implementation (although I don't think it's necessary, we can do:
```
_velocitySamples
    .dropLast()
    .removeAll { currentTime - $0.time > 1.0 }
```

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -672,23 +672,12 @@ open class PieRadarChartViewBase: ChartViewBase
...
-            }
-            else
-            {
-                break
-            }
-            
-            i += 1
-        }
+        // Remove samples older than our sample time - 1 seconds        
+        _velocitySamples = Array(
```
jjatie: I'd prefer to use `removeAll(where:)`. They are both O(n) operations, but we don't need to allocate a new `Array`. i.e.:
```
_velocitySamples.removeAll { currentTime - $0.time > 1.0 }
```
To be more true to the original implementation (although I don't think it's necessary, we can do:
```
_velocitySamples
    .dropLast()
    .removeAll { currentTime - $0.time > 1.0 }
```

RomanPodymov: Hello @jjatie,
Please check my solution again.

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -673,21 +673,8 @@ open class PieRadarChartViewBase: ChartViewBase
...
-                i -= 1
-                count -= 1
-            }
-            else
-            {
-                break
-            }
-            
-            i += 1
+        if let indexOfNewerSample = _velocitySamples.dropLast(2).firstIndex(where: { currentTime - $0.time <= 1.0 }) {
```
jjatie: 1. why are you dropping the last 2? You should only be dropping 1. (remember that the previous implementation uses `count - 2` because `count - 1` is the last accessible index, and they just added a sample for the current time so they can ignore the last item).

2. we don't need to call `drop` at all if we just simply add append the sample for the `currentTime` after removing the old samples.

3. I prefer the solution I provided for two reasons. The first is that it looks cleaner where there is no `if-let`. The second, is that it makes only one call to an `O(n)` operation (`removeAll(where:)`), whereas yours makes two (`firstIndex(where:)` and `removeSubrange`)

4. Finally, I would prefer the condition matches the comment above ( `{ $0.time < currentSample.time - 1 }`)

Putting it all together, and being more explicit about the sample creation (as is common in Swift):
```
    private func sampleVelocity(touchLocation: CGPoint)
    {
        let currentSample: AngularVelocitySample = {
            let time = CACurrentMediaTime()
            let angle = angleForPoint(x: touchLocation.x, y: touchLocation.y)
            return AngularVelocitySample(time: time, angle: angle)
        }()
        // Remove samples older than our sample time - 1 seconds
        _velocitySamples.removeAll { $0.time < currentSample.time - 1 }
        _velocitySamples.append(currentSample)
    }
```

@liuxuan30 ?

RomanPodymov: Hello @jjatie,
As I can see `_velocitySamples` is an array of ordered elements. Therefore there was `break` in `while (i < count - 2) {/* */}`. And that's why I want to use `firstIndex` + `removeSubrange` here to keep the same logic. Unfortunately there is no function like `mutating func dropped(while predicate: (Element) throws -> Bool)` to replace a composition of these 2 functions. 
Also `removeAll(where:)` loops through the whole array every time, it should not be.
I still cannot understand why there is `while (i < count - 2) {/* */}` but not `while (i < count - 1) {/* */}`. OK, the last item is a new `AngularVelocitySample` instance, but why should we skip the item that was the last before we added a new `AngularVelocitySample` instance to `_velocitySamples`? See this example (output is 0123, not 01234):
```
let a = [0, 1, 2, 3, 4]
let count = a.count
var i = 0
while i < count - 1 {
    print(a[i])
    i+=1
}
```

jjatie: I've decided I like your solution better. 2 things:
1. Regarding the `dropLast(2)` vs `count - 2`. 
```
let array = [1, 2, 3, 4]
array.dropLast(0) == array[...(array.count - 1)] // true
array.dropLast(1) == array[...(array.count - 2)] // true
```
In other words, looping through `array.dropLast(1)` is the same as looping through an array by indices where the indices are `0...(array.count - 2)`. In this case, we can avoid dropping anything by simply appending the new sample after we remove the elements. That turns your solution into:
```
private func sampleVelocity(touchLocation: CGPoint)
{
    let currentSample: AngularVelocitySample = {
        let time = CACurrentMediaTime()
        let angle = angleForPoint(x: touchLocation.x, y: touchLocation.y)
        return AngularVelocitySample(time: time, angle: angle)
    }()
    // Remove samples older than our sample time - 1 seconds
    if let indexOfNewerSample = _velocitySamples.firstIndex(where: { $0.time >= currentSample.time - 1 }) {
        _velocitySamples.removeSubrange(..<indexOfNewerSample)
    }
    _velocitySamples.append(currentSample)
}
```
2. I don't know what the more common case is, but I prefer using `lastIndex` here as it allows the code to read more like what we are trying to do (and match the comment). i.e.
```
if let indexOfNewerSample = _velocitySamples.lastIndex(where: { $0.time < newSample.time - 1 }) {
    _velocitySamples.removeSubrange(..<indexOfNewerSample)
}
```
Note the shorthand for the range expression.

RomanPodymov: Hello @jjatie 
Can we add some tests for `PieRadarChartViewBase.sampleVelocity`? Also `lastIndex` loops through the whole array, that's why we should use `firstIndex`.

jjatie: You're encouraged to add tests for this.

`lastIndex`on Bidirectional collections don't loop through the whole collection, they simply start at the end. Here's the implementation from the stdlib:
```
  @inlinable
  public func lastIndex(
    where predicate: (Element) throws -> Bool
  ) rethrows -> Index? {
    var i = endIndex
    while i != startIndex {
      formIndex(before: &i)
      if try predicate(self[i]) {
        return i
      }
    }
    return nil
  }
```

liuxuan30: @jjatie my question has been addressed by your comments already. but want to say we might should have a unit test for this function. Besides we should add more comments/explaining what the function tries to do and what's the condition. e.g. whether `_velocitySamples` is sorted or not and why `i < count - 2`.

jjatie:
#3883

PR #3846, opened 2019-02-16
Align `ChartLimit.LabelPosition` naming with `UIRectCorner`

jjatie:


Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -668,64 +668,40 @@ open class PieRadarChartViewBase: ChartViewBase
...
-        _velocitySamples.append(AngularVelocitySample(time: currentTime, angle: angleForPoint(x: touchLocation.x, y: touchLocation.y)))
-        
+        let currentSample: AngularVelocitySample = {
+            let time = CACurrentMediaTime()
+            let angle = angleForPoint(x: touchLocation.x, y: touchLocation.y)
+            return AngularVelocitySample(time: time, angle: angle)
+        }()
         // Remove samples older than our sample time - 1 seconds
-        var i = 0, count = _velocitySamples.count
-        while (i < count - 2)
```
liuxuan30: it's a bit obscure to say this change is equivalent. can you explain a little?
The main difference here is `while (i < count - 2)` so it seems the last one is not changed. but with the new change it will remove the last one if `$0.time < currentSample.time - 1`

Source/Charts/Renderers/XAxisRenderer.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
 import CoreGraphics
 
 #if !os(OSX)
-    import UIKit
```
liuxuan30: do we have a new swift standard here not to indent? within #if macro

Source/Charts/Charts/PieRadarChartViewBase.swift
jjatie: Oops! This is not supposed to be here, I will remove tonight. This is from #3830. Can you take a look at it? It's a good change.

Source/Charts/Renderers/XAxisRenderer.swift
jjatie: It's what Xcode auto-formatted. I can change back

Source/Charts/Charts/PieRadarChartViewBase.swift
```diff
@@ -686,21 +686,21 @@ open class PieRadarChartViewBase: ChartViewBase
             {
                 break
             }
-
```
liuxuan30: seeing multiple diffs here. what's the standard for us? empty line have space or not?

Source/Charts/Renderers/XAxisRenderer.swift
```diff
@@ -173,15 +173,15 @@ open class XAxisRenderer: AxisRendererBase
             else { return }
         
         #if os(OSX)
-            let paraStyle = NSParagraphStyle.default.mutableCopy() as! NSMutableParagraphStyle
+        let paraStyle = NSParagraphStyle.default.mutableCopy() as! NSMutableParagraphStyle
```
liuxuan30: seems like `import UIKit` within `#if` has indent but here it does not. We need to unify

Source/Charts/Charts/PieRadarChartViewBase.swift
jjatie: I'm sorry, I don't see what you're referring to. Do you mean removing the white space on blank lines?

Source/Charts/Renderers/XAxisRenderer.swift
jjatie: I will actually remove this one since it's the same API on macOS and iOS. In fact, `OSX` isn't even valid anymore.

Source/Charts/Charts/PieRadarChartViewBase.swift
liuxuan30: I mean sometimes the empty line has space while sometime it doesn't. which is the correct way for empty line?

jjatie: An empty line should have no white space

liuxuan30:
conflicts

Source/Charts/Charts/PieRadarChartViewBase.swift
liuxuan30: alright. then let's us this.

PR #3847, opened 2019-02-16
Renamed `values` to `entries` to reflect the property's type

jjatie:
Removed arbitrary setter access to `entries` to encourage use of `Collection` mechanics
Added `replaceEntries`

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -26,30 +26,30 @@ open class ChartDataSet: ChartBaseDataSet
 {
     public required init()
     {
-        values = []
+        entries = []
 
         super.init()
     }
     
     public override init(label: String?)
     {
-        values = []
+        entries = []
 
         super.init(label: label)
     }
     
-    @objc public init(values: [ChartDataEntry]?, label: String?)
+    @objc public init(entries: [ChartDataEntry]?, label: String?)
     {
-        self.values = values ?? []
+        self.entries = entries ?? []
 
         super.init(label: label)
 
         self.calcMinMax()
     }
     
-    @objc public convenience init(values: [ChartDataEntry]?)
+    @objc public convenience init(entries: [ChartDataEntry]?)
     {
-        self.init(values: values, label: "DataSet")
+        self.init(entries: entries, label: "DataSet")
     }
     
     // MARK: - Data functions and accessors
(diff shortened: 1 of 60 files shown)


liuxuan30:
ah I was leaving comments on the merged PR #3815 
This is what I thought about adding a new API. But should we rename to `entries` anyway? I'm thinking `values` is still a fine name.

jjatie:
There are a couple things misleading with `values`:
1. It it stores a collection of reference types, not value types.
2. The naming doesn't match the type it stores (it's not `ChartDataValue` it's `ChartDataEntry`.

jjatie:
Deprecating the direct manipulation of `values` allows us to better optimize algorithms down the road without storing state information. The functionality provided by collection conformance allows for much cleaner API to be written and exposes a much Swift-ier interface to working with data sets. It also allows for algorithms to be written against the `Collection` types (i.e binary search) and get new functionality added to the stdlib for free.

liuxuan30:
ok. but I think this is kind of a breaking change, if we merge this PR, it's kind of big change for users. 

There are two plans with 1 sec thought:
1. use 3.3 not 3.2.3 to indicate this is a 'major' update and merge this
or
2. We keep `values` for 3.x, and move this PR to 4.0.

I'm thinking **1** should be easier for everyone though our users have to make the change now.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -406,15 +409,14 @@ open class ChartDataSet: ChartBaseDataSet
     {
         calcMinMax(entry: e)
         
-        isIndirectValuesCall = true
```
liuxuan30: sorry but I have to ask why we can remove it here?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -26,51 +26,54 @@ open class ChartDataSet: ChartBaseDataSet
...
+    /// of `Collection` conformances.
+    ///
+    /// - Parameter entries: new entries to replace existing entries in the dataset
+    @objc
+    public func replaceEntries(_ entries: [ChartDataEntry]) {
+        self.entries = entries
+        notifyDataSetChanged()
     }
-    // TODO: Temporary fix for performance. Will be removed in 4.0
-    private var isIndirectValuesCall = false
```
liuxuan30: we can remove it now?

liuxuan30:
@jjatie any time you can take a quick look?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
jjatie: Since it's no longer directly accessible, the methods (mostly from collection conformance) call `calcMinMax` only if necessary. No longer need to make an exception for some calls.

liuxuan30:
@jjatie So I assume we merge it into master, and make a bigger release like 3.3

HellowjingQiu:
This attribute doesn't seem to be supported by Objective-C, I couldn't find it in the Charts-Swift.

jjatie:
Refer to our obj-c demo for an example

PR #3862, opened 2019-02-24
Reassess convenience initializers

jjatie:
The only data required to initialize an entry is an x and y value (or y in the case of Pie and Radar). All other data can easily be updated by initializing and assigning properties on the entry. Therefor, those initializers should be the ones marked as convenience. This also has the side effect of faster data set creation when only providing required data as those initializers no longer chain through to designated initializers.

Made initializer declarations consistent.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
@@ -37,21 +37,25 @@ open class BarChartDataEntry: ChartDataEntry
     }
     
     /// Constructor for normal bars (not stacked).
-    public override init(x: Double, y: Double, data: AnyObject?)
+    public convenience init(x: Double, y: Double, data: AnyObject?)
     {
-        super.init(x: x, y: y, data: data)
+        self.init(x: x, y: y)
+        self.data = data
     }
     
     /// Constructor for normal bars (not stacked).
-    public override init(x: Double, y: Double, icon: NSUIImage?)
+    public convenience init(x: Double, y: Double, icon: NSUIImage?)
     {
-        super.init(x: x, y: y, icon: icon)
+        self.init(x: x, y: y)
+        self.icon = icon
     }
     
     /// Constructor for normal bars (not stacked).
-    public override init(x: Double, y: Double, icon: NSUIImage?, data: AnyObject?)
+    public convenience init(x: Double, y: Double, icon: NSUIImage?, data: AnyObject?)
     {
-        super.init(x: x, y: y, icon: icon, data: data)
+        self.init(x: x, y: y)
+        self.icon = icon
+        self.data = data
     }
     
     /// Constructor for stacked bar entries.
@@ -62,32 +66,27 @@ open class BarChartDataEntry: ChartDataEntry
         calcPosNegSum()
         calcRanges()
     }
-        
+
     /// Constructor for stacked bar entries. One data object for whole stack
-    @objc public init(x: Double, yValues: [Double], data: AnyObject?)
+    @objc public convenience init(x: Double, yValues: [Double], icon: NSUIImage?)
     {
-        super.init(x: x, y: BarChartDataEntry.calcSum(values: yValues), data: data)
-        self._yVals = yValues
-        calcPosNegSum()
-        calcRanges()
+        self.init(x: x, yValues: yValues)
+        self.icon = icon
     }
-    
+
     /// Constructor for stacked bar entries. One data object for whole stack
-    @objc public init(x: Double, yValues: [Double], icon: NSUIImage?, data: AnyObject?)
+    @objc public convenience init(x: Double, yValues: [Double], data: AnyObject?)
     {
-        super.init(x: x, y: BarChartDataEntry.calcSum(values: yValues), icon: icon, data: data)
-        self._yVals = yValues
-        calcPosNegSum()
-        calcRanges()
+        self.init(x: x, yValues: yValues)
+        self.data = data
     }
-    
+
     /// Constructor for stacked bar entries. One data object for whole stack
-    @objc public init(x: Double, yValues: [Double], icon: NSUIImage?)
+    @objc public convenience init(x: Double, yValues: [Double], icon: NSUIImage?, data: AnyObject?)
     {
-        super.init(x: x, y: BarChartDataEntry.calcSum(values: yValues), icon: icon)
-        self._yVals = yValues
-        calcPosNegSum()
-        calcRanges()
+        self.init(x: x, yValues: yValues)
+        self.icon = icon
+        self.data = data
     }
     
     @objc open func sumBelow(stackIndex :Int) -> Double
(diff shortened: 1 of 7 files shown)


liuxuan30:
A quick note, would this impact ChartsRealm?

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -62,32 +66,27 @@ open class BarChartDataEntry: ChartDataEntry
...
+
     /// Constructor for stacked bar entries. One data object for whole stack
-    @objc public init(x: Double, yValues: [Double], icon: NSUIImage?)
+    @objc public convenience init(x: Double, yValues: [Double], icon: NSUIImage?, data: AnyObject?)
     {
-        super.init(x: x, y: BarChartDataEntry.calcSum(values: yValues), icon: icon)
-        self._yVals = yValues
-        calcPosNegSum()
-        calcRanges()
+        self.init(x: x, yValues: yValues)
```
liuxuan30: time is up for me today, so I cannot go deeper I suppose
@jjatie I want to be careful here, have you checked those function calls like `calcPosNegSum` and `calcRanges` that do not use `data` or other parameters?

The new change here seems changed the order that calls other calculation functions. e.g. in old IMP it will start calculate after all parameters are set, but with the new init(), the order is no longer there.

I will check back tomorrow

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -62,32 +66,27 @@ open class BarChartDataEntry: ChartDataEntry
         calcPosNegSum()
         calcRanges()
     }
-        
+
     /// Constructor for stacked bar entries. One data object for whole stack
-    @objc public init(x: Double, yValues: [Double], data: AnyObject?)
+    @objc public convenience init(x: Double, yValues: [Double], icon: NSUIImage?)
```
liuxuan30: could you double check here, the diff tool seems weird that the old function is `@objc public init(x: Double, yValues: [Double], data: AnyObject?)` and the new one is `@objc public convenience init(x: Double, yValues: [Double], icon: NSUIImage?)`

it's harder to say whether this is a expected one.

jjatie: I changed the order of the init that takes data and the one that takes an icon to make all of the files declare those initializers in the same order.

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
jjatie: Are you saying you want to assign `icon` and `data` before calling `self.init(x: x, yValues: yValues)`?
I'm okay with that.

jjatie: On second thought, it would be more efficient that way so I will change it.

jjatie:
Regarding ChartsRealm, I don't think it should affect anything. All the method signatures are the same and this only impacts `DataEntry` and ChartsRealm only makes new data sets.

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
liuxuan30: what I'm concerned is if we change order of the setter calls before or after the calculation functions, it might trigger weird or not easy found issues. so wanted to be careful about it. if we sure the properties are not used in the calculation, I'm fine with the new order. but it might take longer to have a review :)

jjatie: I will correct the order tonight

jjatie: There shouldn't be any affect here

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -122,30 +122,13 @@ open class BarChartDataEntry: ChartDataEntry
...
-        var sumNeg: Double = 0.0
-        var sumPos: Double = 0.0
-        
-        for f in _yVals
-        {
-            if f < 0.0
-            {
-                sumNeg += -f
+        (_negativeSum, _positiveSum) = _yVals?.reduce(into: (0,0)) { (result, y) in
+            if y < 0 {
```
liuxuan30: code style inconsistent

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -155,38 +138,34 @@ open class BarChartDataEntry: ChartDataEntry
     /// - Returns:
     @objc open func calcRanges()
     {
-        let values = yValues
-        if values?.isEmpty != false
-        {
+        guard let values = yValues, !values.isEmpty else {
```
liuxuan30: let's put else { return } in one line

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
```diff
@@ -156,38 +138,34 @@ open class BarChartDataEntry: ChartDataEntry
...
-        
+
         if _ranges == nil
         {
             _ranges = [Range]()
         }
         else
         {
-            _ranges?.removeAll()
+            _ranges!.removeAll()
```
liuxuan30: why we bother to force unwrap here? performance gain over `?`?

Source/Charts/Data/Implementations/Standard/PieChartDataEntry.swift
```diff
@@ -19,12 +19,21 @@ open class PieChartDataEntry: ChartDataEntry
         super.init()
     }
     
+
+    /// - Parameters:
+    ///   - value: The value on the y-axis
+    @objc public init(value: Double)
+    {
+        super.init(x: .nan, y: value)
```
liuxuan30: why we use NAN instead of 0?

Source/Charts/Data/Implementations/Standard/RadarChartDataEntry.swift
```diff
@@ -18,20 +18,21 @@ open class RadarChartDataEntry: ChartDataEntry
...
-    
+
     /// - Parameters:
     ///   - value: The value on the y-axis.
-    ///   - data: Spot for additional data this Entry represents.
-    @objc public init(value: Double, data: AnyObject?)
+    @objc public init(value: Double)
     {
-        super.init(x: 0.0, y: value, data: data)
+        super.init(x: .nan, y: value)
```
liuxuan30: same as pie chart. why .nan now

Source/Charts/Data/Implementations/Standard/BarChartDataEntry.swift
jjatie: Yes. `!` removes the runtime check for `nil`

Source/Charts/Data/Implementations/Standard/PieChartDataEntry.swift
jjatie: Because `x` is invalid for PieChartDataEntry and shouldn't be used. If it is being incorrectly used, this will more likely trap.

Source/Charts/Data/Implementations/Standard/RadarChartDataEntry.swift
jjatie: Because `x` is invalid for RadarChartDataEntry and shouldn't be used. If it is being incorrectly used, this will more likely trap.

PR #3863, opened 2019-02-24
Data as any

jjatie:
There is no need to restrict `ChartEntry.data` to `AnyObject`. This PR loosens that restriction. As a result we cannot compare `data`, though this should never have been the case in the first place.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
@@ -104,8 +104,7 @@ extension ChartDataEntry/*: Equatable*/ {
             return true
         }
 
-        return ((data == nil && object.data == nil) || (data?.isEqual(object.data) ?? false))
-            && y == object.y
+        return y == object.y
             && x == object.x
     }
 }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
@@ -39,7 +39,7 @@ open class ChartDataEntry: ChartDataEntryBase, NSCopying
     ///   - y: the y value (the actual value of the entry)
     ///   - data: Space for additional data this Entry represents.
     
-    @objc public convenience init(x: Double, y: Double, data: AnyObject?)
+    @objc public convenience init(x: Double, y: Double, data: Any?)
     {
         self.init(x: x, y: y)
         self.data = data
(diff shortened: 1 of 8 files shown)


Source/Charts/Data/Implementations/Standard/ChartDataEntry.swift
```diff
@@ -104,8 +104,7 @@ extension ChartDataEntry/*: Equatable*/ {
             return true
         }
 
-        return ((data == nil && object.data == nil) || (data?.isEqual(object.data) ?? false))
```
liuxuan30: I think we shouldn't drop `data` in compare?

liuxuan30: @jjatie

jjatie: I don't see how it is relevant to how two values are equal or not. If a consumer really needs to know if they are using the same object, the can use `===`

jjatie: From the `Equatable` documentation:
> Exposing nonvalue aspects of Equatable types other than class identity is discouraged

liuxuan30:
conflicts :)

jjatie:
Fixed

danielgindi:
Sorry for the delay, had been busy. I'm working now on syncing android<->ios for latest changes.

First, the `AnyObject` was the only thing that made sense in the first version of Swift. Only lately they changed ObjC id to be mapped to `Any` instead of `AnyObject`. And we do keep ObjC compatibility.

I have a small question here - why did you drop the `data` comparison completely?
As two entries pointing to a different data - mean different entries. So why not add `data === object.data`?

jjatie:
To support native Swift value types, `Any` is necessary and there is no `===` operator for `Any`. In order to properly support this we would need a type erased `AnyEquatable` type, but the closest thing to that out of the box is `AnyHashable`.

I view `data` more of an attachment to make it more convenient to add meta data to the graph about a given point. It makes sense to me that two points are equal IFF the values it is comprised of are equal, where metadata (i.e. `data`) is irrelevant.

danielgindi:
@jjatie The difference between equality and identity is clear...

But yes there could be cases where a user might consider two equivalent points with different metadata as different. However in such a case where the user actually added two equivalent coordinates to the chart - he will experience a subtle bug anyway, as you don't even know which of them you're viewing. 

I think we should leave it as it is for now, and review when someone complains.
(I can guess the answer would be "don't do this" or "use separate datasets").
Think came up from the Android code I was reviewing, just to make sure we are keeping the expected functionality in the same ballpark.

PR #3864, opened 2019-02-24
Replace AnyObject with Any

jjatie:
This is in line with how Objc interfaces are now imported

### DIFF
--- a/Source/Charts/Utils/Platform.swift
+++ b/Source/Charts/Utils/Platform.swift
@@ -237,7 +237,7 @@ public typealias NSUIGestureRecognizerState = UIGestureRecognizer.State
         private var displayLink: CVDisplayLink?
         private var _timestamp: CFTimeInterval = 0.0
         
-        private weak var _target: Any?
+        private weak var _target: AnyObject?
         private var _selector: Selector
         
         public var timestamp: CFTimeInterval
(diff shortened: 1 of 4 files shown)


Source/Charts/Utils/Platform.swift
```diff
@@ -245,9 +245,9 @@ public typealias NSUIGestureRecognizerState = UIGestureRecognizer.State
             return _timestamp
         }
 
-		init(target: AnyObject, selector: Selector)
+		init(target: Any, selector: Selector)
```
liuxuan30: despite the weird indent here by github, why we use Any and again  cast to AnyObject later
I assume using AnyObject is fine

liuxuan30:
@jjatie why would we change to Any here? I checked the doc:
>You use AnyObject when you need the flexibility of an untyped object or when you use bridged Objective-C methods and properties that return an untyped result. AnyObject can be used as the concrete type for an instance of any class, class type, or class-only protocol.

Source/Charts/Utils/Platform.swift
jjatie: This is to align with how ObjC APIs are imported into Swift. More info here: https://developer.apple.com/swift/blog/?id=39

We were only using `AnyObject` in the initializer because of weird objc interop.

liuxuan30: oh no. I'm using Dash and cannot even found `Any` type. this kind of information is not easily to be aware of in my daily work..

PR #3874, opened 2019-02-28
Minor updates for Swift 5

jjatie:
Need FBSnapshotTestCase to be updated
Travis script needs to be updated


liuxuan30:
I sincerely hope we will have an ABI stable swift asap 😂

jjatie:
Module stability isn't until 5.1

alanzeino:
> Need FBSnapshotTestCase to be updated
> Travis script needs to be updated

Heads up I'm the maintainer of FBSnapshotTestCase (now iOSSnapshotTestCase) and I added support for full Carthage releases in https://github.com/uber/ios-snapshot-test-case/pull/68 so you don't need to use git refs.

liuxuan30:
since swift 5 is released today, adding this PR to 3.3

liuxuan30:
@jjatie  any idea when we can merge this?

RAGOpoR:
I still waiting please merge it :)

jjatie:
@liuxuan30 Code is all updated now. Tests are failing because they can't load the images. I don't have time to look into this soon, can you please take a look?

jjatie:
@alanzeino Thanks for the full Carthage support! We should be doing this soon too.

liuxuan30:
@jjatie 
>error: SWIFT_VERSION '5.0' is unsupported, supported versions are: 3.0, 4.0, 4.2. (in target 'ChartsDemo-iOS')

just like I guess, it's not using swift 5 compiler. I could take a look tomorrow, have to run errands today.

jjatie:
@liuxuan30 If that's Travis, I just bumped the Xcode version. If that's your machine, make sure you've installed Xcode 10.2

liuxuan30:
I have checked the failures, basically two issues:
1. the image file name pattern has been changed, such as from `testDefaultValues_iOS_375.0_667.0@2x.png` to `testDefaultValues_iOS_375_0_667_0@2x.png`

note `375.0_667.0` to `375_0_667_0`

2. there is a `ReferenceImages_32` added, while we only have `ReferenceImages_64` folders.

Since this seems caused by the test case framework update itself, I'm going to ask @alanzeino for a quick help: is it configurable so we don't have to change the name pattern? Can we only have arm64 images? I don't think we need a 32bit arch.

samjarman:
How far away is this @liuxuan30? :)

alanzeino:
@liuxuan30 sorry Xuan but I can't provide help with that library anymore as I no longer work at Uber

liuxuan30:
> @liuxuan30 sorry Xuan but I can't provide help with that library anymore as I no longer work at Uber

😂 fine, I can take a look later. but you don't know the answer of my questions at all? I assume should be easy for you

liuxuan30:
> How far away is this @liuxuan30? :)
I think code is ready, the CI fails due to unable to find the image path. I'm evaluating what's the lowest cost to fix. Bottom line is I can manually rename the files and add 32bit images, but I don't want to in the first place, as this is not our fault.

liuxuan30:
I'm hooking into the testcase framework, and find out the file name logic is changed to
```objective-c
    NSMutableCharacterSet *invalidCharacters = [NSMutableCharacterSet new];
    [invalidCharacters formUnionWithCharacterSet:[NSCharacterSet whitespaceCharacterSet]];
    [invalidCharacters formUnionWithCharacterSet:[NSCharacterSet punctuationCharacterSet]];
    NSArray *validComponents = [fileName componentsSeparatedByCharactersInSet:invalidCharacters];
    fileName = [validComponents componentsJoinedByString:@"_"];
    
    if ((option & FBSnapshotTestCaseFileNameIncludeOptionScreenScale) == FBSnapshotTestCaseFileNameIncludeOptionScreenScale) {
        CGFloat screenScale = [[UIScreen mainScreen] scale];
        fileName = [fileName stringByAppendingFormat:@"@%.fx", screenScale];
    }
```
so basically it replaces every symbol with `_`, even the `.` symbol. This is a breaking change I suppose. So looks like I have to rename the images anyway.

the 32bit images can be ignored, but we are calling a macro function that will use default suffixes:
```objective-c
NSOrderedSet *FBSnapshotTestCaseDefaultSuffixes(void)
{
    NSMutableOrderedSet *suffixesSet = [[NSMutableOrderedSet alloc] init];
    [suffixesSet addObject:@"_32"];
    [suffixesSet addObject:@"_64"];
    if (FBSnapshotTestCaseIs64Bit()) {
        return [suffixesSet reversedOrderedSet];
    }
    return [suffixesSet copy];
}
```
So seems if we use other functions or env variables, we can ignore 32bit. But it took more time to choose a proper way to do it.

liuxuan30:
I'm seeing different images. Holding this PR. will manual check

liuxuan30:
changing tolerance to 0.001 indicate multiple  test failures. Looks like we have troule. Some recent changes break the boundary check:
for example, bar chart test testPositiveValuesWithCustomAxisMaximum:
image
@jjatie I suggest we fix the bugs in this PR or another asap. What you think? another PR in order to quickly merge this, or we fix all in this PR?

liuxuan30:
I will try to fix some obvious ones in this PR first.

liuxuan30:
it turned out that the XBounds() that leads to test failures with tolerance 0.001. After using my way to fix, I found that there is a conflict in master, and it already fixed in master. So I merged master into this branch and rebased on master and force push again. Not interfering the commit log and makes it clear.

seems safe to merge now.

Note: have to update tvOS test image for `testPositiveValuesWithCustomAxisMaximum`. After changing tolerance to 0.001, it fails while the same iOS test pass.
I looked into the image diff, it shows the intersect of bar top edge and axis line at value 50 has diff, but I suppose the image is the same. safe to merge.

liuxuan30:
@jjatie please review my commits. and welcome others to give feedback.

PR #3883, opened 2019-03-05
Velocity samples calculation

jjatie:
Leveraging the Swift stdlib to implement calculations around the `PieRadarChartViewBase.velocitySamples`


jjatie:
Aide from #3830

liuxuan30:
what we do about these two PRs, which first?

PR #3884, opened 2019-03-05
Removed unnecessary #if statements and unified style to align with Xc…

jjatie:
Removed unnecessary #if statements and unified style to align with Xcode's indentation


liuxuan30:
why it's no longer needed for OSX and tvOS?

jjatie:
Those `#if`s have not been used since I’ve started on the project. I left the ones that are.

liuxuan30:
that's weird. because those macro are used for NSUI I suppose at that time? anyway.

PR #3892, opened 2019-03-09
Migrating to built-in algorithms

jjatie:
### Goals
Inspired by WWDC'18 Embracing Algorithms
Charts relies heavily on custom algorithm implementations. Many of these are common algorithms and should instead rely on the built in implementations. In this PR I have updated the obvious candidates.

I have intentionally omitted the longer algorithms as some A) are unique algorithms and B) can become much simpler with some additional changes in the framework (i.e. Collection conformances introduced in 4.0.0).

### DIFF
--- a/Source/Charts/Charts/RadarChartView.swift
+++ b/Source/Charts/Charts/RadarChartView.swift
@@ -168,21 +168,9 @@ open class RadarChartView: PieRadarChartViewBase
         let sliceAngle = self.sliceAngle
         
         let max = _data?.maxEntryCountSet?.entryCount ?? 0
-        
-        var index = 0
-        
-        for i in 0..<max
-        {
-            let referenceAngle = sliceAngle * CGFloat(i + 1) - sliceAngle / 2.0
-            
-            if referenceAngle > a
-            {
-                index = i
-                break
-            }
-        }
-        
-        return index
+        return (0..<max).firstIndex {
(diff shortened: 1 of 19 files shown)


Source/Charts/Charts/RadarChartView.swift
```diff
@@ -168,21 +168,9 @@ open class RadarChartView: PieRadarChartViewBase
...
-            {
-                index = i
-                break
-            }
-        }
-        
-        return index
+        return (0..<max).firstIndex {
+            sliceAngle * CGFloat($0 + 1) - sliceAngle / 2.0 > a
+        } ?? max
```
liuxuan30, line 173: @jjatie while I am working merging 4.0, I found this imp is different than the old one.

The old one would return 0 if no match, but the new logic returns max.

Is this a mistake or intended?

jjatie: It's the same logic.

jjatie: My mistake. Should be `?? 0`
It might be a bit more readable to write it as:
```
return (0..<max).firstIndex {
    let referenceAngle = sliceAngle * CGFloat($0 + 1) - sliceAngle / 2.0
    return referenceAngle > a
} ?? 0
```

liuxuan30: I have changed to ?? 0.
I will let you know when I resolve all conflicts and we could continue from there.

PR #3897, opened 2019-03-12
Add leading and trailing entries for cubic lines

gemedet:
### Goals :soccer:
Since cubic lines require four points, this PR allows extra control points to be specified, which are used to draw the curves at the beginning and end of the line. Essentially, they act as `prev` for the first data point and `next` for the final data point.

This is particularly useful when we want to draw a curved line between only two data points.

Two new properties have been added to the line chart protocol: `cubicLeadingEntry` and `cubicTrailingEntry`. Both are optional, and when set to nil they have no effect.

### Testing Details :mag:
A new test has been added, which uses both properties.


liuxuan30:
sorry for the delay. Do you get screenshot to show what's the main purpose?

gemedet:
@liuxuan30 sure thing, here's a simple example with four data points:
```swift
let values = [
    ChartDataEntry(x: 0, y: 0),
    ChartDataEntry(x: 1, y: 1),
    ChartDataEntry(x: 2, y: 1),
    ChartDataEntry(x: 3, y: 0)
]

let set1 = LineChartDataSet(values: values, label: "")
let set2 = LineChartDataSet(values: [values[1], values[2]], label: "")
```
chart

the left side is how it currently works, and the right side is after adding:
```swift
set2.cubicLeadingValue = values[0]
set2.cubicTrailingValue = values[3]
```
also, since `values` has now been renamed to `entries`, maybe this PR should use `cubicLeadingEntry` and `cubicTrailingEntry`?

liuxuan30:
so this is a PR about gradient for cubic lines?

gemedet:
@liuxuan30 this PR is about controlling the shape of the beginning and end of the line. on the left side of the screenshot you can see that `set2` is straight, while on the right side it's properly curved

PR #3854, opened 2019-02-20
HorizontalBarChar value label offset calculation 

chaaarly:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/3850

### Goals :soccer:
Improve HorizontalBarChart offset calculation for negative value labels.

### Implementation Details :construction:
Just subtract the bar width from negative offset calculation. 

### Testing Details :mag:
Tested with positive and negative value set, value icons, drawValueAboveBarEnabled (true/false)

PR #3906, opened 2019-03-20
Fix horizontal bar chart not drawing values and add unit tests

liuxuan30:
regards https://github.com/danielgindi/Charts/pull/3854
I found the condition seems wrong and duplicated.
add unit tests for horizontal bar charts


Tests/Charts/HorizontalBarChartTests.swift
```diff
@@ -0,0 +1,147 @@
...
+    {
+        var entries: [ChartDataEntry] = Array()
+        for (i, value) in values.enumerated()
+        {
+            entries.append(BarChartDataEntry(x: Double(i), y: value, icon: UIImage(named: "icon", in: Bundle(for: self.classForCoder), compatibleWith: nil)))
+        }
+        return entries
+    }
+
+    func setupStackedvaluesDataEntries() -> [ChartDataEntry]
```
xkampotx: setupStackedValuesDataEntries ;)

liuxuan30: thanks. fixed

liuxuan30:
one thing to mention is that I couldn't subclass BarChartTests for HorizontalBarCharts, always return failures. so I have to duplicate code for now.  
maybe @alanzeino can give some insights?

alanzeino:
> one thing to mention is that I couldn't subclass BarChartTests for HorizontalBarCharts, always return failures no matter it's in record mode or not, the record mode is even not creating images on my side when subclassing from BarChartTests. so I have to duplicate test code for now.
> maybe @alanzeino can give some insights?

I don't think subclasses test classes is a good idea, you might want to look into the library with some breakpoints to figure out what's going on, for example in `FBSnapshotTestController.m`.

liuxuan30:
@alanzeino noted. thanks

Source/Charts/Renderers/HorizontalBarChartRenderer.swift
```diff
@@ -332,7 +332,7 @@ open class HorizontalBarChartRenderer: BarChartRenderer
             {
                 guard let dataSet = dataSets[dataSetIndex] as? IBarChartDataSet else { continue }
                 
-                if !shouldDrawValues(forDataSet: dataSet) || !(dataSet.isDrawIconsEnabled && dataSet.isVisible)
+                if !shouldDrawValues(forDataSet: dataSet)
```
jjatie: This should be a guard statement:
```
guard shouldDrawValues(for: dataSet),
    dataSet.isDrawIconsEnabled,
    dataSet.isVisible else 
{
    continue
}
```

liuxuan30: I have updated all related code to guard statement, have a look please.

PR #3973, opened 2019-04-26
For #3917. make init(label: String?) convenient initializer

### LINKED ISSUE #3917: Crash at BarChartDataSet as! IBarChartDataSet
## What did you do?

Create BarChartData with BarChartDataSet with BarChartDataEntry.

See bar chart.

## What happened instead?

App crashed.
Zrzut ekranu 2019-03-27 o 11 21 53

"Could not cast value of type 'Charts.ChartDataSet' (0x108ede670) to 'Charts.IBarChartDataSet' (0x108e8f868)."

BarChartDataSet is  IBarChartDataSet.

Before update to xCode 10.2 it works.

## Charts Environment

**Charts version/Branch/Commit Number:** 3.2.2
**Xcode version:** 10.2
**Swift version:** 4
**Platform(s) running Charts:** iOS 12.2

## Demo Project

https://github.com/anekk/ChartsIssue.git

liuxuan30:
fix #3917. make `init(label: String?)` to be a convenient init to enable convenient initializers auto inheritance.


Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -31,11 +31,9 @@ open class ChartDataSet: ChartBaseDataSet
         super.init()
     }
     
-    public override init(label: String?)
+    public override convenience init(label: String?)
```
jjatie: Why not do this in the ChartBaseDataSet?

liuxuan30: I think there is no obvious reason. `init(label: String?)` is right in ChartDataSet already.

Issue #3861, opened 2019-02-21
Ability to deselect chart value upon highlighting already selected value


jordanhbuiltbyhq:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## Enhancement Request

In my pie chart, I allow users to select values upon tap. This triggers the `chartValueSelected(_ chartView: ChartViewBase, entry: ChartDataEntry, highlight: Highlight)` delegate function. If the user taps the same slice again, this same function is called. Instead of selecting the already selected value, I would like a subsequent tap to deselect the selected value, thus triggering `chartValueNothingSelected(_ chartView: ChartViewBase)`.

I imagine it would be best to expose a bool to configure this behavior, and simply compare the `lastHighlighted` value to the new `highlight` in `highlightValue(_ highlight: Highlight?, callDelegate: Bool)` to know when to select vs deselect.

## Charts Environment

**Charts version/Branch/Commit Number:** 3.2.2
**Platform(s) running Charts:** iOS

liuxuan30:
the current behavior is if you tap an empty space, it will be deselected. I'd say we can do it but not a high priority, but you are welcome to file a PR for this.

hiworld75018:
In PieRadarChartViewBase.tapGestureRecognized(_ recognizer: NSUITapGestureRecognizer),

change :
`let high = self.getHighlightByTouchPoint(location)`
`self.highlightValue(high, callDelegate: true)`

to
`let high = self.getHighlightByTouchPoint(location)`
`if (high === nil || high == self.lastHighlighted)`
`{`
`    self.lastHighlighted = nil`
`    self.highlightValue(nil, callDelegate: true)`
`}`
`else`
`{`
`    self.lastHighlighted = high`
`    self.highlightValue(high, callDelegate: true)`
`}`

Sorry for the inconvenience, I'm not used to PR.

ankit0409:
When you resolve the issue when we select the empty space then selected value is deselect please resolve this issue as soon as possible. I have used your api.

PR #3982, opened 2019-05-01
Deselect chart value upon highlighting already selected value

hipwelljo:
### Issue Link :link:
#3861 

### Goals :soccer:
When you select an already selected value, it will deselect the selected value rather than select it again. This is how MPAndroidChart behaves as well.

It checks if it will be the same and if so deselects instead

### Testing Details :mag:
No changes made to tests

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -529,7 +529,15 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
             }
             else
             {
-                _indicesToHighlight = [h!]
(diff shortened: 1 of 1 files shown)


Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -529,7 +529,15 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
             }
             else
             {
-                _indicesToHighlight = [h!]
+                if (_indicesToHighlight == [h!])
```
liuxuan30: I think you will need a switch to be able control the  behavior. not every one needs this I suppose?

PR #3996, opened 2019-05-13
fix #3975 (pie chart highlight disabled will lead to empty slice)

### LINKED ISSUE #3975: PieChart highlightEnabled flag operation is reverse of setting
PieChartView / PieChartDataSet.highlightEnabled boolean flag works the opposite of how one would expect.

Setting this flag to true DISABLES highlighting.
Setting it to false ENABLES highlighting.  

I would expect it to be reversed.

Seems like a bug.

## What did you do?

let dataset = PieChartDataSet()
dataset.highlightEnabled = false

I expected highlighted to be disabled.

## What happened instead?

Highlighting was enabled.

## Charts Environment

**Charts version/Branch/Commit Number:** 34dafbc6aaac4316656748f1a864426ac42e3d2d
**Xcode version:**  10.2
**Swift version:**  5.0
**Platform(s) running Charts:**  iOS 12.2

## Demo Project

(trivial, as above)

liuxuan30:
if we disable highlight for pie chart data set, we still need to render without a shift.

the old code will ignore the highlighted slice in `drawDataSet()` by `if !chart.needsHighlight(index: j)`, if `set.highlightEnabled=false`, and leave a blank space.


liuxuan30:
@jjatie any time recently to take a look at this?

Issue #4001, opened 2019-05-17
Crash on PieChartRenderer line 833


UberNick:
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

I'm seeing two crash reports failing on the same line of PieChartRenderer from different users with different devices.  This could be memory or timing-related, and I'm writing in the hopes that we can either find a way to replicate this issue or wrap defensive code around the line that's crashing.

## What did you expect to happen?

I'd like PieChartView to do safety-checks before unsafe array operations to prevent hard-crashes.  Even in case of misconfiguration, error logs would be preferred over hard crashes.

## What happened instead?

Hard crash.  Stack trace below.

## Charts Environment

**Charts version/Branch/Commit Number: v3.3.0**
**Xcode version: 10.2.1 (10E1001)**
**Swift version: 5**
**Platform(s) running Charts: 12.2.0 (16E227)**
**macOS version running Xcode: 10.14.4 (18E226)**

## Suggested workaround

Add a defensive if statement around the following unsafe insert:

**PieChartRenderer.swift:833**
```
accessibleChartElements.insert(contentsOf: highlightedAccessibleElements, at: 1)
```

Pull request: https://github.com/danielgindi/Charts/pull/4006

## Stack Trace

```
#0. Crashed: com.apple.main-thread
0  Charts                         0x10084d6a8 specialized Array.replaceSubrange<A>(_:with:) (<compiler-generated>)
1  Charts                         0x10084b890 PieChartRenderer.drawHighlighted(context:indices:) (PieChartRenderer.swift:883)
2  Charts                         0x10078c738 PieChartView.draw(_:) (PieChartView.swift:110)
3  Charts                         0x10078caa4 @objc PieChartView.draw(_:) (<compiler-generated>)
4  UIKitCore                      0x1e479d44c -[UIView(CALayerDelegate) drawLayer:inContext:] + 476
5  (Missing)                      0x693b01bb93bfd8 (Missing)
```

## Relevant Project Code

*For the sake of completeness, I'm including all of my project code that interacts with Charts below.  This code seems to work fine during testing and I'm unable to replicate the crash, leading me to believe the underlying problem related to environment conditions such as low memory.  Crash reports also seem to indicate that the devices that crashed were low on memory at the time.*

```
import UIKit
import Charts

class MyView: UIView {
[...]

PR #4006, opened 2019-05-22
Added a safety check before an unsafe array operation

UberNick:
### Issue Link :link:

https://github.com/danielgindi/Charts/issues/4001

### Goals :soccer:

This code change helps safeguard against hard crashes caused by an array operation.  It doesn't fix potential issues that could cause such a condition in the first place.

### Testing Details :mag:

I manually stepped through the attached demo project and verified the PieChart views looked and behaved as expected.

### DIFF
--- a/Source/Charts/Renderers/PieChartRenderer.swift
+++ b/Source/Charts/Renderers/PieChartRenderer.swift
@@ -880,7 +880,9 @@ open class PieChartRenderer: DataRenderer
 
         // Prepend selected slices before the already rendered unselected ones.
         // NOTE: - This relies on drawDataSet() being called before drawHighlighted in PieChartView.
-        accessibleChartElements.insert(contentsOf: highlightedAccessibleElements, at: 1)
+        if !accessibleChartElements.isEmpty {
+            accessibleChartElements.insert(contentsOf: highlightedAccessibleElements, at: 1)
+        }
 
         context.restoreGState()
     }
--- a/Source/Supporting Files/Info.plist
+++ b/Source/Supporting Files/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.3.0</string>
+	<string>3.3.1</string>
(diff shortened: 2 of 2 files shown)


Source/Charts/Renderers/PieChartRenderer.swift
```diff
@@ -880,7 +880,9 @@ open class PieChartRenderer: DataRenderer
 
         // Prepend selected slices before the already rendered unselected ones.
         // NOTE: - This relies on drawDataSet() being called before drawHighlighted in PieChartView.
-        accessibleChartElements.insert(contentsOf: highlightedAccessibleElements, at: 1)
+        if !accessibleChartElements.isEmpty {
```
liuxuan30: A) I checked the code, it seems `accessibleChartElements` will have at least one element in `drawDataSet` normally, unless `drawDataSet` exits early, such as 
```swift
    @objc open func drawDataSet(context: CGContext, dataSet: IPieChartDataSet)
    {
        guard let chart = chart else {return }
```
otherwise it should have at least 1 default element and `insert(:_, at: 1)` won't crash then. I don't think a low memory case would cause the crash here, if it's really true, then I think it's Swift bug rather than ours.

Now since you said it's crashed, I suppose above guard failed probably. Can you double check?

B) The safe check seems harmless to me, and I'm kind of ok to merge here. But we need to discuss a little more before we merge.

UberNick: Thanks for looking into this, @liuxuan30.  I agree with your assessment that this condition *shouldn't* happen through normal use, and the rarity of the error among our user base combined with my inability to reproduce the problem reinforces that conclusion.

In my use case, Charts elements are being used within recycled TableViewCells and being driven by asynchronous data requests.  Perhaps the error only occurs due to a rare race-condition or thread-related problem.  One that may be more likely, perhaps even necessary, on devices where asynchronous processes are being suspended or terminated due to out-of-memory issues.

This is total speculation from me, and I don't have data to support the above statement.  In the absence of other data, though, I think it's a good guess that's reasonably addressed by this otherwise innocuous PR.

jonnyhaozhao:
test

Issue #4016, opened 2019-06-07
Unable to compile with Swift Package Manager due to missing dependencies


rynecheow:
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag [`ios-charts`.](https://stackoverflow.com/questions/tagged/ios-charts)
>
> Please remove this line and everything above it before submitting.

* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

1. Clone the project
2. `cd` to working directory
3. Run `swift build`
4. Compile failure due to multiple missing reference frameworks

## What did you expect to happen?
Compilation successful for all supported platforms

## What happened instead?
Compilation failed for all platforms. Log attached.
```
Apple Swift version 5.0.1 (swiftlang-1001.0.82.4 clang-1001.0.46.5)
Target: x86_64-apple-darwin19.0.0
[1/2] Compiling Swift Module 'Charts' (136 sources)
/Users/rynecheow/Desktop/Charts/Source/Charts/Components/YAxis.swift:73:36: error: use of undeclared type 'NSTextAlignment'
    @objc open var labelAlignment: NSTextAlignment = .left
                                   ^~~~~~~~~~~~~~~
/Users/rynecheow/Desktop/Charts/Source/Charts/Components/YAxis.swift:73:36: error: use of undeclared type 'NSTextAlignment'
    @objc open var labelAlignment: NSTextAlignment = .left
                                   ^~~~~~~~~~~~~~~
/Users/rynecheow/Desktop/Charts/Source/Charts/Charts/PieRadarChartViewBase.swift:606:49: error: use of undeclared type 'NSEvent'
    open override func mouseDown(with theEvent: NSEvent)
                                                ^~~~~~~
/Users/rynecheow/Desktop/Charts/Source/Charts/Jobs/AnimatedViewPortJob.swift:63:22: error: use of unresolved identifier 'CACurrentMediaTime'
        _startTime = CACurrentMediaTime()
                     ^~~~~~~~~~~~~~~~~~
[...]

PR #4017, opened 2019-06-07
Fix Swift Package Manager compile issue

rynecheow:
Signed-off-by: Ryne Cheow <rynecheow@gmail.com>

### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4016

### Goals :soccer:
1. Resolve dependency errors from SPM builds or `swift build`

Resolve dependencies for correct platform

### Testing Details :mag:
Compile for all platforms using Swift 5.0 toolchain

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,4 +72,4 @@ fastlane/test_output
 Carthage
 Charts.framework.zip
 ChartsRealm.framework.zip
-.swiftpm
+.swiftpm/
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ fastlane/test_output
 Carthage
 Charts.framework.zip
 ChartsRealm.framework.zip
+.swiftpm
--- a/ChartsDemo-iOS/Swift/Demos/CombinedChartViewController.swift
+++ b/ChartsDemo-iOS/Swift/Demos/CombinedChartViewController.swift
@@ -6,8 +6,13 @@
 //  Copyright © 2017 jc. All rights reserved.
 //
 
-import UIKit
+#if canImport(UIKit)
+    import UIKit
+#endif
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 private let ITEM_COUNT = 12
 
--- a/ChartsDemo-iOS/Objective-C/Components/BalloonMarker.swift
+++ b/ChartsDemo-iOS/Objective-C/Components/BalloonMarker.swift
@@ -11,7 +11,9 @@
 
 import Foundation
 import Charts
-
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 open class BalloonMarker: MarkerImage
 {
--- a/ChartsDemo-iOS/Objective-C/Components/RadarMarkerView.swift
+++ b/ChartsDemo-iOS/Objective-C/Components/RadarMarkerView.swift
@@ -11,6 +11,9 @@
 
 import Foundation
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 open class RadarMarkerView: MarkerView
 {
--- a/ChartsDemo-iOS/Objective-C/Components/XYMarkerView.swift
+++ b/ChartsDemo-iOS/Objective-C/Components/XYMarkerView.swift
@@ -6,6 +6,9 @@
 
 import Foundation
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 open class XYMarkerView: BalloonMarker
 {
--- a/ChartsDemo-iOS/Swift/AppDelegate.swift
+++ b/ChartsDemo-iOS/Swift/AppDelegate.swift
@@ -6,7 +6,9 @@
 //  Copyright © 2017 jc. All rights reserved.
 //
 
-import UIKit
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
--- a/ChartsDemo-iOS/Swift/Components/BalloonMarker.swift
+++ b/ChartsDemo-iOS/Swift/Components/BalloonMarker.swift
@@ -11,6 +11,9 @@
 
 import Foundation
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 open class BalloonMarker: MarkerImage
 {
--- a/ChartsDemo-iOS/Swift/Components/RadarMarkerView.swift
+++ b/ChartsDemo-iOS/Swift/Components/RadarMarkerView.swift
@@ -11,6 +11,9 @@
 
 import Foundation
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 public class RadarMarkerView: MarkerView {
     @IBOutlet var label: UILabel!
--- a/ChartsDemo-iOS/Swift/Components/XYMarkerView.swift
+++ b/ChartsDemo-iOS/Swift/Components/XYMarkerView.swift
@@ -8,6 +8,9 @@
 
 import Foundation
 import Charts
+#if canImport(UIKit)
+    import UIKit
+#endif
 
 public class XYMarkerView: BalloonMarker {
     public var xAxisValueFormatter: IAxisValueFormatter
--- a/ChartsDemo-iOS/Swift/DemoBaseViewController.swift
+++ b/ChartsDemo-iOS/Swift/DemoBaseViewController.swift
@@ -6,7 +6,9 @@
 //  Copyright © 2017 jc. All rights reserved.
 //
 
-import UIKit
+#if canImport(UIKit)
+    import UIKit
+#endif
 import Charts
 
 enum Option {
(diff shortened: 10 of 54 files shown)


rynecheow:
@liuxuan30 Could you help to review this?

liuxuan30:
Hmm this seems caused by Xcode 11 beta? @rynecheow any documentation or reason, about this behavior, that it's expected for Xcode 11, or new Swift version?

Because given the history dealing with swift or xcode upgrade, I have seen the same changes about this import back and forth. What I'm concerning is this might be a bug for Xcode 11 beta, not us. 

Any idea @jjatie ?

I could also prefer holding this PR until Xcode 11 GM is released. 

But, for anyone testing Xcode 11 or iOS13, we could create a branch for it, and you can commit your PR to this branch. Is this ok for you?

rynecheow:
Hey @liuxuan30 I just ran a deeper test using **current** stable tools (Swift 5.0 toolchain and Xcode 10.2.1) and build has failed as well. Updated the issue to reflect more accurate problem. 

Basically `swift build` would fail.

liuxuan30:
@rynecheow it's strange, our CI and my Xcode 10.2.1 just works well.

rynecheow:
@liuxuan30 Currently CI is not building the SPM package according build script / build log I believe.

jjatie:
This shouldn't be merged until later this summer. We are only on beta 1 and things are subject to change (especially where all the SPM stuff was developed in private). I noticed the toolchain wasn't bumped in this PR

rynecheow:
@jjatie This pull request aims to fix the build error on current stable Swift toolchain - not Xcode 11 
re. #4016

liuxuan30:
@rynecheow I kind of understand what you mean now.

 if this is about fixing 'Swift Package Manager' compiling - which is rarely used by me or other people so far (you are the first one reporting this), I would like to know or maybe send a bug to APPL first to figure out why only this `swift compile` fails to build, while `xcodebuild` works well.

It's kind of inconsistent, as SPM fails to find the framework while xcodebuild can handle it?

jjatie:
@rynecheow The current toolchain doesn't support iOS. I'm not sure how we could fix it without migrating to the new toolchain.

jjatie:
Why was Package.swift not updated to the 5.0 toolchain?
Why were the demos touched at all when they aren't part of Package.swift?

mattjgalloway:
Chiming in here - I think that we should commit this. Reason being, things are being used when they're not explicitly imported. I haven't looked into why this works when compiling using the Xcode project, but just take a look at this:

`DataApproximator+N.swift`:

```
import Foundation

extension CGPoint {
    fileprivate func distanceToLine(from linePoint1: CGPoint, to linePoint2: CGPoint) -> CGFloat {
        let dx = linePoint2.x - linePoint1.x
        let dy = linePoint2.y - linePoint1.y
        
        let dividend = abs(dy * self.x - dx * self.y - linePoint1.x * linePoint2.y + linePoint2.x * linePoint1.y)
        let divisor = sqrt(dx * dx + dy * dy)
        
        return dividend / divisor
    }
}
```

How can we be extending `CGPoint` without importing CoreGraphics where `CGPoint` resides? AFAIK Foundation in no way shouldn't be importing CoreGraphics. Must be something about the Xcode project that means that CoreGraphics (or UIKit so CoreGraphics recursively) is imported when compiling this file. But really, we should be importing what we use anyway.

RamblinWreck77:
@liuxuan30 Also confirming this issue still occurs in SwiftPM on the Gold Master of Xcode11.

RamblinWreck77:
Also, the Package.swift specifies:

```
 .library(name: "Charts", type: .dynamic, targets: ["Charts"])
```

While the docs specify it is best to leave it nil and let SwiftPM decide. Is there any reason to force .dynamic? Dynamic libraries impact app boot times on iOS.

liuxuan30:
@jjoelson can you check what @RamblinWreck77 said?

I'm open to merge this, if the final compiler takes it this way.

Source/Supporting Files/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.3.0</string>
+	<string>3.3.1</string>
```
liuxuan30: please don't touch it in this PR

ChartsDemo-iOS/Swift/Demos/CombinedChartViewController.swift
```diff
@@ -40,11 +41,11 @@ class CombinedChartViewController: DemoBaseViewController {
         chartView.highlightFullBarEnabled = false
         
         
-        chartView.drawOrder = [DrawOrder.bar.rawValue,
-                               DrawOrder.bubble.rawValue,
-                               DrawOrder.candle.rawValue,
-                               DrawOrder.line.rawValue,
-                               DrawOrder.scatter.rawValue]
+        chartView.drawOrder = [CombinedChartView.DrawOrder.bar.rawValue,
```
liuxuan30: why change this?

liuxuan30:
https://github.com/danielgindi/Charts/pull/4118
seems this is ready to merge, if you don't want to commit more?

ChartsDemo-iOS/Swift/Demos/CombinedChartViewController.swift
rynecheow: Due to the ambiguity of the enum

Source/Supporting Files/Info.plist
rynecheow: I'll revert.

ChartsDemo-iOS/Swift/Demos/CombinedChartViewController.swift
rynecheow: Changed back to original - my bad here!

jjoelson:
@liuxuan30 The SwiftPM docs do say to leave `type` nil unless the library disallows a linkage type for other reasons:

>Leave this parameter unspecified to let to let the Swift Package Manager choose between static or dynamic linking (recommended). If you do not support both linkage types, use `.static` or `.dynamic` for this parameter.

But then there's also an example that has separate products for static and dynamic, so it's a bit confusing:
```
let package = Package(
    name: "Paper",
    products: [
        .executable(name: "tool", targets: ["tool"]),
        .library(name: "Paper", targets: ["Paper"]),
        .library(name: "PaperStatic", type: .static, targets: ["Paper"]),
        .library(name: "PaperDynamic", type: .dynamic, targets: ["Paper"]),
    ],
    dependencies: [
        .package(url: "http://example.com.com/ExamplePackage/ExamplePackage", from: "1.2.3"),
        .package(url: "http://some/other/lib", .exact("1.2.3")),
    ],
    targets: [
        .target(
            name: "tool",
            dependencies: [
                "Paper",
                "ExamplePackage"
            ]),
        .target(
            name: "Paper",
            dependencies: [
                "TSCBasic",
                .target(name: "Utility"),
                .product(name: "AnotherExamplePackage"),
            ])
    ]
)
```

liuxuan30:
there is no reason to only use dynamic, but by default, this "library" is built with framework, so I guess that's why it's dynamic. I think we should be fine with nil type, as long as SPM has no bugs handling it.

@rynecheow can you also check what #4118 did for `Package.swift `? Basically I would like to merge this as you are the first PR. I didn't see you touch this file, but #4118 change a few, especially 
>.target(name: "Charts", dependencies: [], path: "./Source", publicHeadersPath: "./Source/Supporting Files")

I'm not sure the impact about this, as I'm not following SPM trend now. Besides, I will be grateful if you could modify the type to nil as discussed above, so I don't have to pull your branch. Thanks.

.gitignore
```diff
@@ -72,3 +72,4 @@ fastlane/test_output
 Carthage
 Charts.framework.zip
 ChartsRealm.framework.zip
+.swiftpm
```
liuxuan30: `.swiftpm` or `.swiftpm/`?

rynecheow: Rightfully `.swiftpm/` - let me change that. Did `.swiftpm` cause it does the trick.

rynecheow:
> there is no reason to only use dynamic, but by default, this "library" is built as framework, so I guess that's why it's dynamic. I think we should be fine with nil type, as long as SPM has no bugs handling it.
> 
> @rynecheow can you also check what #4118 did for `Package.swift `? Basically I would like to merge this as you are the first PR. I didn't see you touch this file, but #4118 change a few, especially
> 
> > .target(name: "Charts", dependencies: [], path: "./Source", publicHeadersPath: "./Source/Supporting Files")
> 
> I'm not sure the impact about this, as I'm not following SPM trend now. Besides, I will be grateful if you could modify the type to nil as discussed above, so I don't have to pull your branch. Thanks.

`path: "./Source"` is unnecessary as it is the parameter automatically defaults to a certain pattern if it is not specified. 
https://developer.apple.com/documentation/swift_packages/target/2880335-path

`publicHeadersPath: "./Source/Supporting Files"` I am not sure if this is completely necessary as I believe SPM already handles that at its build process.

liuxuan30:
Seems green light now. BTW, do you think you can add script code for travis to build via SPM? Just to build it to make sure it builds fine, No UT is needed.

Otherwise I will merge next Monday.

liuxuan30:
A second thought we should just go ahead, if you think travis ci is needed, let's add later. Not surehowpopular spm can be

Issue #3185, opened 2018-01-16
Scatter circle disappear when chart scaled


kou0117:
if you have more than 2 scatter on the chart and scale the chart to the extent that any scatter get invisible, all scatters disappear.

![scatter disappeared](https://user-images.githubusercontent.com/26572756/34977050-7205d22c-fadc-11e7-8515-3143214ad0de.gif)

if you know the solution, please tell me.
Thank you.

liuxuan30:
can this be reproduced with ChartsDemo? It's a bug and should be fixed.

kou0117:
Thank you for your reply but I got fixed.
when I appended ChartDataEntry In order of new, scatter circle disappeared when chart scaled.
However, when I appended ChartDataEntry In order of oldest, scatter circle didn't disappear.

I assume that when the scatter appended first get invisible, every scatter also get invisible. 

In the case of my GIF attached above, I appended the left scatter first.
Thank you.

liuxuan30:
hmm. so you think everything is fine? Good to know and closing.

Issue #2884, opened 2017-10-16
ScatterViewChart entries hide when zooming


sss989870:
ScatterViewChart entries hide when zooming
When scrolling the maximum item is hidden, all other items disappear.
is it have option?

liuxuan30:
not quite understand what are you asking? a bug?

sss989870:
a bug.
When I scrolled after I zoomed in, some of the entries in the data set disappeared.

liuxuan30:
can we reproduce it with ChartsDemo? It looks fine though

sss989870:
before zooming.
https://drive.google.com/file/d/0B3Lv_uliEGSjWkZfUTlNT3Z5MGs/view?usp=sharing
shown all entries.

but after zooming.
https://drive.google.com/file/d/0B3Lv_uliEGSjMXoyT0lwUWU2WFU/view?usp=sharing
When the entry on the far right disappears, center entries disappears.

liuxuan30:
ok. but we need more details then; the axis range seems not changed much.

have you checked all your data entries are there in `drawDataSet`? need you give a minimum reproducible steps or, code that can reproduce with charts demo

sss989870:
Append Entry Code
```
func updateChartData() {
        var dataSets: [ScatterChartDataSet] = []

        let aliveEntry = aliveSensors.map {
            return Double($0.sensorVoltage) > self.VOLTAGE_MINIMUM ? ChartDataEntry(x: valueToGraphPosition(x: Double($0.sensorVoltage)), y: Double($0.sensorIlluminationAvg)) : ChartDataEntry(x: Double($0.sensorVoltage), y: Double($0.sensorIlluminationAvg))
        }
        
        let deadEntry = deadSensors.map {
            return Double($0.sensorVoltage) > self.VOLTAGE_MINIMUM ? ChartDataEntry(x: valueToGraphPosition(x: Double($0.sensorVoltage)), y: Double($0.sensorIlluminationAvg)) : ChartDataEntry(x: Double($0.sensorVoltage), y: Double($0.sensorIlluminationAvg))
        }
        print("Entry Count : \(aliveEntry.count), \(deadEntry.count)")
        
        let aliveSet = ScatterChartDataSet(values: aliveEntry, label: "Alive")
        aliveSet.setColor(UIColor.blue, alpha: 1.0)
        aliveSet.scatterShapeSize = 3
        aliveSet.setScatterShape(ScatterChartDataSet.Shape.circle)
[...]

PR #4027, opened 2019-06-16
Fix fatal error with x values in reverse order

jimmythai:
### Issue Link :link:

* https://github.com/danielgindi/Charts/issues/2884
* https://github.com/danielgindi/Charts/issues/3185

### Goals :soccer:

Fixing bugs of the issues above and bugs as follows. Those bugs happen on `BarChartViewController` when you set entries which have x values in reverse order. A fatal error occurs on `BubbleChartViewController`, `CandleStickChartViewController`, `LineChart1ViewController` and `ScatterChartViewController` when you do the same things.

Arrays of `ChartDataEntry` should be sorted in an ascending order when you initialize `ChartDataSet` and whenever you set them again, for instance, using `replaceEntries(_:)`.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -38,7 +38,7 @@ open class ChartDataSet: ChartBaseDataSet
     
     @objc public init(entries: [ChartDataEntry]?, label: String?)
     {
-        self.entries = entries ?? []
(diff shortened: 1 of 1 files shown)


liuxuan30:
I wonder if it's our job to sort the array, or let the user handle it? @jjatie @petester42

xymtek:
sorting every-time the data is updated seems expensive.  Maybe having some validation when drawing the next x should check if its larger than the previous x would be better.

Issue #3659, opened 2018-09-20
Index out of bounds exception in v3.2 in createAccessibleElement


simonbengtsson:
After updating to charts 3.2 I started getting `Fatal error: Index out of range` in the new `BarChartRenderer#createAccessibleElement():809`. It seems this code `dataSet.stackLabels[idx % stackSize]` makes stackLabels required?

My workaround for now is to always specify `stackLabels`. Can be done with something like this:

```swift
let dataSet = BarChartDataSet(values: entries, label: nil)
dataSet.stackLabels = [String](repeating: "", count: dataSet.stackSize)
```

## Charts Environment

**Charts version:** 3.2
**Xcode version:** 10
**Swift version:** 4.2
**Platform:** iOS
**macOS:** 10.13.6

liuxuan30:
@petester42 may have answer who's responsible for checking the bound for accessibility

Talgat777:
+1 


1 | Charts | createAccessibleElement (ChartViewBase.swift:47)
2 | Charts | createAccessibleElement (<compiler-generated>:809)
3 | Charts | drawDataSet (HorizontalBarChartRenderer.swift:280)
4 | Charts | drawData (BarChartRenderer.swift:253)
5 | Charts | draw (BarLineChartViewBase.swift:231)

AflyingApple:
I also encountered the same problem.

nathantannar4:
+1

nathantannar4:
@simonbengtsson's workaround still fails fit me

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

petester42:
I believe I've fixed the issue but we should not allow setting colours/stackLabels individually. There are too many cases where they won't be of the some size but they should be. @liuxuan30, @danielgindi what do you guys think?

petester42:
I've fixed the crashes but I found some accessibility issue with our current implementation. The fix will take a little longer

petester42:
Problably the end of the week. I’ll see if I have time tomorrow. Please keep in mind that we are all just volunteering our time here.

Scalman:
Hi, im still getting this error when using stacked bar chart. I can provide you with code if it is necessary!

**Error:**
Thread 1: Fatal error: Index out of range

**Where**
            elementValueText = dataSet.valueFormatter?.stringForValue(
                vals[idx % stackSize],                <-- HERE
                entry: e,
                dataSetIndex: dataSetIndex,
[...]

PR #4029, opened 2019-06-16
fixed stacked chart bug when there are different stacks on columns.

Scalman:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/3659

### Goals :soccer:
Fixing the issue, see link above.

Preventing the stacked bar chart from crashing when there was different number of stacks in columns.

### Testing Details :mag:
No test added, but i did test this against other chart views and confirmed that this fix did not make an affect on them.

### DIFF
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -867,9 +867,12 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
             } else {
                 stackLabel = nil
             }
-
+            
+            //Handles empty array of yValues
+            let yValue = vals.isEmpty ? 0.0 : vals[idx % vals.count]
+            
             elementValueText = dataSet.valueFormatter?.stringForValue(
-                vals[idx % vals.count],
+                yValue,
                 entry: e,
                 dataSetIndex: dataSetIndex,
                 viewPortHandler: viewPortHandler) ?? "\(e.y)"
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -869,7 +869,7 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
             }
 
             elementValueText = dataSet.valueFormatter?.stringForValue(
-                vals[idx % stackSize],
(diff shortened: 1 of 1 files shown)


liuxuan30:
to quote:
```
EDIT: Turns out, I'm just an idiot and wasn't passing uniform data along! If you have data that looks like this:

[0.0, 1.0, 2.0, 3.0]
[0.0, 0.0, 1.0, 2.5]
[]

It crashes with out of bounds because I've given it a reason to believe the stack size is 4 and then totally ignored that. On one hand, it would be kind of cool if it handled this, but on the other hand, it's my own fault. Thanks anyway! :)
```
I wonder, if the size is not respected, so it should crash to give warnings? 

and want to more context, why `stackSize` not equal to `vals.count` ?

Scalman:
@liuxuan30 

I think the way of defining stackSize as if all stacks is equal is wrong and therefore should be used with carefulness due to the misleading nameing convention. stackSize should be renamed to MaxStackSize to minimize the confusion of the actual value it holds. stackSize will always have the value of the entry with the highest count which is in my opinion wrong.  

The BarChartRenderar should use _yVals.count instead of the dataset strange use of stackSize to make it more dynamic. If you ask me I would remove the stackSize entirely and make use of the entry yVals.count as a size.

stackSize as mentioned in the text above is the count of the entry with the highest count of yVals
VS
vals.count is just an entry count of the current entry.

This code snippet below is used to calculate the stackSize.

```    private func calcStackSize(entries: [BarChartDataEntry])
    {
        for i in 0 ..< entries.count
        {
            if let vals = entries[i].yValues
            {
                if vals.count > _stackSize
                {
                    _stackSize = vals.count
                }
            }
        }
    }
```

liuxuan30:
Hi @Scalman, sorry for the late response.

I understood your point here, and I checked the code, the problematic code is in Accessibility feature only. The original bar renderer has no crash when using stackSize. The logic is a bit different. Though it's directly using stackSize, it creates zero bar rect, and if val.count < stackSize, it do nothing and just use this zero bar rect, so no crash at all.

But the Accessibility is directly using `vals[idx % stackSize]` for getting the value text, unlike what it did at line 861: 
>let labelCount = min(dataSet.colors.count, stackSize)

So it seems inconsistent, but your fix is legit in this scenario.

PR #4044, opened 2019-06-27
Fixed a crash while zooming in the chart

S3Paramasivan:
Change done -
Checked for isNan on both on 1) AxisRendererBase.swift and YAxisRendererRadarChart.swift

### DIFF
--- a/Source/Charts/Renderers/AxisRendererBase.swift
+++ b/Source/Charts/Renderers/AxisRendererBase.swift
@@ -97,7 +97,7 @@ open class AxisRendererBase: Renderer
         let labelCount = axis.labelCount
         let range = abs(yMax - yMin)
         
-        if labelCount == 0 || range <= 0 || range.isInfinite
+        if labelCount == 0 || range <= 0 || range.isInfinite || range.isNaN
         {
             axis.entries = [Double]()
             axis.centeredEntries = [Double]()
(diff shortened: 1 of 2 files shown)


S3Paramasivan:
Does the below line mean that you will not commit the change into your repository?

>>>>>>Merging #4044 into master will not change coverage.

What should i do so that my next pull will have these changes so that the app will not crash while zooming in the chart?

liuxuan30:
why range is NAN after your zooming? Can you post the log/data, etc?

For all NAN issues, we need to verify why it's NAN in the first place

S3Paramasivan:
It did not happen during development, in production, it is happening for half of the users.

liuxuan30:
if it's happening for half of your users, it should be easy to find out why. I'm asking for the data value that cause it. Usually you should validate the data first.

I'm holding all NaN PRs to understand why there is NaN in the first place, but so far I didn't get any feedback. The problem is we could definitely silence the crash, however it just hides the chance to find out the issue. Silencing NaN crash and proceed, usually it will make renderer wonder what to draw, either give you a blank space, or a default thing? IMO it's not renderer's job to decide what to do, but the users.

S3Paramasivan:
> if it's happening for half of your users, it should be easy to find out why. I'm asking for the data value that causes it. Is it our fault during calculating, or yours passing an invalid value?
> 
> I'm holding all NaN PRs to understand why there is NaN in the first place, but so far I didn't get any useful feedback. The problem is we could definitely silence the crash, however it just hides the chance to find out the issue. If we silence NaN crashes and proceed, usually it will make renderer wonder what to draw, either give you a blank space, or a default axis? IMO it's not renderer's job to decide what to do, but the users.

You are correct, i will try to get some data causing the issue soon. Thank you.

PR #4045, opened 2019-06-27
fixed Xcode11 SwiftPM build 

yarshure:
### Xcode 11 now can build and test Swift Package:
Should support !


liuxuan30:
dup to https://github.com/danielgindi/Charts/pull/4017

PR #4050, opened 2019-06-28
move isDrawCirclesEnabled check further up in code to avoid creating …

xymtek:
…accessibility labels that will never appear anyway.  Reduces CPU significantly for larger datasets.

### Goals :soccer:
Improve draw performance for larger datasets.

No need to create accessibility labels for each circle if the circle will never be drawn away.

### Testing Details :mag:
After this change drawing and updating 4500 points every 200ms the cpu went from approximately 130% down to about 35%


JimmyMAndersson:
This will do wonders. Any idea of when it can be merged and released?

liuxuan30:
sorry guys for the late. It's merged now.

PR #4064, opened 2019-07-09
Silence a bunch of warnings from Xcode

richo:
### Goals :soccer:
Make Xcode not have a bunch of warnings in the sidebar.

Mostly just let xcode take the wheel. The one exception is that setter, which I think was actually incorrect before.

### Testing Details :mag:
Unfortunately Carthage isn't working correctly, so I wasn't able to run the tests. I was able to confirm that it still builds and works though.


liuxuan30:
I'm not seeing warnings for master. If it's for iOS - there is already PR for it.

PR #4084, opened 2019-07-31
PieChart add onlyShowHighlightValue,  only show valueLine what  your choose

songfeiSF:
PieChart add onlyShowHighlightValue,  only show valueLine what  your choose
image


ChartsDemo-iOS/Objective-C/Demos/PiePolylineChartViewController.m
```diff
@@ -102,7 +102,7 @@ - (void)setDataCount:(int)count range:(double)range
     dataSet.valueLinePart2Length = 0.4;
     //dataSet.xValuePosition = PieChartValuePositionOutsideSlice;
     dataSet.yValuePosition = PieChartValuePositionOutsideSlice;
-    
+//    dataSet.onlyShowHighlightValue = true;
```
liuxuan30: remove this

Source/Charts/Data/Implementations/Standard/PieChartDataSet.swift
```diff
@@ -110,7 +110,9 @@ open class PieChartDataSet: ChartDataSet, IPieChartDataSet
 
     /// the color for the highlighted sector
     open var highlightColor: NSUIColor? = nil
-
+    
+    /// When onlyShowHighlightValue is true ,only show what you choose. default is false
+    open var onlyShowHighlightValue : Bool = false
```
liuxuan30: need a better name to reflect what's the property about.
Also, I think it should be a per chart view setting, as pie chart only has one data set, and it seems only show text value line makes more sense if it's for PieChartView itself?

Source/Charts/Renderers/PieChartRenderer.swift
```diff
@@ -345,6 +345,11 @@ open class PieChartRenderer: DataRenderer
 
             for j in 0 ..< dataSet.entryCount
             {
+                if !chart.needsHighlight(index: j) && dataSet.onlyShowHighlightValue{
```
liuxuan30: code style, put `{` another line

PR #4089, opened 2019-08-05
Allowing overriding for YAxisRenderer.drawYLabels

muclemente:
### Goals :soccer:
MPAndroidChart library allows overriding of this method in a subclass of YAxisRenderer. This PR changes the access modifier so it can be overridden in iOS as well.


liuxuan30:
Thanks!

PR #4094, opened 2019-08-07
Fix line chart x axis animation #4093, also close #3960

### LINKED ISSUE #3960: Horizontal Bezier Line Graph Not Drawing Through All Points
## What did you do?

Set `LineChartDataSet.mode` to `.horizontalBezier`. 

The line would connect all points in a horizontal-bezier fashion.

## What happened instead?

The first point is being skipped and is not connected to the line at all.
Screen Shot 2019-04-22 at 9 28 51 AM

## Charts Environment

**Charts version/Branch/Commit Number:**
**Xcode version:** 10.2
**Swift version:** 5
**macOS version running Xcode:** 10.14.3

## Demo Project

[ChartsTesting.zip](https://github.com/danielgindi/Charts/files/3103529/ChartsTesting.zip)

### LINKED ISSUE #4043: LineChartView.highlightValue causes CoreGraphics API errors
## What did you do?

Called `LineChartView.highlightValue(x: Double, dataSetIndex: Int)` with a valid `x` & `dataSetIndex`

I expected the Chart to Highlight the appropriate X/Y point for the given X value

## What happened instead?

Nothing was drawn, console spews out the error: `Error: this application, or a library it uses, has passed an invalid numeric value (NaN, or not-a-number) to CoreGraphics API and this value is being ignored. Please fix this problem.`

## Charts Environment

**Charts version/Branch/Commit Number:** 3.3.0
**Xcode version:** 10.2.1
**Swift version:** 5.0
**Platform(s) running Charts:** iOS 12.3.1

## Demo Project

```swift
        let chartView = LineChartView(frame: self.view.frame)

        let dataSet = LineChartDataSet(entries: [], label: "Test Dataset")
        let data = LineChartData(dataSets: [])
        chartView.data = data

        chartView.data!.addDataSet(dataSet)

        for i in 0..<10 {
            let entry = ChartDataEntry(x: Double(i), y: Double(i))
            chartView.data!.addEntry(entry, dataSetIndex: 0)
        }

        chartView.notifyDataSetChanged()

        chartView.highlightValue(x: 2.0, dataSetIndex: 0)
```

**Note:** From my own digging the issue here is in the `LineChartRenderer.drawHighlighted` func. Specifically the following lines:

```swift
            guard let e = set.entryForXValue(high.x, closestToY: high.y) else { continue }
            
            if !isInBoundsX(entry: e, dataSet: set)
            {
                continue
            }
            .......
            .......
            let x = high.x // get the x-position
            let y = high.y * Double(animator.phaseY)
            
            if x > chartXMax * animator.phaseX
            {
                continue
            }
            
            let trans = dataProvider.getTransformer(forAxis: set.axisDependency)
            
            let pt = trans.pixelForValues(x: x, y: y)
            
            high.setDraw(pt: pt)

            // draw the lines
            drawHighlightLines(context: context, point: pt, set: set)
```

Since `high.y` is NaN, the transformer attempts to transform a CGPoint where Y is NaN. This causes pt X & Y to both be NaN which is then passed to `drawHighlightLines` resulting in CoreGraphics API errors.

I believe a fix here would be to use the result from `set.entryForXValue` above? But i'm not sure what implications that would have on other highlighting funcs.

### LINKED ISSUE #4093: Line chart x axis animation is broken
Line chart animation on x axis is broken:
1. the value labels are always on the screen;
2. horizontal bezier animation is broken
3. additional dash line while animting `drawLinear`:
image

liuxuan30:
1. change XBounds iterator to use self.min + self.range rather than self.x
2. align drawLinear,  drawCubicBezier to new XBounds iterator.
3. fix unexpected dash line during linear animation due to reading the next entry point

### Issue Link :link:
#4093, #3960

It seems to fix issue 3, the simple way is to call `for j in _xBounds.dropLast()` instead of `for j in _xBounds` so it ignores the last point, then j+1 will always be the last point to avoid additional dash line

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -117,10 +117,8 @@ open class LineChartRenderer: LineRadarRenderer
             // Take an extra point from the left, and an extra from the right.
             // That's because we need 4 points for a cubic bezier (cubic=4), otherwise we get lines moving and doing weird stuff on the edges of the chart.
             // So in the starting `prev` and `cur`, go -2, -1
-            // And in the `lastIndex`, add +1
             
             let firstIndex = _xBounds.min + 1
-            let lastIndex = _xBounds.min + _xBounds.range
             
             var prevPrev: ChartDataEntry! = nil
             var prev: ChartDataEntry! = dataSet.entryForIndex(max(firstIndex - 2, 0))
@@ -133,7 +131,7 @@ open class LineChartRenderer: LineRadarRenderer
             // let the spline start
             cubicPath.move(to: CGPoint(x: CGFloat(cur.x), y: CGFloat(cur.y * phaseY)), transform: valueToPixelMatrix)
             
-            for j in stride(from: firstIndex, through: lastIndex, by: 1)
+            for j in _xBounds.dropFirst()  // same as firstIndex
             {
                 prevPrev = prev
                 prev = cur
@@ -316,8 +314,8 @@ open class LineChartRenderer: LineRadarRenderer
                 // Allocate once in correct size
                 _lineSegments = [CGPoint](repeating: CGPoint(), count: pointsPerEntryPair)
             }
-            
-        for j in stride(from: _xBounds.min, through: _xBounds.range + _xBounds.min, by: 1)
+
+        for j in _xBounds.dropLast()
         {
             var e: ChartDataEntry! = dataSet.entryForIndex(j)
             
@@ -328,6 +326,9 @@ open class LineChartRenderer: LineRadarRenderer
             
             if j < _xBounds.max
             {
+                // TODO: remove the check.
+                // With the new XBounds iterator, j is always smaller than _xBounds.max
+                // Keeping this check for a while, if xBounds have no further breaking changes, it should be safe to remove the check
                 e = dataSet.entryForIndex(j + 1)
                 
                 if e == nil { break }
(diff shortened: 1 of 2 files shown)


Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -328,6 +326,9 @@ open class LineChartRenderer: LineRadarRenderer
             
             if j < _xBounds.max
             {
+                // TODO: remove the check.
```
danielgindi: This is somewhat incorrect, and the iterator in `XBounds` is theoretically incorrect.
As the point is to iterate through `min + range` - where range *could* go over `max` due to a bouncing animation or something similar.

The thing is that this `XBounds` represents (in all current cases) a range of entries, and going over `max` could mean an `IndexOutOfBounds`.
That's why the check is there.

If the iterator only iterates until `max` - then it's a bug and it's not respecting the animation (range is basically `(max - min) * animation.PhaseX`). 

Also I would change the iterator's `fileprivate init(min: Int, max: Int) {` to `fileprivate init(min: Int, through: Int) {`, as it's really confusing right now, both with the `max` property and with `to vs through` semantics.

liuxuan30: Hi @danielgindi I came across here while I am working on the 4.0->master merge. 

I'm a little confusing that you saying `the iterator in XBounds is theoretically incorrect.` 

Which iterator you say is wrong? my `return Iterator(min: self.min, max: self.min + self.range)` or `return Iterator(min: self.min, max: self.max)` ? 

because from the old session I wrote here https://github.com/danielgindi/Charts/issues/4093
It seems the iterator of xbounds now is correct:
```swift
    public func makeIterator() -> Iterator {
        return Iterator(min: self.min, max: self.min + self.range)
    }
```
I will rename the iterator to `through` while merging 4.0.

PR #4100, opened 2019-08-08
Fixes #4099: Line renderer did not render lines if their coordinates fell outside of the viewport.

### LINKED ISSUE #4099: Line charts: Line sections disappear when zoomed in (has PR).
# Line sections disappear when zoomed in

Regularly, line sections of line-based charts (line, area) will disappear when zoomed in (make sure `scaleXEnabled = true`, `scaleYEnabled = true` and `dragEnabled = true`). 

Possibly related issues: #3253, #3933, #3898?

# Steps to reproduce

1. Launch the iOS Charts demo app
2. Open the `Line Chart` demo
3. Zoom in deep (so that the left y-axis ranges from 20 to 70, zoom level > 3.8)
4. Drag the chart so that a sharp line corner will _just_ go below the x-axis
5. Observe the left line section will disappear

issue-4099

(View the [full video](https://www.dropbox.com/s/c8fbmt8m7i5q31x/iOS%20Charts%20Issue%204099.MOV?dl=0) for more context)

# Cause of the issue

The issue is caused by some logic in [LineChartRenderer.swift](https://github.com/danielgindi/Charts/blob/master/Source/Charts/Renderers/LineChartRenderer.swift#L356) that determines whether or not a line should be rendered. It does so by checking if a line lies outside the viewport, however it does not take into account that lines might _intersect_ the viewport while its start and endpoint lie outside of the viewport.

In the above gif, the start point of the line was outside of the viewport while the end point lies within the viewport. Dragging the chart downward makes the end point move outside of the viewport, causing the logic below to skip rendering.

```swift
if (!viewPortHandler.isInBoundsRight(_lineSegments[0].x))
{
    break
}

// make sure the lines don't do shitty things outside bounds
if !viewPortHandler.isInBoundsLeft(_lineSegments[1].x)
    || (!viewPortHandler.isInBoundsTop(_lineSegments[0].y) && !viewPortHandler.isInBoundsBottom(_lineSegments[1].y))
{
    continue
}
```

# Fix

In Pull Request #4100 I have improved the logic to determine whether or not a line's start- or endpoint lie within the viewport, _or_ if the line intersects the viewport.

# Environment

- Charts version 3.3.0 (it is also present in 3.2.2)
- Xcode 10.3
- Swift 5
- macOS Mojave 10.14.16

4np:
### Fixes #4099 

The line renderer is optimized to only render line sections that are inside the viewport. However, there is an issue with the logic as it does not take into account that a line might intersect the viewport without having its points inside the viewport.

This pull requests adds collision detection to see if a line intersects the viewport and, if so, render the line.


liuxuan30:
this is somehow a big change. I need more time to work on this.

liuxuan30:
@jjatie @petester42 @danielgindi any feedbacks?

liuxuan30:
hi, I plan review and possibly merge this week. Can you pull master and rebase on it? I fixed the Xcode11 CI failures. Just in case it breaks something

Source/Charts/Utils/ViewPortHandler.swift
```diff
@@ -443,6 +454,37 @@ open class ViewPortHandler: NSObject
...
+    @objc open func isIntersectingLine(from startPoint: CGPoint, to endPoint: CGPoint) -> Bool
+    {
+        // If the start or endpoint fall within the viewport, bail out early.
+        if isInBounds(point: startPoint) || isInBounds(point: endPoint) { return true }
+        
+        // Calculate the slope of the line.
+        let slope = (endPoint.y - startPoint.y) / (endPoint.x - startPoint.x)
+
+        // Check for colission with the left edge of the view port.
+        if isInBoundsY((slope * (contentRect.minX - startPoint.x)) + startPoint.y) { return true }
```
liuxuan30: sorry my math is a little rusty here, can you explain what is `contentRect.minX - startPoint.x` stands for here?

I'd expect details about the calculation, maybe a screenshot of the points you want to calculate could explain all

4np: It will calculate the `y-value` of the sloped line, by using the `x-value` of the left edge of the view port. It will then check if that t-value `isInBoundsY`. If so, the line intersects the left edge of the view port.

Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -353,17 +353,20 @@ open class LineChartRenderer: LineRadarRenderer
...
-            if !viewPortHandler.isInBoundsLeft(_lineSegments[1].x)
-                || (!viewPortHandler.isInBoundsTop(_lineSegments[0].y) && !viewPortHandler.isInBoundsBottom(_lineSegments[1].y))
-            {
-                continue
-            }
+            // If both points lie left of viewport, skip stroking.
+            if !viewPortHandler.isInBoundsLeft(firstCoordinate.x) && !viewPortHandler.isInBoundsLeft(lastCoordinate.x) { continue }
+            
+            // If both points lie right of the viewport, break out early.
+            if !viewPortHandler.isInBoundsRight(firstCoordinate.x) && !viewPortHandler.isInBoundsRight(lastCoordinate.x) { break }
```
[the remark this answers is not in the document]
4np: We _should_ not need to check for `lastCoordinate.x` as lines are always rendered left-to right. So we can probably remove the second part of the condition.

4np:
I have merged master into the branch, updated the logic based on your comments, and made sure that everything still worked 👍

Source/Charts/Utils/ViewPortHandler.swift
liuxuan30: hi @4np, yes, I just couldn't remember what point location is for `contentRect.minX - startPoint.x`, if we want `left edge of the view port`, isn't it `contentRect.minX`?

liuxuan30:
> I have merged master into the branch, updated the logic based on your comments, and made sure that everything still worked 👍

thanks, let's just focus on the review section, once it's sorted out, we should be able  to go.

Source/Charts/Utils/ViewPortHandler.swift
4np: In #ababe6f I updated the method that determines whether or not the line intersects the view port by:

- explicitly using the linear function `y = ax + b` and its variables `a` and `b`
- quoting the equations used
- as lines run from left to right, skip unnecessary checking for collision with the right edge of the view port

liuxuan30: I have added a check for `startPoint.x == endPoint.x`
looks good to merge now.

liuxuan30:
I have added a check for startPoint.x == endPoint.x
looks good to merge now.

liuxuan30:
@4np please have a look. if no question I will merge like tomorrow

danielgindi:
Hey, sorry for the delay!
This code actually seems irrelevant, except for the check with correct `lastCoordinate` for stepped lines.
This is because if you are checking simply that `lastCoordinate.x > left && firstCoordinate.x < right`, then that's enough and will include the scenario you are describing.
Same goes for vertical.

Actually the only wrong thing there with is the check for vertical lines was inverted.

The collision check costs too much there with no benefit.

PR #4105, opened 2019-08-13
Handles NaN values (Fixes #3989)

### LINKED ISSUE #3989: ChartDataSet.entryIndex(xValue:yValue:rounding) doesn't handle NaN for xValue
https://github.com/danielgindi/Charts/blob/899fe1d13208ee6c08ce8db790a834f77dc9aee7/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift#L274

Passing a nan as the xValue currently causes an infinite loop. 

https://github.com/chrisjeane/Charts/tree/fixes-entry-index-nan-handling

chrisjeane:
Adds nan handling for the xValue of entryIndex(xValue:yValue:rounding)

### Issue Link :link:
#3989

### Goals :soccer:
This would be best described as not supporting default PieChartDataEntries that have their x values set to Nan. I'm don't think that is an issue, but passing a nan as the xValue should not cause an infinite loop.

Adds a guard statement to ensure `func entryIndex` does not enter a infinite loop

### Testing Details :mag:
`func entryIndex` should now support PieChartDataEntries with Nan xValues

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -278,6 +278,7 @@ open class ChartDataSet: ChartBaseDataSet
         closestToY yValue: Double,
         rounding: ChartDataSetRounding) -> Int
     {
+        guard xValue.isFinite else { return -1 }
         var low = startIndex
         var high = endIndex - 1
         var closest = high


Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -278,6 +278,7 @@ open class ChartDataSet: ChartBaseDataSet
         closestToY yValue: Double,
         rounding: ChartDataSetRounding) -> Int
     {
+        guard xValue.isFinite else { return -1 }
```
liuxuan30: why guard inf here? I was thinking you are talking about NaN here, but

chrisjeane: It is an attempt to validate input before it causes the library to enter an infinite loop. From my observations NaN values cause this behavior.

liuxuan30: why you get NaN here?

chrisjeane: NaN is the default x value for PieChartDataEntry

liuxuan30: where it goes into a dead loop? Currently pie chart data entry is working fine I suppose in ChartsDemo and our CI tests. We need to understand where this lead to a dead loop.

PR #4118, opened 2019-08-23
SPM support for iOS, macOS, tvOS

Ragnalorn:
Currently Swift Package Manager for Xcode11 can only build static libraries. This means that UIKit can't be linked dynamically, so we need to check if UIKit is available and import it if possible. “canImport” is available from Swift 4.1, so these changes won’t cause any compiling issues and everything will work correctly (Carthage, Pods, etc). Since it's CRITICAL to make this lib Swift Package Manager compatible for appleOS platforms, could you please merge this pull request as soon as possible?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ fastlane/test_output
 Carthage
 Charts.framework.zip
 ChartsRealm.framework.zip
+.swiftpm/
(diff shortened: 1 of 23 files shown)


liuxuan30:
https://github.com/danielgindi/Charts/pull/4017

We only need one, can you guys work together, or which one is slightly better to merge?

rynecheow:
@Ragnalorn `swift build` is failing for macOS target in your PR.

liuxuan30:
Thank you for your effort. We have merged the other one, so closing this

PR #4123, opened 2019-08-26
Use interpolation instead of '+' concatenation for problematic expression

Jumhyn:
### Goals :soccer:
Reduces the type-checking time of a `debugDescription` variable by at least an order of magnitude.

Because type inference with `+` and literals has to do a lot of overload checking (for every `ExpressibleBy____Literal`), many expressions which use `+` for string concatenation with literals in contexts where type information is sparse may inadvertently cause a type checker performance explosion. In this case, we are concatenating an anonymous closure argument with a string literal, which caused a compilation time on the order of multiple seconds (on my MBP, YMMV). With this change, the type checking of this expression completes in under 100ms.

### Testing Details :mag:
There's no behavior change here. We could consider adding `-Xfrontend -warn-long-expression-type-checking=<limit>` to the Xcode project to catch warnings about type checker performance bottlenecks.

### DIFF
--- a/Source/Charts/Data/Implementations/ChartBaseDataSet.swift
+++ b/Source/Charts/Data/Implementations/ChartBaseDataSet.swift
@@ -403,7 +403,7 @@ open class ChartBaseDataSet: NSObject, IChartDataSet, NSCopying
     open override var debugDescription: String
     {
         return (0..<entryCount).reduce(description + ":") {
-            $0 + "\n" + (self.entryForIndex($1)?.description ?? "")
+            "\($0)\n\(self.entryForIndex($1)?.description ?? "")"
         }
     }
     


liuxuan30:
@jjatie any feedback? But I think only one line change has no perceivable performance gain

Source/Charts/Data/Implementations/ChartBaseDataSet.swift
```diff
@@ -403,7 +403,7 @@ open class ChartBaseDataSet: NSObject, IChartDataSet, NSCopying
     open override var debugDescription: String
     {
         return (0..<entryCount).reduce(description + ":") {
-            $0 + "\n" + (self.entryForIndex($1)?.description ?? "")
+            "\($0)\n\(self.entryForIndex($1)?.description ?? "")"
```
jjatie: For readability
```
"""
\($0)
\(self.entryForIndex($1)?.description ?? "")
"""
```

Stephanv94:
Running xcodebuild with the following flags shows it takes a while to compile
> OTHER_SWIFT_FLAGS="-Xfrontend -debug-time-function-bodies"

2346.70ms	
Pods/Charts/Source/Charts/Data/Implementations/ChartBaseDataSet.swift:404:5	getter debugDescription

2180.13ms	
Pods/Charts/Source/Charts/Data/Implementations/ChartBaseDataSet.swift:404:5	getter debugDescription

Source/Charts/Data/Implementations/ChartBaseDataSet.swift
liuxuan30: @jjatie we are good to merge? if this is the case, I can modify and proceed

PR #4136, opened 2019-09-06
update ChartViewBase.lastHighlighted for ChartViewBase.highlightValue()

mikelow:
### Goals :soccer:
The `lastHighlighted` property of `ChartViewBase` is only set to a non-nil value within `highlightValues`(). We should set the lastHighlighted value for the singular `highlightValue`()  method as well.

### DIFF
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -520,6 +520,8 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
         }
         else
         {
+            self.lastHighlighted = highlight
+            
             // set the indices to highlight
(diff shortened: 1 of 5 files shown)


liuxuan30:
may I ask what's the user case for this issue?

```
    /// The last value that was highlighted via touch.
    @objc open var lastHighlighted: Highlight?
```
this property is originally for touch events, so in the gesture handler, it should already be set. 

Are you calling `highlightValue(_ highlight: Highlight?, callDelegate: Bool)` by yourself, thus the property is not set?

liuxuan30:
it does seems this property is rarely used:
image

@mikelow could you add some details, like when did you met this issue?

danielgindi:
The title and explanation of this PR is completely different from what the PR actually does.

Please split it, so we can discuss each change separately.

PR #4141, opened 2019-09-11
Resolve swift warnings that data was never mutated.

Tanaschita:


liuxuan30:
Thanks!
this should be addressed in https://github.com/danielgindi/Charts/pull/4153.

PR #4153, opened 2019-09-23
Apply Xcode11 changes

liuxuan30:
### Apply Xcode11 changes
automatic remove warnings detected by Xcode11


liuxuan30:
ah crap. get multiple errors, first is rebuild the test case framework with 5.1
```
*** Downloading ios-snapshot-test-case.framework binary at "6.1.0"
***  Skipped installing ios-snapshot-test-case.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.0.1 (swiftlang-1001.0.82.4 clang-1001.0.46.5) and the local version is 5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)."
```

liuxuan30:
it seems iOS 13 render the image has diff more than 0.1% so it's all failing. 
Need more time to take a look if this is false alert, so far I only see the black/grey color different on the same pixel from diff tool:
image

it's a little annoying. If we increase the tolerance rate, more tests would pass.

liuxuan30:
another image:
image
this is very obvious, but still, the same issue.

liuxuan30:
among the false alerts, I actually capture a nasty change of `minimumIntegerDigits`
```
Declaration
var minimumIntegerDigits: Int { get set }
Discussion
By default, this property is set to 0.
```
however for iOS 13, it's default is 1 now
```
(lldb) po [[[NSNumberFormatter alloc] init] minimumIntegerDigits]
1
```

liuxuan30:
I think I'm gonna just rebuild the images, and use the default formatting, since 0.9 makes more sense than `.9` by default. And people used to complain why the default formatter has no prefix zero...

PR #4163, opened 2019-10-02
Adds 'Source' folder as path to the target in Package.swift

abbasmousavi:
### Goals :soccer:
Adds path to the target in Package.swift


liuxuan30:
if we don't add it, what would goes off?
I didn't see people talk about this param when dealing with SPM

PR #4170, opened 2019-10-10
Gradient background Line Charts & Addition of Gradient on Bar Charts Bars

applebyte1992:


liuxuan30:
please fix the UT failures

MobileMon:
This is no good because the gradient is applied as a whole to the chart instead of individually to each bar item

**How it should look:**
Screen Shot 2019-11-11 at 3 31 59 PM

**How it does look:**
Screen Shot 2019-11-11 at 3 32 09 PM

Issue #4133, opened 2019-09-03
Default Colors Should Support Dark Mode [iOS 13]


swupnil:
## What did you do?

ℹ Run project in Xcode 11, on a iOS 13 device.

## What did you expect to happen?

ℹ Black text should invert to white (i.e. Charts should use `UIColor.label` instead of `UIColor.black`).

## What happened instead?

ℹ Black text remains black. 

## Charts Environment

**Charts version/Branch/Commit Number: 3.2.1**
**Xcode version: 11.0**
**Swift version: 4**
**Platform(s) running Charts: iOS**
**macOS version running Xcode: 10.14.4**

liuxuan30:
a pull request is always wanted!

ybeapps:
Those merged commits are very important now since Apple announced that they won't keep apps that don't compile with iOS 13.0 SDK.
Now we want Charts to give us the labels colors for free..
Are you going to make a release soon?
Thanks!!

liuxuan30:
@ybeapps what you mean? The label colors already support Dark mode, see #4171 is merged above?

ybeapps:
I meant that those changes aren’t “released” yet in a new version
If I install from cocoapods I won’t get it.

liuxuan30:
I pushed a new version.

I personally suggest if you like to try new stuff, use swift package or carthage, not relying on we build it for pods

PR #4171, opened 2019-10-10
introduce gracefully degrading abstractions for dark mode for ios and…

motocodeltd:
… macos and use them to draw text

### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4133
https://github.com/danielgindi/Charts/issues/3742

### Goals :soccer:
Introduce dark mode for label and gracefully degrade to black for the older operating systems

Introduce labelOrBlack in NSUIColor. Set it to either NSUIColor.label or black depending on os versions. use it in charts

### Testing Details :mag:
decided against it as it's colours that in some cases are not configurable and hence easily testable

### DIFF
--- a/ChartsDemo-macOS/PlaygroundChart.playground/Pages/BubbleChart.xcplaygroundpage/Contents.swift
+++ b/ChartsDemo-macOS/PlaygroundChart.playground/Pages/BubbleChart.xcplaygroundpage/Contents.swift
@@ -59,7 +59,7 @@ for index in 0..<ITEM_COUNT
 //: ### BubbleChartDataSet
 let set = BubbleChartDataSet(values: entries, label: "Bubble DataSet")
 set.colors = ChartColorTemplates.vordiplom()
-set.valueTextColor = NSUIColor.black
+set.valueTextColor = NSUIColor.labelOrBlack
 set.valueFont = NSUIFont.systemFont(ofSize: CGFloat(10.0))
 set.drawValuesEnabled = true
 set.normalizeSizeEnabled = false
--- a/ChartsDemo-macOS/PlaygroundChart.playground/Pages/LineChart.xcplaygroundpage/Contents.swift
+++ b/ChartsDemo-macOS/PlaygroundChart.playground/Pages/LineChart.xcplaygroundpage/Contents.swift
@@ -60,7 +60,7 @@ rightAxis.granularityEnabled = false
 //: ### Legend
 let legend = chartView.legend
 legend.font = NSUIFont(name: "HelveticaNeue-Light", size: CGFloat(12.0))!
-legend.textColor = NSUIColor.black
+legend.textColor = NSUIColor.labelOrBlack
 legend.form = .square
 legend.drawInside = false
 legend.orientation = .horizontal
--- a/Source/Charts/Charts/ChartViewBase.swift
+++ b/Source/Charts/Charts/ChartViewBase.swift
@@ -98,7 +98,7 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
     @objc open var noDataFont = NSUIFont.systemFont(ofSize: 12)
     
     /// color of the no data text
-    @objc open var noDataTextColor: NSUIColor = NSUIColor.black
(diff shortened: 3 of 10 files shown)


liuxuan30:
hmm, I need some background here.

I'm feeling the ChartsDemo is just a demo. But you guys want the library itself, to support dark mode? More specific, only the label supporting dark mode, or all default colors? 

Besides, I don't think we should support backward compatibility, since it's only started on iOS13/macOS 10.14?

motocodeltd:
Hi,

I'll have as look at the macos failures when I get home (it built fine for me :-)

With the current release version of Charts, when displaying a chart in a dark mode app, the chart itself is perfectly visible (and dark) but all labels and legend etc isn't because they are dark text on top of a dark background. This is solved in iOS13 and Mojave by providing a colour called label, that automatically changes depending on if we're in dark mode or light mode.

The graceful degradation is required because {UIColor/NSColor}.label is only available on mojave, ios13 and not before.

I'll also attach a before and after screenshot later on.

motocodeltd:
I'm attaching a couple of screenshots of our app, [CycleMaps](https://cyclema.ps). Both are of iPhones (simulator) running iOS13 in dark mode. The first one shows the chart without the patch, the second one with the patch. They're using a line chart provided by the Charts library (thanks!). As you can see, without the patch the axis and legend are invinsible.

without patch

with patch

liuxuan30:
ah amazing. i will take a deep look at this topic and review. The change seems pretty straight forward to merge, so if nothing special likely it can be merged soon.

motocodeltd:
that's great, thanks

liuxuan30:
@motocodeltd the changes is fine to me. Just a quick question, you mentioned
>when displaying a chart in a dark mode app, the chart itself is perfectly visible (and dark) 

In my understanding, the current chart color/background color setting like 
```
    internal func initialize()
    {
        #if os(iOS)
        self.backgroundColor = NSUIColor.clear
        #endif
```
is not getting in your way right?

liuxuan30:
@jjatie any feedback? If nothing special I would merge this week, and maybe a minor 3.4.1 release for dark mode

jjatie:
I think this is a much more elegant solution as I mentioned in #3766:

```
extension UIColor {
    @available(iOS, introduced: 9.0, obsoleted: 13.0)
    @available(tvOS, introduced: 9.0, obsoleted: 13.0)
    static var label: UIColor { .black }
}
```
and doing the same for `NSColor`
```
extension NSColor {
    @available(macOS, introduced: 10.11, obsoleted: 10.14)
    static var label: UIColor { .black }
}
```

jjatie:
@liuxuan30 Note the `introduced` is only as far back as the framework supports

motocodeltd:
Hi @liuxuan30 - yes, clear works just fine for me, apologies for the delay.

motocodeltd:
@jjatie this unfortunately does not work for me (I tried it) on my iOS13 devices. The colour comes back as .black.

liuxuan30:
@motocodeltd what's not working? you mean #4175?

motocodeltd:
@liuxuan30 @jjatie 
The solution in #3766 does not work for me.

```
extension UIColor {
    @available(iOS, introduced: 9.0, obsoleted: 13.0)
    @available(tvOS, introduced: 9.0, obsoleted: 13.0)
    static var label: UIColor { .black }
}
```

Always gives me black on iOS13 devices and simulators, which is why I implemented it differently in this PR. In terms of elegance, I'm personally not a big fan of extensions either because they make finding out what the API is slightly harder, but this is more of a matter of personal preference.

liuxuan30:
I'm feeling it's Apple's bug? I have filed a radar today to see what's their opinion.

motocodeltd:
Hi @liuxuan30 - thanks for this. I'm wondering if it's worth merging this, because it gives dark mode in iOS13 and macos mojave & catalina. It can be refactored if and when Apple respond and also allows us not to depend on a branch too :-)

liuxuan30:
@motocodeltd thanks for your patience. I think both PR is ok. But I just want to hear from @jjatie if we merge this, what would be for the following PRs from him, because they are based on the #4175. 
However seems @jjatie is busy right now. So have to wait :( sorry.

PR #4175, opened 2019-10-15
Dark Mode: Primary Labels

jjatie:
Added system dark mode support for primary label colors in the frameworkMoved Color to its own Platform file

### Issue Link :link:
#4133
#3742

### Goals :soccer:
Introduce dark mode for label and gracefully degrade to black for the older operating systems

Introduced `NSUIColor.label` to match modern (UI/App)Kit semantic colors for earlier versions of the respective OSes.

### Testing Details :mag:
N/A


jjatie:
@liuxuan30 Let me know if there are any issues.

jjatie:
#3766 
#4171

Source/Charts/Utils/Platform+Color.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Charts
+//
+//  Created by Jacob Christie on 2019-10-15.
+//
+
+#if canImport(UIKit)
+import UIKit
+
+public typealias NSUIColor = UIColor
+extension UIColor {
```
liuxuan30: code style reminder :)

Source/Charts/Utils/Platform+Color.swift
```diff
@@ -0,0 +1,28 @@
...
+    static var label: UIColor { .black }
+}
+#endif
+
+#if canImport(AppKit)
+
+import AppKit
+
+public typealias NSUIColor = NSColor
+extension NSColor {
```
liuxuan30: code style reminder :)

liuxuan30:
so this is to replace #3766 ?

ChartsDemo-iOS/Swift/Demos/PiePolylineChartViewController.swift
```diff
@@ -6,9 +6,7 @@
 //  Copyright © 2017 jc. All rights reserved.
 //
 
-#if canImport(UIKit)
```
liuxuan30: why we change here? I remember it's for SPM recently

liuxuan30:
@jjatie I searched NSUIColor.black and shows below
image

There is some colors like 
`@objc open var borderColor = NSUIColor.black`
```swift
                attrString = NSMutableAttributedString(string: newValue!)
                attrString?.setAttributes([
                    .foregroundColor: NSUIColor.black,
                    .font: NSUIFont.systemFont(ofSize: 12.0),
                    .paragraphStyle: paragraphStyle
                    ], range: NSMakeRange(0, attrString!.length))
```
```swift
    /// the color drawing borders around the bars.
    open var barBorderColor = NSUIColor.black
```
```swift
open class BarChartDataSet: BarLineScatterCandleBubbleChartDataSet, IBarChartDataSet
{
    private func initialize()
    {
        self.highlightColor = NSUIColor.black
```
do you think we should also change them?

For the border+highlight colors, I think it may be users duty to do it, so we should leave it there;
but for the attributed string, should we change as well?

ChartsDemo-iOS/Swift/Demos/PiePolylineChartViewController.swift
jjatie: It wasn't ever necessary. SPM doesn't compile `ChartsDemo-iOS/`

jjatie:
> @jjatie I searched NSUIColor.black and shows below, I assume you already checked?
> image
> 
> There is some colors like
> `@objc open var borderColor = NSUIColor.black`
> 
> ```swift
>                 attrString = NSMutableAttributedString(string: newValue!)
>                 attrString?.setAttributes([
>                     .foregroundColor: NSUIColor.black,
>                     .font: NSUIFont.systemFont(ofSize: 12.0),
>                     .paragraphStyle: paragraphStyle
>                     ], range: NSMakeRange(0, attrString!.length))
> ```
> 
> ```swift
>     /// the color drawing borders around the bars.
>     open var barBorderColor = NSUIColor.black
> ```
> 
> ```swift
> open class BarChartDataSet: BarLineScatterCandleBubbleChartDataSet, IBarChartDataSet
> {
>     private func initialize()
>     {
>         self.highlightColor = NSUIColor.black
> ```
> 
> do you think we should also change them?
> 
> For the border+highlight colors, I think it may be user's duty to do it, so we should leave it there;
> but for the attributed string, I think it's ours?

I think it's unclear whether or not we should change border and highlight colors, and users have a way to change it themselves. If we decided we should, we can do it in a later PR. Perhaps a separate `Color.border` and `Color.highlight` semantic color extensions.

I think you might have an older copy checked out, because I only found one text element I missed.

liuxuan30:
let's ignore line color and border colors.
seems `@available(tvOS, introduced: 9.0, obsoleted: 13.0)` is not working as expected, always return `.black` color on iOS 13 simulators

Source/Charts/Utils/Platform+Color.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+
+#if canImport(UIKit)
+import UIKit
+
+public typealias NSUIColor = UIColor
+extension UIColor
+{
+    @available(iOS, introduced: 8.0, obsoleted: 13.0)
+    @available(tvOS, introduced: 9.0, obsoleted: 13.0)
```
liuxuan30: API not working

liuxuan30:
reply from Apple
```
Apple
Nov 7, 2019 at 8:58 AM
Thanks for your report. `obsoleted` is not designed to work the way you want it to work here—it’s supposed to indicate that the symbol may have been removed by iOS 13, but it will still be used if it’s there. We don’t currently have a “polyfill” feature like the one you’re looking for, but a Swift Evolution pitch adding one might be well-received.

Given the language we have today, the original `labelOrBlack` approach discussed in the PR is probably the cleanest solution. If you’re dead-set on using the name “label”, my best idea is to redeclare the property in an Objective-C category and use the Objective-C runtime from a +load method to conditionally install the implementation if it’s not already present.
```

PR #4178, opened 2019-10-15
Platform separation

jjatie:
There are no code changes here. I only split out the different pieces of platform to make it more easily maintainable.

Must be merged after #4175


Source/Charts/Utils/Platform+Color.swift
```diff
@@ -0,0 +1,53 @@
...
+//  Created by Jacob Christie on 2019-10-15.
+//
+
+#if canImport(UIKit)
+import UIKit
+
+public typealias NSUIColor = UIColor
+private func fetchLabelColor() -> UIColor
+{
+    if #available(iOS 13, tvOS 13, *)
```
liuxuan30: is this legacy? After closing #4175 I thought we don't use `available` any more

Source/Charts/Utils/Platform+Touch Handling.swift
```diff
@@ -0,0 +1,130 @@
+//
+//  Platform+Touch Handling.swift
```
liuxuan30: I was thinking if we should have space in file name

jjatie: No reason not to

Source/Charts/Utils/Platform+Color.swift
jjatie: No, this matches the implementation that was accepted and is in master.

liuxuan30: Have you tested it? I don't have an iOS 12 device at home. If need test on older iOS, I could do it tomorrow

liuxuan30: FYI I just get reply from Apple today regarding #4175 
>Apple
Nov 7, 2019 at 8:58 AM
Thanks for your report. `obsoleted` is not designed to work the way you want it to work here—it’s supposed to indicate that the symbol may have been removed by iOS 13, but it will still be used if it’s there. We don’t currently have a “polyfill” feature like the one you’re looking for, but a Swift Evolution pitch adding one might be well-received.

>Given the language we have today, the original `labelOrBlack` approach discussed in the PR is probably the cleanest solution. If you’re dead-set on using the name “label”, my best idea is to redeclare the property in an Objective-C category and use the Objective-C runtime from a +load method to conditionally install the implementation if it’s not already present.

jjatie: This is the exact implementation that was accepted, and is referred to as the better solution by Apple: #4171

PR #4197, opened 2019-10-26
Added Default Case, TableView Delegate Improvemnets, added default va…

dhanrajchavan7:
…lues

### Issue Link :link:

No Default Cases provided

### Goals :soccer:

Improve User Experience
Improve all possible conditions by adding every possibility

Added Default Case to Switch To add default functionality to that condition
Added Table View Delegate Method to improve user experience


liuxuan30:
I didn't think it's quite useful for the changes.. It's just a demo. And the default value seems would never happen to the demo, unless you add something yourself.

Issue #4037, opened 2019-06-21
leftAxis line disappears when filling area between two line charts.


hachinobu:
## What did you do?
i use filling area between two line charts.

## What did you expect to happen?
A fill should appear between the two line graphs, and the leftAxis line should appear.

## What happened instead?
leftAxis line disappears.

|not filled|filled|
|:--:|:--:|
|![Simulator_Screen_Shot_-_iPhone_X_-_2019-06-21_at_11_41_31](https://user-images.githubusercontent.com/1317847/59896351-2e1e1380-9423-11e9-8515-a077341ca8e5.png)|![Simulator_Screen_Shot_-_iPhone_X_-_2019-06-21_at_11_42_06](https://user-images.githubusercontent.com/1317847/59896356-324a3100-9423-11e9-9d1c-c49cc1589d1e.png)|

ℹ Please replace this with of what happened instead.  

## Charts Environment

**Charts version/Branch/Commit Number: 3.3.0**
**Xcode version:10.2**
**Swift version:5.0**
**Platform(s) running Charts:iOS 12.2**
**macOS version running Xcode: 10.14.5**

Is there an easy way to solve this problem?

Is it better to customize it according to IFillFormatter or make it inherit from LineChartRenderer?

hachinobu:
To solve this problem, it is necessary to advance the timing of calling renderer.drawData(context: context) in the draw(_ rect: CGRect) method of BarLineChartViewBase.

However, I am worried about the degradation caused by this change.

I will send a PR if there is no problem, how is it?

```BarLineChartViewBase.swift

    // BarLineChartViewBase.swift
    open override func draw(_ rect: CGRect)
    {
        super.draw(rect)

        guard data != nil, let renderer = renderer else { return }
        
        let optionalContext = NSUIGraphicsGetCurrentContext()
        guard let context = optionalContext else { return }

        // execute all drawing commands
        drawGridBackground(context: context)
        renderer.drawData(context: context) // <-- solve this issue.

```

![Simulator Screen Shot - iPhone X - 2019-06-22 at 16 18 54](https://user-images.githubusercontent.com/1317847/59960825-73b50c00-9509-11e9-9a0a-b5b9dab40536.png)

liuxuan30:
please send us a PR, thanks

PR #4199, opened 2019-10-29
changed to display leftAxis line even if it fills between two line gr…

hachinobu:
…aphs.

### Issue Link :link:
#4037 

### Goals :soccer:
The leftAxis line is displayed even if you fill between two line graphs.

|Before|After|
|:--:|:--:|
|Simulator_Screen_Shot_-_iPhone_11_Pro_Max_-_2019-10-29_at_13_49_23|Simulator_Screen_Shot_-_iPhone_11_Pro_Max_-_2019-10-29_at_13_51_44|

Changed timing to call `renderer.drawData (context: context)`

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -181,7 +181,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
 
         // execute all drawing commands
         drawGridBackground(context: context)
-        
+        renderer.drawData(context: context)
 
         if _autoScaleMinMaxEnabled
         {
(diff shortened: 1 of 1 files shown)


liuxuan30:
Hi there,

Sorry for the delay. But I'm feeling, you are filling all the contentRect, instead of the area of your charts, which seems not making sense?

Source/Charts/Charts/BarLineChartViewBase.swift
```diff
@@ -181,7 +181,7 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
 
         // execute all drawing commands
         drawGridBackground(context: context)
-        
+        renderer.drawData(context: context)
```
liuxuan30: if you pull this up before clip, you would get multiple UT failures.
You have to restrict the filling area between the lines, not the whole chart.

PR #4211, opened 2019-11-12
Take into account the rendered size for the labels

mmazzei:
Take into account the rendered size for the labels instead of the char count for size calculations.

### Goals :soccer:
When setting up the margin size for the axis labels, right now it uses the string characters count. In some cases it is not the right approach (example: strings with multilines, or "iiiiii" vs "kkkkkk".

I've replaced the string characters count by the calculated size given the attributes.


danielgindi:
There are multiple features here in one PR. You cannot sneak one feature on the back of another...
Also - regarding the label measuring, that's intended. As measuring every label - means doing the work of virtually rendering it to figure out the size, it has a cost. So if anyone has use for this feature, it should be optional and should not regress.

PR #4265, opened 2020-01-16
Improve render performance if draw circles is disabled

ekurutepe:
### Goals :soccer:
Improve line graph drawing performance for data sets with multiple thousands of points.

Draw circles is an expensive function. By not running it if `drawCirclesEnabled` returns false, we're able to easily draw a line graph >7000 data points on an iPhone X with 60fps.

### Testing Details :mag:
None

### DIFF
--- a/Source/Charts/Renderers/LineChartRenderer.swift
+++ b/Source/Charts/Renderers/LineChartRenderer.swift
@@ -562,12 +562,12 @@ open class LineChartRenderer: LineRadarRenderer
 
         for i in 0 ..< dataSets.count
         {
-            guard let dataSet = lineData.getDataSetByIndex(i) as? ILineChartDataSet else { continue }
-            
-            if !dataSet.isVisible || dataSet.entryCount == 0
-            {
-                continue
-            }
+            guard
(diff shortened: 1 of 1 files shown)


Source/Charts/Renderers/LineChartRenderer.swift
```diff
@@ -562,12 +562,12 @@ open class LineChartRenderer: LineRadarRenderer
...
         {
-            guard let dataSet = lineData.getDataSetByIndex(i) as? ILineChartDataSet else { continue }
-            
-            if !dataSet.isVisible || dataSet.entryCount == 0
-            {
-                continue
-            }
+            guard
+                let dataSet = lineData.getDataSetByIndex(i) as? ILineChartDataSet,
+                dataSet.drawCirclesEnabled,
```
4np: 👍

4np: The ` if !dataSet.isDrawCirclesEnabled` in the `for j in _xBounds` loop can then be removed.

ekurutepe: Good point! Updated the PR to fix it.

Issue #4033, opened 2019-06-20
Height of Half Pie Chart (white space in the chart where the removed bottom half from pie chart should be)


ratala1:
I am using Half Pie Chart in my project. When I reduce the height component of the pieChart, the chart size also goes down. attaching screenshots for reference. How do I use HalfPieChart without the extra space below the chart?

Screenshot 1:
Height = Width
![Simulator Screen Shot - iPhone Xʀ - 2019-06-20 at 15 31 10](https://user-images.githubusercontent.com/11667120/59841594-441fcb80-9372-11e9-8dd0-aa51f1fb558f.png)

Screenshot 2:
Height = Width/2
![Simulator Screen Shot - iPhone Xʀ - 2019-06-20 at 15 28 20](https://user-images.githubusercontent.com/11667120/59841621-5437ab00-9372-11e9-8008-08c23a65db22.png)

Thanks

kamokaizen:
did you solve it? same problem i have and can not resolved yet.

ratala1:
@kamokaizen no

kamokaizen:
to solve with workaround; give below item of chart a minus top padding.  i m using snapkit;
make.top.equalTo(chartView.snp.bottom).offset(-100) // 100 mean height/2

JMSerruya:
Hi @ratala1 , did you found a fix to this issue? facing the same problem.

liuxuan30:
I'm feeling it's kind of a defect, that it still counts as a full pie chart view and reserves the bottom half. The workaround seems correct by @kamokaizen, which is to reduce the offset. However, a true solution is not that easy - and we might need someone to look into it.

liuxuan30:
The tricky part here is if you want the radius to be larger, you have to calculate the radius properly;

for example, I faked the radius with ChartsDemo half pie chart:
```swift
    internal override func calculateOffsets()
    {
        super.calculateOffsets()
        
        // prevent nullpointer when no data set
        if _data === nil
        {
            return
        }
        
        let radius = maxAngle > 180 ? diameter / 2.0 : (_viewPortHandler.contentRect.size.width > _viewPortHandler.contentRect.size.height ? diameter : diameter / 2)
        
        let c = self.centerOffsets
        
        let shift = (data as? PieChartData)?.dataSet?.selectionShift ?? 0.0
        
        // create the circle box that will contain the pie-chart (the bounds of the pie-chart)
[...]

PR #4266, opened 2020-01-19
address #4033 draw half pie chart more accurate

liuxuan30:
address #4033. when it's half pie and width is larger than height, draw bigger pie chart. 

I‘m not in a hurry to merge before it's fully tested and reviewed.

The key idea is to fake the radius and adjust the circle box.

when screen width <= height, it should be the same as a full pie chart, since the radius should be diameter/2 anyway;

but when screen width > height, the radius is min(width, height) to use the maximum space

before:
image

after:
image

the portrait portion (width < height) should not be affected.

### DIFF
--- a/Source/Charts/Charts/PieChartView.swift
+++ b/Source/Charts/Charts/PieChartView.swift
@@ -132,7 +132,7 @@ open class PieChartView: PieRadarChartViewBase
     /// if width is larger than height
     private var widthLarger: Bool
     {
-        return _viewPortHandler.contentRect.size.width > _viewPortHandler.contentRect.size.height
+        return _viewPortHandler.contentRect.orientation == .landscape
     }
 
     /// adjusted radius. Use diameter when it's half pie and width is larger
--- a/Source/Charts/Charts/PieChartView.swift
+++ b/Source/Charts/Charts/PieChartView.swift
@@ -128,6 +128,26 @@ open class PieChartView: PieRadarChartViewBase
         
         drawMarkers(context: context)
     }
+
+    /// if width is larger than height
+    private var widthLarger: Bool
+    {
+        return _viewPortHandler.contentRect.size.width > _viewPortHandler.contentRect.size.height
+    }
+
+    /// adjusted radius. Use diameter when it's half pie and width is larger
+    private var adjustedRadius: CGFloat
+    {
+        return maxAngle <= 180 && widthLarger ? diameter : diameter / 2.0
+    }
+
+    /// true centerOffsets considering half pie & width is larger
+    private func adjustedCenterOffsets() -> CGPoint
+    {
+        var c = self.centerOffsets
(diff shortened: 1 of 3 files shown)


Source/Charts/Charts/PieChartView.swift
```diff
@@ -128,6 +128,26 @@ open class PieChartView: PieRadarChartViewBase
         
         drawMarkers(context: context)
     }
+
+    /// if width is larger than height
+    private var widthLarger: Bool
```
jjatie: I believe there is already an extension on `CGSize` and `CGRect` to provide an aspect ratio which would simplify this to
```
private var widthLarger: Bool { _viewPortHandler.contentRect.aspect > 1
```

jjatie: It could be useful to provide an orientation instead
```
enum Orientation { case portrait, landscape }

extension CGSize {
    var orientation: Orientation { aspect > 1 }
}

extension CGRect {
    var orientation: Orientation { size.orientation }
}
```

Source/Charts/Charts/PieChartView.swift
```diff
@@ -138,25 +158,88 @@ open class PieChartView: PieRadarChartViewBase
...
+    @objc open override func angleForPoint(x: CGFloat, y: CGFloat) -> CGFloat
+    {
+        let c = adjustedCenterOffsets()
+
+        let tx = Double(x - c.x)
+        let ty = Double(y - c.y)
+        let length = sqrt(tx * tx + ty * ty)
+        let r = acos(ty / length)
+
+        var angle = r.RAD2DEG
```
jjatie: It is probably worth making an `Angle` type in the future much like [SwiftUI.Angle](https://developer.apple.com/documentation/swiftui/angle)

Source/Charts/Charts/PieChartView.swift
liuxuan30: Hi @jjatie sorry for the deplay. Finally got a little time to work on this. I searched it seems current master didn't have  aspect?

Source/Charts/Charts/PieChartView.swift
liuxuan30: yep. but currently this should serve

Source/Charts/Charts/PieChartView.swift
liuxuan30: @jjatie mind take a look recently? I want to merge this and release a new build for Swift 5.2. I didn't see `aspect` so I use the width > height for now

liuxuan30: @jjatie plan to merge this this week and release new build for Swift 5.2.

Would you mind take a look?

liuxuan30:
@danielgindi have time to take a look? want to be careful

danielgindi:
Seems fine to me!

PR #4271, opened 2020-01-23
Call chartViewDidEndPanning on when *panning* is ended

danielgindi:
### Goals :soccer:
Straighten the semantics of the event.

Currently the relatively new protocol method is called in the end of the gesture, even if the gesture does not mean "pan". This could cause unexpected bugs.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -787,6 +787,8 @@ open class BarLineChartViewBase: ChartViewBase, BarLineScatterCandleBubbleChartD
                 }
                 
                 _isDragging = false
+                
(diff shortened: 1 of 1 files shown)


odanu:
It looks like this PR introduced a regression. The `chartViewDidEndPanning` stopped being called when user finishes the interaction with the chart.

odanu:
In the latest version it looks like that `chartViewDidEndPanning` might be called only and only when `__isDragging` is true.
Code comments say that `isDragging` is true when:
```
// If drag is enabled and we are in a position where there's something to drag:
            //  * If we're zoomed in, then obviously we have something to drag.
            //  * If we have a drag offset - we always have something to drag
```
But, panning is not always dragging.

liuxuan30:
will  revert

PR #4275, opened 2020-01-24
Fix for #4274 string comparison issue in ChartData::getDataSetByLabel

### LINKED ISSUE #4274: ChartData::getDataSetByLabel always return the first IChartDataSet when ignorecase = true
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## Issue 
It appears that this commit (https://github.com/danielgindi/Charts/pull/3892/files) added the comparison `label.caseInsensitiveCompare(label)` to `ChartData::getDataSetByLabel`.

This will always be true, so statements like `chartData.getDataSetByLabel("myLabel", ignorecase: true)` will return whatever `IChartDataSet` is at index 0.

## Effected code

```
 internal func getDataSetIndexByLabel(_ label: String, ignorecase: Bool) -> Int
    {
        // TODO: Return nil instead of -1
        if ignorecase
        {
            return dataSets.firstIndex {
                guard let label = $0.label else { return false }
                return label.caseInsensitiveCompare(label) == .orderedSame
            } ?? -1
        }
        ...
    }
```

## Follow-up
I'm happy to submit a pull-request to fix this issue 😺 

## Charts Environment

**Charts version/Branch/Commit Number:** 3.4.0
**Xcode version:** 11.3.1
**Swift version:** 5.0
**macOS version running Xcode:** 10.15.2

PeterKaminski09:
Fixes an issue with string comparison in ChartData for finding a dataset by its label (Fixes #4274)

ChartData::getDataSetByLabel has a bug where the string comparison is checking the same string against itself due to the functions parameter having the same name as a variable within the closures scope. This change renames the variable to fix the issue.

### Issue Link :link:
#4274 

### Goals :soccer:
- Fixes the bug
- Adds some unit tests to prevent the relevant issue from regressing

- No major changes here, just a quick fix

### Testing Details :mag:
- Added some tests that check case sensitivity
- Added some tests that ignore case
- Added some tests that check for the expected nil case

### DIFF
--- a/ChartDataTests.swift
+++ b/ChartDataTests.swift
@@ -1,0 +1,69 @@
+//
+//  ChartDataTests.swift
+//  ChartsTests
+//
+//  Created by Peter Kaminski on 1/23/20.
+//
+
+import XCTest
+@testable import Charts
+
+class ChartDataTests: XCTestCase {
+    
+    var data: ScatterChartData!
+
+    private enum SetLabels {
+        static let one = "label1"
+        static let two = "label2"
+        static let three = "label3"
+        static let badLabel = "Bad label"
+    }
+    
+    override func setUp() {
+        super.setUp()
+        
+        let setCount = 5
+        let range: UInt32 = 32
+        let values1 = (0..<setCount).map { (i) -> ChartDataEntry in
+            let val = Double(arc4random_uniform(range) + 3)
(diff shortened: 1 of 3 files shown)


danielgindi:
What a funny bug. Thank you!

PR #3754, opened 2018-11-21
Add support for rounded corners for BarChart

skarol:
### Issue Link :link:
Rounding the corners of bars #1066

### Goals :soccer:
image
* Add ability to draw rounded corners for `BarChartView` and `HorizontalBarChartView`.
* Support also stacked bar charts.
* Corner radius should be calculated automatically, depending on chart's width and height.
* To use it simply add `dataSet.roundedCorners = [.allCorners]`.

### Implementation Details :construction:
* Add `roundedCorners` property for `BarChartDataSet`, where you can specify which corners should be rounded.
* Introduce changes in `drawDataSet` function in `BarChartRenderer` and `HorizontalBarChartRenderer`, where `UIBezierPath` is created to clip bar filling.
* Similar changes apply to `drawHighlighted` function in `BarChartRenderer`.

### Testing Details :mag:
Add `testRoundedCorners` test to `BarChartTests`.

Issue #3701, opened 2018-10-19
xAxis rotated labels cut off


rgmez:
## What did you do?

I want to rotate xAxis labels because I need to put large text.

## What did you expect to happen?

I expect to show all text, I thought that I can show it with "fitScreen" function.

## What happened instead?

Labels with large width, are cutting off :(

## Code:

        charView.xAxis.labelRotationAngle = -90.0
        charView.setExtraOffsets(left: 30, top: 30, right: 30, bottom: 30)
        charView.fitScreen()

liuxuan30:
wow, never tried out `fitScreen()`.. maybe you find a new bug... can you dig more? e.g. will it be cut off if you turn off rotation.

kuwait:
I think there is an error here @liuxuan30. 

```
 let longest = xAxis.getLongestLabel()
 let labelSize = longest.size(attributes: [NSFontAttributeName: xAxis.labelFont])
```

Your calculating the right size for the wrong longest label. getLongestLabel counts the Chars of that string but there are characters bigger than other depending on the font you use. You should get the longestLabelSize with the font you are using.

liuxuan30:
@kuwait are you able to send us a pull request? Thanks

4np:
@liuxuan30 : I added a fix for this to #4297

PR #4297, opened 2020-02-25
Update the renderers with dedicated atomic rendering methods in order to make rendering overrides easier to implement.

4np:
# Context

Often when implementing charting features, there is a requirement to implement certain UI changes. The more simple changes can often be accomplished by subclassing a particular renderer and overriding some of its methods, but more elaborate changes can be hard to accomplish.

While many methods are marked with `open` visibility and can theoretically be overloaded, in practice this is not as straightforward as most of those methods either call other methods or use variables that have `private` or `internal` visibility. 

Consider the use case where you want to fill a bar chart with a pattern instead of a color. In order to do so, you would need to subclass the `BarChartRenderer` and override `drawDataSet(context:, dataSet:, index:)` (or `drawHighlighted(context:, indices:)` to update the highlight rendering). 

There are a couple of challenges with this approach:
- `drawDataSet(context:, dataSet:, index:)` performs quite some calculations you _really_ do not want to replicate and want to keep '_in the charting framework_' (separation of concerns).
- these methods call a number of other methods that have `internal` or `private` visibility, which you cannot call from your subclassed renderer (for example `prepareBuffer(dataSet:, index:)` or `BarLineScatterCandleBubbleRenderer.isInBoundsX(entry:, dataSet:)`), which in turn requires you to duplicate _those_ implementation as well.
- similarly these methods use a number of variables that have `internal` or `private` visibility (e.g. `_barShadowRectBuffer`, `_buffers`, etcetera), again requiring unnecessary code duplications.
- you want to focus on the rendering override, instead of re-implementing a lot of code in your subclassed renderer that, in essence, does not really matter for what you aim to accomplish.
- because of all the code duplication from the base class, up-taking a new `Charts` version is cumbersome and requires diffing duplicate logic for changes.

## Solution

This Pull Request lifts the rendering logic out of the methods into their own dedicated methods with `open` visibility. This way it becomes really easy to create a custom renderer that implements custom rendering behavior, without the need of having to duplicate complex logic or re-implement `private` / `internal` methods and/or variables. It allows you to focus on what matters, while keeping your subclassed renderer focussed and simple.

## Example: Bar Chart Pattern Fill.

This is an example of creating a custom bar chart renderer that will fill bars with a circular pattern:

```swift
/// Custom bar chart renderer that fills bars with a circle based pattern.
class PatternFilledBarChartRenderer: BarChartRenderer {
    override func renderFill(with color: UIColor, for rect: CGRect, in context: CGContext,  dataSet: IBarChartDataSet) {
        let patternSize = CGSize(width: 10, height: 10)
        let patternCallback: CGPatternDrawPatternCallback = { (pointer, context) -> Void in
            context.addArc(center: CGPoint(x: 5, y: 5), radius: 5, startAngle: 0, endAngle: CGFloat(2.0 * .pi), clockwise: true)
            context.fillPath()
        }
        var callbacks = CGPatternCallbacks(version: 0, drawPattern: patternCallback, releaseInfo: nil)
        let pattern = CGPattern(info: nil,
                                bounds: CGRect(origin: CGPoint(x: 0, y: 0), size: patternSize),
                                matrix: .identity,
                                xStep: patternSize.width,
                                yStep: patternSize.height,
                                tiling: .constantSpacing,
                                isColored: false,
                                callbacks: &callbacks)
        let patternColorSpace = CGColorSpace(patternBaseSpace: CGColorSpaceCreateDeviceRGB())
        let colorComponents = UnsafeMutablePointer<CGFloat>.allocate(capacity: 4)
        color.getRed(&colorComponents[0], green: &colorComponents[1], blue: &colorComponents[2], alpha: &colorComponents[3])

        defer {
            colorComponents.deinitialize(count: 4)
            colorComponents.deallocate()
        }

        context.saveGState()
        context.setFillColorSpace(patternColorSpace!) //swiftlint:disable:this force_unwrapping
        context.setFillPattern(pattern!, colorComponents: colorComponents) //swiftlint:disable:this force_unwrapping
        context.fill(rect)
        context.restoreGState()
    }
}
```

Implementing this rendering override can now be done by overriding a single method in a straightforward manner, whereas before you needed to copy over pretty much the complete implementation of `BarChartRenderer.swift` (and more) while making small alterations to accomplish the same.

### Screenshot:

Screen Shot 2020-02-25 at 16 12 43

### Lastly

It would be greatly appreciated if this could make it into a newly drafted release in the not so distant future 🙏🏻 😉


jjatie:
I'm all for this, but there has been resistance to adding more function calls in the past. I believe the justification for this hasn't been valid since about Swift 3, but as a result @danielgindi will have to decide.

4np:
Thanks for your feedback and thoughts @jjatie 👍

4np:
@danielgindi : any thoughts? This has been open for a while...

4np:
@danielgindi @liuxuan30 : I updated the PR to the latest revision from `master`.

Updated example code (which you can add to the `BarChartViewController.swift` to test):
- add this code to the bottom of `BarChartViewController.swift`
- change `@IBOutlet var chartView: BarChartView!` to `@IBOutlet var chartView: PatternFilledBarChartView!`
- change the view's class from `BarChartView ` to `PatternFilledBarChartView ` in `BarChartViewController.xib`

```swift
/// Custom bar chart view that fills bars with a circle based pattern.
class PatternFilledBarChartView: BarChartView {
    override init(frame: CGRect) {
        super.init(frame: frame)
        
        let chartAnimator = renderer?.animator ?? Animator()
        renderer = PatternFilledBarChartRenderer(dataProvider: self,
                                                 animator: chartAnimator,
                                                 viewPortHandler: viewPortHandler)
    }
    
    required init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
        
        let chartAnimator = renderer?.animator ?? Animator()
        renderer = PatternFilledBarChartRenderer(dataProvider: self,
                                                 animator: chartAnimator,
                                                 viewPortHandler: viewPortHandler)
    }
}

/// Custom bar chart renderer that fills bars with a circle based pattern.
class PatternFilledBarChartRenderer: BarChartRenderer {
    override func renderFill(with color: NSUIColor, for rect: CGRect, in context: CGContext, dataSet: BarChartDataSetProtocol, entry: BarChartDataEntry?) {
        let patternSize = CGSize(width: 10, height: 10)
        let patternCallback: CGPatternDrawPatternCallback = { (pointer, context) -> Void in
            context.addArc(center: CGPoint(x: 5, y: 5), radius: 5, startAngle: 0, endAngle: CGFloat(2.0 * .pi), clockwise: true)
            context.fillPath()
        }
        var callbacks = CGPatternCallbacks(version: 0, drawPattern: patternCallback, releaseInfo: nil)
        let pattern = CGPattern(info: nil,
                                bounds: CGRect(origin: CGPoint(x: 0, y: 0), size: patternSize),
                                matrix: .identity,
                                xStep: patternSize.width,
                                yStep: patternSize.height,
                                tiling: .constantSpacing,
                                isColored: false,
                                callbacks: &callbacks)
        let patternColorSpace = CGColorSpace(patternBaseSpace: CGColorSpaceCreateDeviceRGB())
        let colorComponents = UnsafeMutablePointer<CGFloat>.allocate(capacity: 4)
        color.getRed(&colorComponents[0], green: &colorComponents[1], blue: &colorComponents[2], alpha: &colorComponents[3])

        defer {
            colorComponents.deinitialize(count: 4)
            colorComponents.deallocate()
        }

        context.saveGState()
        context.setFillColorSpace(patternColorSpace!) //swiftlint:disable:this force_unwrapping
        context.setFillPattern(pattern!, colorComponents: colorComponents) //swiftlint:disable:this force_unwrapping
        context.fill(rect)
        context.restoreGState()
    }
}
```

### Screenshots

Afbeelding

4np:
Added a fix for #3701 that will make sure the labels are handled correctly.

4np:
@liuxuan30 : just a quick look at open issue, it looks like this PR will solve quite a number of them. Rather extending the charting library with specific logic, it allows people to create their custom renderers and override rendering to fit their specific needs.

bivant:
The only issue I see here is a performance question. Does it change with this implementation and how much if yes. Other than that this is a perfect solution to many-many cases. I would say it should be done for all renderers (or should have such option) because recently had to modify xAxis one to match designer idea of the beauty :) 
https://github.com/danielgindi/Charts/pull/3754

4np:
It shouldn't really, as it is only introducing dedicated `render` methods rather than doing the rendering inline. It does not add additional complexity or computational overhead. It depends, of course, on what you do yourself inside a custom renderer.

4np:
@danielgindi @liuxuan30 @jjatie: anybody?

4np:
@danielgindi @liuxuan30 @jjatie: bump...

scinfu:
When this will be merged?

4np:
@scinfu see #4569

liuxuan30:
@jjatie what you think for this PR now?

liuxuan30:
I had a quick look and it seems great, if we don't consider the android part; yes it's kind of the time we need to think about iOS and android interface again.

jjatie:
I think that as Charts is today, the implementation makes sense. Two notes:
1. It seems heavy to subclass the chart in the example. Why not just set the renderer on a BarChartView?
2. It feels strange to have to drill down the existing renderer to pull out its animator. Looking at our implementation, it looks like we can just pass in the animator directly from the chart which feels better.

If we agree that we should move forward with this, we need to audit for inline-ability. We also need to ensure all `DataRenderer` types are updated to reflect this model. It is also a good time to audit the naming of a lot of these methods and their parameters. (I would say `render*()` should be all renamed `draw*()` to be more in line with UIKit convention and to be more accurate)

4np:
@jjatie : Yes true, you don't need to subclass `BarChartView`, I though it would just be a good example class for people to understand. If the animator is available in the chart that should work well, although I think it wasn't available in the chart itself? Not sure though. 

May I suggest that this perhaps does not merge to `danielgindi:master` but rather to a branch you create of master? Then perhaps you can make the changes you want to make to this (for example `render` > `draw` or perhaps some of the method signatures)? Or do you want me to make some changes on my end?

liuxuan30:
@danielgindi  please take a look at this PR

danielgindi:
I think this is something that I would actually want to port Android too. Will try to dedicate some time soon

kelvinofula:
Quick question, how do I apply the same logic to achieve a piechart with rounded corners? I've been really trying it for a long time without success. Any help will really be much appreciated. Thanks a lot.

liuxuan30:
@kelvinofula do not hijack.

@jjatie I think we reach an agreement with @danielgindi  that, we could go ahead to make some iOS features and refactorings.

AppCodeZip:
Recently I have been working on a project, to which I have added a chart library and display a beautiful rounded bar chart in swift. We can achieve this functionality easily. Please check this [demo](https://www.appcodezip.com/2021/03/rounded-bars-in-iOS-Charts.html).
Screenshot 2021-03-27 at 7 42 42 PM

4np:
@AppCodeZip : Do not hijack this Pull Request, your example is exactly what this PR is trying to prevent.

liuxuan30:
@jjatie I'm sorry to say that, I got a layoff and find a new job now, so it would be super busy for the rest of the year, tons of pressure from corp projects.. I might be missing for a few months (and already) 

If you think the PR is good to go, probably you want to go ahead without my review.

4np:
Ouch @liuxuan30, I am sorry to hear that. I can imagine your mind is not with this repository. Best of luck with your job hunt. Perhaps [iOS Dev Jobs](https://iosdevjobs.com/) might be useful to you?

liuxuan30:
@4np haha thank you. I think my new job is lovely and challenging, especialy there are 6 months trial period for me, so I have to spend all my time on that. I really wanted to spend more time on Charts but recently I guess I couldn't afford. 

So I have to stay quiet for some time. I trust @jjatie and @danielgindi could take care of some PRs.

4np:
@pmairoldi @rph8 @kcome @liuxuan30 @jjatie @danielgindi Can I please get you to take another look at this PR?

417-72KI:
@4np @pmairoldi @liuxuan30 @jjatie @danielgindi 
Any updates?

4np:
@pmairoldi @liuxuan30 @jjatie @danielgindi 

This has been open for close to 3 years. Would still like to see this make it into the framework.

4np:
We just celebrated the 4th anniversary of this PR! We have since moved onto Swift Charts, so feel free to close if this will never be implemented...

4np:
We just celebrated the 4th anniversary of this PR! 🥳🎊  As we have since migrated over to [Swift Charts](https://developer.apple.com/documentation/charts) and it doesn't look like this PR will ever be merged, I am hereby closing it.

vadimbelyaev:
maxresdefault

PR #4321, opened 2020-03-30
Fix warnings on current code base

liuxuan30:
Fix warnings for vars;
Fix unused instants;
Remove a duplicated function NSUIMainScreen()

### DIFF
--- a/Source/Charts/Utils/Platform.swift
+++ b/Source/Charts/Utils/Platform.swift
@@ -234,7 +234,7 @@ extension NSScrollView
 
 #endif
 
-func NSUIMainScreen() -> NSUIScreen?
+extension NSUIScreen
 {
-    return NSUIScreen.main
+    class var nsuiMain: NSUIScreen? { .main }
 }
--- a/Source/Charts/Utils/Platform.swift
+++ b/Source/Charts/Utils/Platform.swift
@@ -39,11 +39,6 @@ extension UIScreen
     }
 }
 
-func NSUIMainScreen() -> NSUIScreen?
-{
-    return NSUIScreen.main
-}
-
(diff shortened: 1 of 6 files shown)


Source/Charts/Utils/Platform.swift
```diff
@@ -39,11 +39,6 @@ extension UIScreen
     }
 }
 
-func NSUIMainScreen() -> NSUIScreen?
```
liuxuan30: @jjatie I found this func is duplicated under two macros iOS & macOS; I manually tested on my iOS simulator and macOS, it seems we only need one outside these macros. We are safe to delete one right? I don't come up a case that we need to put them into two places

jjatie: Yea, we don't need both. Really this should be an extension on `NSUIScreen`
```
extension NSUIScreen {
    var nsuiMain: Self? { .main }
}
```

liuxuan30: @jjatie I have pushed a new commit. but mine is
```
class var nsuiMain: NSUIScreen? { .main }
```
is there any reason to use `Self?` here? Because if I use `Self?`, Xcode would complain
>Type 'Self?' has no member 'main'

and I have to use something like
```
class var nsuiMain: Self? { NSUIScreen.main as? Self }
```
(I can't even use `{ .main as? Self }` ) So I think use NSUIScreen explicitly makes it cleaner.

gligorkot:
@liuxuan30 would be good to have this merged in and released so we can eliminate XCode 11.4 warnings. 👍

gligorkot:
Champion! Thank you @liuxuan30

Issue #4323, opened 2020-03-31
Distribute Charts as XCFramework


gk-brown:
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Wrote shell script to build Charts as an XCFramework (PR to be submitted).

## What did you expect to happen?

n/a

## What happened instead?

n/a

## Charts Environment

master
Xcode 11.4
Swift 5.2
iOS
macOS 10.15.4

## Demo Project

n/a

gk-brown:
This is a follow-up to comments I posted on this earlier (currently closed) ticket:

https://github.com/danielgindi/Charts/issues/4317

gk-brown:
Pull request submitted: https://github.com/danielgindi/Charts/pull/4325

Issue #4317, opened 2020-03-25
Module compiled with Swift 5.1.2 cannot be imported by Swift 5.2 compiler


afridikhalid:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

ℹ Updated to Xcode 11.4 and run Carthage update  

## What happened instead?

ℹ importing chart Xcode gives the following error: 
Module compiled with Swift 5.1.2 cannot be imported by Swift 5.2 compiler: /User/username/ios/Carthage/Build/iOS/Charts.framework/Modules/Charts.swiftmodule/arm64-apple-ios.swiftmodule  

## Charts Environment

**Charts version/Branch/Commit Number:**
**Xcode version: 11.4
**Swift version: 5.2
**Platform(s) running Charts: iOS
**macOS version running Xcode:10.15.4

liuxuan30:
damn. I will see if we can release a new build. Such things still waiting to be done after the full ABI support is enabled.

gk-brown:
Have you considered converting to an XCFramework?

gk-brown:
I was able to resolve the compilation issues. They were all method visibility issues (internal vs. open/public). Would be happy to submit a pull request.

PR #4325, opened 2020-03-31
Add shell script for building XCFramework.

gk-brown:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4323

### Goals :soccer:
Build Charts as an XCFramework:
https://developer.apple.com/videos/play/wwdc2019/416/
https://developer.apple.com/documentation/xcode_release_notes/xcode_11_release_notes (see Build System/New Features)

No architectural changes. Method visibility changes were required to resolve compilation errors when generated framework is added to an application.

### Testing Details :mag:
_package.sh_ script generates _Charts.xcframework_ along with _Charts.xcframework.tar.gz_ for distribution. Verified that the archive expands correctly and that the generated XCFramework works as expected when embedded in an iOS application.


liuxuan30:
Hi @gk-brown, thanks for looking into this new area! It's great to embrace a new format.

we didn't provide a pre-built binary framework, and we do support pod and carthage. Carthage seems fine because it should build from the source, while pod could be using the binary framework when we run 'pod trunk push'.

I do see pod seems support new XCFramework now https://github.com/CocoaPods/CocoaPods/pull/9334, but carthage seems not.

So in this term, supporting XCFramework can solve the original problem in #4317, that we don't have to build every time only because swift compiler upgrades.

However in this PR, I see you changed the access level and add `@objc` tag for some APIs, which seems  incorrect. Why is what? It would impact lots of users who override those functions.

Secondly, we should let pod know we support XCFramework and build it for distribution. I see you added a shell script but not in build phase, is it a must? Ideally we don't want to have scripts if it's necessary. If needed, it should be recognized by pod and do proper jobs. I don't think we should build a XCFramework by ourselves, since it's open sourced, and we already support swift package.

gk-brown:
> in this PR, I see you changed the access level and add @objc tag for some APIs, which seems incorrect.

The access level change was needed to resolve a compilation error when embedding the generated framework in an application. The methods were overriding `internal` APIs and making them `open`, which apparently is not supported in XCFrameworks. It also seemed like it might have been a mistake, since it exposes implementation details. However, making the base methods `open` instead of `internal` also resolves the problem. It might be worth fixing this regardless of whether you support XCFramework or not.

The `@objc` tag was added to resolve some compilation errors in the framework itself, after setting "Build Libraries for Distribution" to "Yes". I would not expect this change to have an impact on downstream code.

> I see you added a shell script but not in build phase, is it a must? 

The script is only required to package the XCFramework for distribution. I would prefer that Charts provide a pre-compiled binary like this since I don't use Carthage or CocoaPods. I did not realize that it was already available as a Swift package. That may be sufficient for my needs.

gk-brown:
Confirmed that Swift package works in my project. I will withdraw the PR.

Baaroon:
Hi @gk-brown, I'm trying to use Charts in as a dependency in a framework and when I try to create and xcframework, I faced the @objc and dynamic issues that've you mentioned in this PR. I tried to use Charts by Swift Package Manager but nothing has changed. May I know who you managed to fix your problem? Thanks

gk-brown:
@Baaroon I was able to get Charts working in my app using Swift Package Manager. Maybe the issue is that you are trying to use it from a framework? Not sure if that is supported or not as I have never tried it.

Did you try applying my patch and running the package.sh script? I wonder if this might work better since I assume a framework can have a dependency on another framework.

Baaroon:
Hey @gk-brown and thanks for your reply. What I'm trying to do is that I'm preparing a framework which I'll deliver it as a `.xcframework` to the user. I'm using Charts through cocoapods. So everything is fine except when I try to create the `xcframework`. I tried to use Swift Package Manager but I faced the same issue as `Overriding non-@objc declarations from extensions is not supported`. Your fix works fine as if I add a post_install script to my podfile and replace the `open func` with `@objc open func` in `Platform+Touch Handling.swift`, then I can proceed with creating the framework. 
I'm wondering is there any better way to do this or not.

gk-brown:
Hi @Baaroon - that makes sense. Since both Cocoapods and Swift Package Manager download and build the source locally, I'm not surprised that the result is the same. It seems like the problem is that the visibility modifiers for these methods as currently written simply do not work in an XCFramework. 

It might be worth resubmitting just the modifier changes from this PR as a separate ticket (minus the _package.sh_ addition), since I imagine this problem would affect anyone wanting to build/distribute Charts as part of an XCFramework.

PR #4345, opened 2020-04-20
Expose properties of scaling, tracking, and scrolling 

joeboyscout04:
### Goals :soccer:
This PR adds a few methods to expose some of the `UIScrollView`-like properties of `BarLineChartViewBase` so that we can better keep track of the chart's motion and react appropriately under conditions when the chart is moving.  

Implementation is trivial, as it just exposes internal properties.


liuxuan30:
I'm not sure if this is the right way to do it. what are you trying to do? Normally we don't want to expose private vars

PR #4366, opened 2020-05-11
Fix AxisBase Formatter Getter (Fixes #4365)

### LINKED ISSUE #4365: Overriding DefaultAxisValueFormatter Doesn't Work
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Overrode Default Axis Formatter 

```
class InvertedAxisFormatter: DefaultAxisValueFormatter {

    @objc override func stringForValue(_ value: Double, axis: AxisBase?) -> String {
        return super.stringForValue(value.isAlmostZero() ? 0 : -value, axis: axis)
    }
}
```

My method gets called. 

## What happened instead?

The getter for the base formatter checks if the decimals match and replaces my formatter.  

## Charts Environment

**Charts version/Branch/Commit Number:** 7ac4a05e83d279bac305a472af9128c2a0eca3d7
**Xcode version:** 11.3.1
**Swift version:** 5.0
**macOS version running Xcode:** 10.15.4

giantramen:
Fixes issue where if you overrode the
formatter it didn't matter because a new one would
just replace it.

### Issue Link https://github.com/danielgindi/Charts/issues/4365

### DIFF
--- a/Source/Charts/Components/AxisBase.swift
+++ b/Source/Charts/Components/AxisBase.swift
@@ -166,14 +166,17 @@ open class AxisBase: ComponentBase
     {
         get
         {
-            if _axisValueFormatter == nil ||
-                (_axisValueFormatter is DefaultAxisValueFormatter &&
-                    (_axisValueFormatter as! DefaultAxisValueFormatter).hasAutoDecimals &&
-                    (_axisValueFormatter as! DefaultAxisValueFormatter).decimals != decimals)
+            if _axisValueFormatter == nil
             {
                 _axisValueFormatter = DefaultAxisValueFormatter(decimals: decimals)
             }
-            
+            else if _axisValueFormatter is DefaultAxisValueFormatter &&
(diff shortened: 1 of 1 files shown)


giantramen:
@liuxuan30 Would you mind reviewing (or adding someone to review) this?

liuxuan30:
can you give some context why this change? e.g. If you set the formatter first, the old should work.

giantramen:
I am a little confused by what you're asking for. 
"Fixes issue where if you overrode the
formatter it didn't matter because a new one would
just replace it."

liuxuan30:
Ah at first I mean I didn't clearly see why old code not work.  but in your case, are you saying that when you have a formatter and meet the condition, it will create a new one, so some of your properties will be lost? is this correct?

PR #4375, opened 2020-05-22
Rounded bars in bar charts

Pearapps:
### Issue Link :link:

Other PR: https://github.com/danielgindi/Charts/pull/3754/

I understand that there are other attempts at this, but I am working on a project that requires this - so I forked this and gave it a shot with the latest master. 

### Goals :soccer:

Provide ability to round bar charts for charts that contain only positive values.

Pretty straightforward - provides a property to set the multiplier you wish to drive the corner radius - and the renderer searches for the correct bar segment to apply that to, and draws it with a path.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift
@@ -148,7 +148,8 @@ open class BarChartDataSet: BarLineScatterCandleBubbleChartDataSet, IBarChartDat
 
     /// the alpha value (transparency) that is used for drawing the highlight indicator bar. min = 0.0 (fully transparent), max = 1.0 (fully opaque)
     open var highlightAlpha = CGFloat(120.0 / 255.0)
-        
+    
+    /// The multiplier to generate a corner radius for the bar. 0.0 means no round radius.
     open var roundRadiusWidthMultiplier: CGFloat = 0.0
     
     // MARK: - NSCopying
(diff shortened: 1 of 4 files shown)


liuxuan30:
any reason you send this PR to us?

Pearapps:
> any reason/highlight you send this PR to us?

Didn't seem like there was any movement on the existing efforts, and people really want and this (including myself) - and I wanted to provide this work for you all to use or iterate on.

amiantos:
Do you have an example of how this is supposed to work? Any value I pass to `roundRadiusWidthMultiplier` seems to result in a crash.

Pearapps:
> Do you have an example of how this is supposed to work? Any value I pass to roundRadiusWidthMultiplier seems to result in a crash. I am using a stacked bar chart.

What values are you passing in?

Here is what I am setting it to: `1.0 / 4.0`

amiantos:
I tried `1.0`, `0.2`, `0.5`..

I don't really know what the crash is. Seems to happen no matter what value I put in there.

image

image

liuxuan30:
for rounded bar, I think the best PR so far is https://github.com/danielgindi/Charts/pull/3754. However the author not responding, and we really are also busy with our jobs... so, if you think you want to contribute, I guess you will take some from #3754?

The main problem for rounded bar is, when the bar is thinner, the rounded bar will look like an oval, which is not ideal. #3754 said she/he solved it. if you want to take it from here, if you can solve this, we'd happy to review

4np:
This would be easy to accomplish by providing your custom renderer, if #4297 was merged.

pmairoldi:

PR #4382, opened 2020-06-01
[FIX] Accessibility frame calculation

mstana:
### Goals :soccer:
Fixing `accessibilityFrame` for VoiceOver to show the size and position of a frame on screen in any given time and container scroll view.

Current implementation ignores redraws of the screen and calculates frame for accessibility only when setting property - which is incorrect because the screen can resize/change during reloading. The proposed fix will calculate the correct frame when accessibility "asks" for the frame, not when the external component is setting frame. 

The issue is primarily seen when the graphs are inside collection or scroll views.


zmian:
@liuxuan30 Would you mind reviewing this please?

liuxuan30:
this CI failed, plese take a look?

mstana:
@liuxuan30 thanks for letting me know. Now the CI passed. Seems like the first fail was my mistake, last was CI error.

liuxuan30:
so
>I would consider this fix also for the MacOS part but I am not in the right position to evaluate if it would help there also.

I think we only have one code base for macOS and iOS, you mean the fix could potentially not work on macOS?

Source/Charts/Utils/Platform+Accessibility.swift
```diff
@@ -58,13 +58,13 @@ open class NSUIAccessibilityElement: UIAccessibilityElement
     {
         get
         {
-            return super.accessibilityFrame
+            let frame = super.accessibilityFrame
+            return containerView?.convert(frame, to: UIScreen.main.coordinateSpace) ?? frame
```
liuxuan30, line 62: Maybe you can consider using our 'NSUIView' in platform.swift or similar what we did in platform+Color.swift for macOS and iOS, for `containerView?.convert()` to work on both iOS and macOS?

mstana: The macOS part is using different calculations of the `accessibilityFrame` as it's already addressing this problem which I am trying to fix inside the iOS part. I believe it's because the bug on the macOS was more visible (as it occurred right after moving/resizing the application window). 

Here is what's done for macOS: 

```
    open var accessibilityFrame: NSRect
    {
        get
        {
            return accessibilityFrame()
        }

        set
        {
            guard let containerView = containerView else { return }

            let bounds = NSAccessibility.screenRect(fromView: containerView, rect: newValue)

            // This works, but won't auto update if the window is resized or moved.
            // setAccessibilityFrame(bounds)

            // using FrameInParentSpace allows for automatic updating of frame when windows are moved and resized.
            // However, there seems to be a bug right now where using it causes an offset in the frame.
            // This is a slightly hacky workaround that calculates the offset and removes it from frame calculation.
            setAccessibilityFrameInParentSpace(bounds)
            let axFrame = accessibilityFrame()
            let widthOffset = abs(axFrame.origin.x - bounds.origin.x)
            let heightOffset = abs(axFrame.origin.y - bounds.origin.y)
            let rect = NSRect(x: bounds.origin.x - widthOffset,
                              y: bounds.origin.y - heightOffset,
                              width: bounds.width,
                              height: bounds.height)
            setAccessibilityFrameInParentSpace(rect)
        }
    }
```

Because of this, I am thinking that with this PR both - macOS and iOS will work just fine as macOS was already fixed and this PR will fix iOS.

Thank you and I am sorry for the late answer.

mstana: @liuxuan30 would you mind reviewing this, please?

PR #4422, opened 2020-07-21
Add @objc to fix dynamic inference

joelbanzatto:
### Goals :soccer:
Building on Xcode 11.6 using Swift 5 the compilation failed with the message "marking non-'@objc' swift declaration 'dynamic' in library evolution mode is not supported"

I added @objc and the Build goes Successfully.

### Testing Details :mag:
No tests required.

### DIFF
--- a/Source/Charts/Utils/Platform+Touch Handling.swift
+++ b/Source/Charts/Utils/Platform+Touch Handling.swift
@@ -37,22 +37,22 @@ extension NSUIView {
         self.nsuiTouchesCancelled(touches, withEvent: event)
     }
 
-    open func nsuiTouchesBegan(_ touches: Set<NSUITouch>, withEvent event: NSUIEvent?)
+    @objc open func nsuiTouchesBegan(_ touches: Set<NSUITouch>, withEvent event: NSUIEvent?)
     {
         super.touchesBegan(touches, with: event!)
     }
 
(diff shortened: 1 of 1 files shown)


danilobias:
Hi, I'm using Charts with Xcode 11.7 and Swift 5.0 and I have the same problem. It would be helpful if this correction could be approved.

@liuxuan30

liuxuan30:
I'm on this. May I ask the context about this? I don't remember people are complaning about this lots, and you seem the first. Under what context you meet this error?

danilobias:
Hi, 

This happens when I configure Build Libraries For Distribution with "YES" in my project's target and try to make Archive. When I build it doesn't happen.

liuxuan30:
This `Build Libraries For Distribution with "YES"` seems troublesome that after I turn it on, I can't even build ChartsDemo, with weird issues like:
image

any idea? @jjatie 

@danilobias why you need to turn this on? could you give more context? like you are embedding Charts into your own framework?

jjatie:
@liuxuan30 it appears that `BalloonMark` is not implicitly exposed to objective-c, but I'm not sure why. Explicitly marking it as `@objc` requires iOS 13 or later (also coincides with many Swift compatibility libraries being removed from the OS since it was the first major release since Swift was bundled with the OS). I'm not sure it makes sense for us to fully support the use case of `Build Libraries for Distribution` either, would love to hear @joelbanzatto use case.

Issue #709, opened 2016-01-26
HorizontalBarChartView negative the value label out of view


jasonxiaochao:
i use HorizontalBarChartView  and when the value is negative  ,than the drawvalue will be out of the chartview , i tried         
chart.leftAxis.axisMaximum
chart.leftAxis.axisMinimum
and still not working
what should i do?
![tmp7b6be8df](https://cloud.githubusercontent.com/assets/10300497/12571868/76d2b148-c420-11e5-9b5a-18fa90eb09b2.png)

liuxuan30:
reproduce it with ChartsDemo and post the data set code

jasonxiaochao:
barView.delegate = self
        barView.descriptionText = ""
        barView.noDataTextDescription = "Loading..."
        barView.drawBarShadowEnabled = false
        barView.pinchZoomEnabled = false
        barView.drawGridBackgroundEnabled = false
        barView.xAxis.drawAxisLineEnabled = false
        barView.xAxis.drawGridLinesEnabled = false
        barView.rightAxis.drawGridLinesEnabled = false
        barView.rightAxis.drawAxisLineEnabled = false
        barView.rightAxis.drawLabelsEnabled = false
        barView.leftAxis.drawGridLinesEnabled = false
        barView.leftAxis.drawLabelsEnabled = false
        barView.leftAxis.drawAxisLineEnabled = false
        barView.drawValueAboveBarEnabled = true
        barView.legend.enabled = false
        barView.xAxis.drawLabelsEnabled = false
        
        let lineZero = ChartLimitLine(limit: 0)
        lineZero.lineColor = UIColor.blackColor()
        lineZero.lineWidth = 0.5
        //lineZero.lineDashLengths = [5,5]
        lineZero.labelPosition = ChartLimitLine.ChartLimitLabelPosition.RightTop
        
        barView.rightAxis.addLimitLine(lineZero)

  var xVals = ["个股","行业","大盘"]
        var yVals = [BarChartDataEntry]()
        var a = BarChartDataEntry(value: -1.31, xIndex: 0)
        var b = BarChartDataEntry(value:2.15, xIndex: 1)
        var c = BarChartDataEntry(value: 3.08, xIndex: 2)
        yVals.append(a)
        yVals.append(b)
        yVals.append(c)
        
        let formatter = NSNumberFormatter()
        //formatter.maximumFractionDigits = 2;
        formatter.positiveFormat = "0.00"
        formatter.negativeSuffix = "%";
[...]

Issue #777, opened 2016-02-26
Stacked BarChartView: Position of Value-Labels


AndreasIgelCC:
Is there any option to set the Position of Value-Labels in Stacked BarChartViews? Is this feature in focus or is it not planed to be implemented?
I have something like that in Mind (Possibility to rotate Value-Labels and to position them inside the Stack):
<img width="710" alt="screenshot_279" src="https://cloud.githubusercontent.com/assets/2921976/13349839/97516e54-dc7c-11e5-8bd9-29ae015086fb.png">

liuxuan30:
it can show each sub bar's label already, but you cannot change the position with a simple enum.
What you can do is looking at bar chart renderer's drawValues() to change the position. A PR is more than welcome if you think it's a good feature to have and implemented in a flexible way.

MiladFaridnia:
After a year it's not implemented as a feature yet ?

liuxuan30:
yep, lack of time + people working on new features; even reviews, plus android sync..

PR #4431, opened 2020-07-30
Bar chart custom value offset

bivant:
### Issue Link :link:
Vertical version the HorizontalBarChartView issue  https://github.com/danielgindi/Charts/issues/124
Possibly (in future) https://github.com/danielgindi/Charts/issues/777

### Goals :soccer:
Scope: **BarChart** (no horizontal bar logic for now)
Add ability to specify **custom value offset** for a value label (prior value is 4.5, what is still true by default)
When was implementing this feature I faced with the issue with low bars - **text was unreadable** when it has the same color as the background. So secondary value colors were added. After that I realized that inverting the value text color would be more universal and require almost no additional logic (except stacked, but that was later), so added this feature also.
Although coloring logic can be a separate PR, I have no interest in it without custom/flexible value positioning, so please not send me to split it right away.
While I need a non-stacked version I saw that drawing logic is very similar to "single bar" so added support of the features to it as well.

**BarChartDataProvider** (affecting BarChartView):
```swift
    /// if set to true values those
    /// 1.do not fit into the value bar are drawn above them (below for negative), instead of partially inside and/or below (above if negative)
    /// 2.do not fit into the visible area above/below their bars are drawn inside
    var isDrawValueSideFlexible: Bool { get }
    /// distance from top (bottom in negative) for values drawn outside/inside the bar
    var valuesOffset: CGFloat { get }
```
BarChartView `valuesOffset` for a value label has same default value as before (4.5).
Setting a `isDrawValueSideFlexible` value resets the offset of vertical bars to zero as the padding is not needed because of the flexible position

**ChartBaseDataSet**:
```swift
    /// valueColors in case of soft positions in effect (BarCharView:isDrawValueSideFlexible)
    open var valueColorsSecondary = [NSUIColor]()

    /// Check if colors for the values for the DataSet are distinct from the background/bar color and apply inverted color if opposite
    open var valueColorsAdjustment: Bool = false
```
`valueColorsSecondary` are applied for the values affected by `isDrawValueSideFlexible` only.
`valueColorsAdjustment` in effect for all value labels in the chart if enabled

### Testing Details :mag:
Tried to add tests for all the edge cases I could think of: custom value offset, flexible positioning, colors inverting+secondary, combination of them. Stacked and non-stacked versions.


bivant:
I am ready to port the changes to the HorizontalBar if they are accepted.
These issues may be fixed:
https://github.com/danielgindi/Charts/issues/709
https://github.com/danielgindi/Charts/issues/124

For the https://github.com/danielgindi/Charts/issues/777 need guidance where to put the parameters as there should be a lot of them - position factor, offset, rotation, alignment, maybe something else.

bivant:
Seems like the reference images for the new tests need to be updated, that was done for existing tests in master already after the merge with 4.0 branch.

PR #4452, opened 2020-08-21
Scanner Functionality Deprecated (Catalyst)

AdamFootDev:
### Goals :soccer:
Replace deprecated Scanner functionality in ChartColorTemplates.swift that'll work with iOS 13+ and Mac Catalyst 13+.

### DIFF
--- a/Source/Charts/Utils/ChartColorTemplates.swift
+++ b/Source/Charts/Utils/ChartColorTemplates.swift
@@ -129,19 +129,39 @@ open class ChartColorTemplates: NSObject
         else if colorString.hasPrefix("rgba")
         {
             var a: Float = 1.0
-            var r: Int32 = 0
-            var g: Int32 = 0
-            var b: Int32 = 0
-            let scanner: Scanner = Scanner(string: colorString)
-            scanner.scanString("rgba", into: nil)
-            scanner.scanCharacters(from: leftParenCharset, into: nil)
-            scanner.scanInt32(&r)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanInt32(&g)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanInt32(&b)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanFloat(&a)
+            var r: Int64 = 0
+            var g: Int64 = 0
+            var b: Int64 = 0
+            
+            // Some scanner functionality has been deprecated and replaced in iOS 13
+            
+            if #available(iOS 13.0, *)
+            {
+                let scanner = Scanner(string: colorString)
+                _ = scanner.scanString("rgba")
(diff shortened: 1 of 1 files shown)


Source/Charts/Utils/ChartColorTemplates.swift
```diff
@@ -129,19 +129,39 @@ open class ChartColorTemplates: NSObject
...
-            scanner.scanString("rgba", into: nil)
-            scanner.scanCharacters(from: leftParenCharset, into: nil)
-            scanner.scanInt32(&r)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanInt32(&g)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanInt32(&b)
-            scanner.scanCharacters(from: commaCharset, into: nil)
-            scanner.scanFloat(&a)
+            var r: Int64 = 0
```
jjatie, line 132: Eventually (when only supporting iOS 13+ we will be removing the alternative implementation. Rather than try to match the old implementation, it would be better to use a new implementation that will allow us to just delete the old code altogether without having to think about what we are deleting. It's also a good opportunity to improve the implementation. One way to do this, which I think feels a little better:
```
let scanner: Scanner = Scanner(string: colorString)
scanner.charactersToBeSkipped = commaCharset.union(leftParenCharset)
if #available(iOS 13.0, *)
{
    _ = scanner.scanString("rgba")
    let r = scanner.scanInt() ?? 0
    let g = scanner.scanInt() ?? 0
    let b = scanner.scanInt() ?? 0
    let a = scanner.scanFloat(representation: .decimal) ?? 1.0

    return NSUIColor(
        red: CGFloat(r) / 255.0,
        green: CGFloat(g) / 255.0,
        blue: CGFloat(b) / 255.0,
        alpha: CGFloat(a)
    )
}
else
{
    var a: Float = 1.0
    var r: Int32 = 0
    var g: Int32 = 0
    var b: Int32 = 0

    scanner.scanString("rgba", into: nil)
    scanner.scanInt32(&r)
    scanner.scanInt32(&g)
    scanner.scanInt32(&b)
    scanner.scanFloat(&a)

    return NSUIColor(
        red: CGFloat(r) / 255.0,
        green: CGFloat(g) / 255.0,
        blue: CGFloat(b) / 255.0,
        alpha: CGFloat(a)
    )
}
```

jjatie (changes requested):
I think it would be best to move the scanner declaration to the top of the method. Rather than check the starting characters of the string, you can use `if scanner.scanString(XXX) != nil {`

Also, let's migrate the `#` prefixed implementation to use a scanner as well.

Source/Charts/Utils/ChartColorTemplates.swift
jjatie: Forgot to show:
```
let scanner = Scanner(string: colorString)
scanner.charactersToBeSkipped = commaCharset.union(leftParenCharset)
```

PR #4454, opened 2020-08-24
fix perfomance issues for very large dataSets

mhecker:
replaces linear search implicit in call dataSet.entryForXValue(..)) by existing binary search via dataSet.entryIndex(...).

### Goals :soccer:
This pull requests aims to improve performance on very large DataSets.
We observed perfomance issues for LineCharts with very large number of Entries.
It is caused by a use of a linear search using the default firstIndexOf() method called in ChartDataSet.firstIndex().

No architectural changes.

### Testing Details :mag:
No added Tests.

### DIFF
--- a/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
+++ b/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
@@ -78,11 +78,11 @@ open class BarLineScatterCandleBubbleRenderer: DataRenderer
             let low = chart.lowestVisibleX
             let high = chart.highestVisibleX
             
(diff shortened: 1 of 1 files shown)


jjatie:
@mhecker Were you running into performance issues? How large was the data set? Has MPAndroidCharts done the same (we try to match implementations where possible).

This has been a long standing issue. IIRC @liuxuan30 @danielgindi the binary search was removed because we don't guarantee that datasets are ordered. I have thoughts on how we can move towards this.

enamor:
Has the performance issue of large data sets been merged into the main branch?

liuxuan30:
would need more details I suppose. Just changing a few lines can fix it seems too good to be true for me..

PR #4456, opened 2020-08-27
Resolve conflict for 4.0 branch and master

liuxuan30:
This is the pull request for merging 4.0 branch into master.

It's my bad that I forgot to directly merge master into 4.0 branch. Instead, I branch master to this `ResolveConflict` and merge 4.0 into this branch.

I have tried my best to resolve all conflicts and inconsistent code, but the code that does not trigger a conflict still has errors.

The plan is to review this carefully and get it merged before iOS 14 rolls out.

@jjatie @danielgindi if you have some time now, please help review it and maybe fix some compile errors. I'll keep fixing the compile errors until it could build and run. but meantime please help review as well. I will graducally add comment for things needed disccusion.


liuxuan30:
I have fixed the compile errors in https://github.com/danielgindi/Charts/pull/4456/commits/3f84f8777e18ba241292a5c2bc023ab47c6cc3b2

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -55,12 +56,18 @@ open class ChartDataSet: ChartBaseDataSet
...
-    @available(*, unavailable, renamed: "entries")
-    @objc
-    open var values: [ChartDataEntry] { return entries }
-
     @objc
     open private(set) var entries: [ChartDataEntry]
+    {
+        didSet
+        {
+            if isIndirectValuesCall {
```
liuxuan30, line 64: @jjatie one of the main blocker now is this `isIndirectValuesCall`. I have removed the old `values` var to make it simple 

Could you take a look if it can be removed now?

liuxuan30:
@jjatie @danielgindi could any of you can take a look at why Bar Chart all failed? I see a shift on y axis, but couldn't remember what's causing the shift:
image
you can see the bar chart in whole on left side is a bit lower on y-axis direction than right one.

@jjatie besides the `-0` come  back again.

Source/Charts/Charts/ChartViewBase.swift
```diff
@@ -51,43 +47,53 @@ open class ChartViewBase: NSUIView, ChartDataProvider, AnimatorDelegate
...
+            offsetsCalculated = false
+
+            guard let data = data else { return }
+
+            // calculate how many digits are needed
+            setupDefaultFormatter(min: data.yMin, max: data.yMax)
+
+            for set in data.dataSets
+            {
+                if set.valueFormatter === defaultValueFormatter
```
liuxuan30, line 67: I'm kind of confused for this, as the old logic is 
```swift
// calculate how many digits are needed	
            setupDefaultFormatter(min: _data.getYMin(), max: _data.getYMax())	

            for set in _data.dataSets	
            {	
                if set.needsFormatter || set.valueFormatter === _defaultValueFormatter	
                {	
                    set.valueFormatter = _defaultValueFormatter	
                }	
            }	

            // let the chart know there is new data	
            notifyDataSetChanged()
```
which will hit set.needsFormatter so set.valueFormatter is set anyway. But the new code is confusing, as if it really `===`, there is no reason to set again?

besides, this is calling bar charts and pie charts unittests failing

I would like to remove it and try to modify the UT to avoid updating test images.

@jjatie @danielgindi what you think?

liuxuan30:
ok, only line chart and combined chart still fails. I will continue next week

liuxuan30:
all UT fixed. Let's continue the original review process.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -382,93 +268,49 @@ open class ChartData: NSObject
...
-            let set = _dataSets[dataSetIndex]
-            
-            if !set.addEntry(e) { return }
-            
-            calcMinMax(entry: e, axis: set.axisDependency)
-        }
-        else
-        {
-            print("ChartData.addEntry() - Cannot add Entry because dataSetIndex too high or too low.", terminator: "\n")
+        guard dataSets.count > dataSetIndex && dataSetIndex >= 0 else {
```
liuxuan30, line 298: @jjatie I revert back the similar commits whose pattern is `dataSets.contains(x)` as it is O(n) complexity, which is unnecessary, though it's simpler,  but I would still prefer direct compare

jjatie: `dataSets.indices.contains(dataSetIndex)` is O(1). This is true for all sodlib `Range` types.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -382,93 +268,49 @@ open class ChartData: NSObject
...
-        calcMinMax()
-        
-        return true
-    }
-    
+
     /// Adds an Entry to the DataSet at the specified index. Entries are added to the end of the list.
-    @objc open func addEntry(_ e: ChartDataEntry, dataSetIndex: Int)
+    @objc(addEntry:dataSetIndex:)
+    open func appendEntry(_ e: ChartDataEntry, toDataSet dataSetIndex: Index)
```
jjatie, line 296: This method should probably be deprecated altogether now that there is `Collection` conformance for the `*Data` and `*DataSet` types. Previously that was a nice convenience, but now it just becomes:
```
chartData[datSetIndex].append(entry)
// vs
chartData.appendEntry(entry, toDataSet: dataSetIndex)
```
The former is the way all other Swift collections work

jjatie: Actually, I just remembered this method guarantees the index is valid. This is different than how all Swift collections work. It is actually in the [common rejections](https://github.com/apple/swift-evolution/blob/master/commonly_proposed.md#strings-characters-and-collection-types) for Swift Evolution.

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -382,93 +268,49 @@ open class ChartData: NSObject
...
-        }
-        else
-        {
-            print("ChartData.addEntry() - Cannot add Entry because dataSetIndex too high or too low.", terminator: "\n")
+        guard dataSets.count > dataSetIndex && dataSetIndex >= 0 else {
+            return print("ChartData.addEntry() - Cannot add Entry because dataSetIndex too high or too low.", terminator: "\n")
         }
+
+        let set = self[dataSetIndex]
+        if !set.addEntry(e) { return }
```
jjatie, line 303: This always returns `true` and is deprecated. Should be replace by `self[dataSetIndex].append(e)`

jjatie: I just checked and this is the only use of `addEntry` remaining in the project

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -382,93 +268,49 @@ open class ChartData: NSObject
...
+        calcMinMax(entry: e, axis: set.axisDependency)
```
jjatie, line 304: Since large changes are being made for 4.0.0, I think we should reconsider some of these, potentially complex, side effects. We should leave them there in the deprecated old API, but I think we should recommend that they call it themselves with the new `Collection` API. I've found in my own experience, when working with live data, most data is around the same value and I had to remove the various `calc` methods that fire when modifying a dataSet to get the desired performance

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -382,93 +268,49 @@ open class ChartData: NSObject
...
     }
-    
+
     /// Removes the given Entry object from the DataSet at the specified index.
-    @objc @discardableResult open func removeEntry(_ entry: ChartDataEntry, dataSetIndex: Int) -> Bool
+    @objc @discardableResult open func removeEntry(_ entry: ChartDataEntry, dataSetIndex: Index) -> Bool
```
jjatie, line 308: We should probably deprecate these `removeEntry` methods to encourage standard Swift Collection API. i.e.
```
if let index = chartData[dataSetIndex].index(of: entry) {
    chartData[dataSetIndex].remove(at: index)
}
```

Source/Charts/Data/Implementations/Standard/ChartData.swift
```diff
@@ -482,129 +324,251 @@ open class ChartData: NSObject
     /// specified index. 
     ///
     /// - Returns: `true` if an entry was removed, `false` ifno Entry was found that meets the specified requirements.
-    @objc @discardableResult open func removeEntry(xValue: Double, dataSetIndex: Int) -> Bool
+    @objc @discardableResult open func removeEntry(xValue: Double, dataSetIndex: Index) -> Bool
```
jjatie, line 327: Then this one could be `index(ofXValue:)` (on `ChartDataSet`)

Source/Charts/Data/Implementations/Standard/ChartData.swift
liuxuan30: ok

Source/Charts/Data/Implementations/Standard/ChartData.swift
liuxuan30: for simple improvements, please pull this branch and just update the code, usually I should be ok with it. I will check what you updated

Source/Charts/Data/Implementations/Standard/ChartData.swift
liuxuan30: how about mark some 2.0/3.0 APIs to deprecate, and remove in 4.1? or remove it now?

@danielgindi idea?

Source/Charts/Data/Implementations/Standard/ChartData.swift
liuxuan30: plese go ahread update it

Source/Charts/Data/Implementations/Standard/ChartData.swift
jjatie: I suppose this and the above comment were for @danielgindi . There have been disagreements in the past around API changes related to this. We didn't have the `Collection` API then though.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: @jjatie this is kind of higher priority, not sure if can remove it safely now?

jjatie: I made the necessary changes to remove it.

danielgindi:
> @jjatie @danielgindi could any of you can take a look at why Bar Chart all failed? I see a shift on y axis, but couldn't remember what's causing the shift:
> image
> you can see the bar chart in whole on left side is a bit lower on y-axis direction than the right one.
> 
> It seems it's because `func calculateLegendOffsets`
> 

There was a fix a while back to the legend offsets. It shifted all tests too... I'm not sure how the merge did not catch this as there are snapshots involved.

danielgindi:
I think that if this branch is tested and works fine, and the whole point about it is a bigger shift towards Swift, away from ObjC compatibility - then just go ahead, merge it.
Let people test it as a `beta` or an `rc` for a few days.

I don't see anything that breaks compatibility with the Java counterpart also. The naming is converted to swift-like namings, but the workings stays the same.

liuxuan30:
> > @jjatie @danielgindi could any of you can take a look at why Bar Chart all failed? I see a shift on y axis, but couldn't remember what's causing the shift:
> > image
> > you can see the bar chart in whole on left side is a bit lower on y-axis direction than the right one.
> > It seems it's because `func calculateLegendOffsets`
> 
> There was a fix a while back to the legend offsets. It shifted all tests too... I'm not sure how the merge did not catch this as there are snapshots involved.

yes this has been fixed.

liuxuan30:
> I think that if this branch is tested and works fine, and the whole point about it is a bigger shift towards Swift, away from ObjC compatibility - then just go ahead, merge it.
> Let people test it as a `beta` or an `rc` for a few days.
> 
> I don't see anything that breaks compatibility with the Java counterpart also. The naming is converted to swift-like namings, but the workings stays the same.

so about deprecating APIs, green light from you? @danielgindi

danielgindi:
> > I think that if this branch is tested and works fine, and the whole point about it is a bigger shift towards Swift, away from ObjC compatibility - then just go ahead, merge it.
> > Let people test it as a `beta` or an `rc` for a few days.
> > I don't see anything that breaks compatibility with the Java counterpart also. The naming is converted to swift-like namings, but the workings stays the same.
> 
> so about deprecating APIs, green light from you? @danielgindi

I see breaking changes, not deprecations, maybe I missed something. Anyway - that's what major versions are for right? :-)

liuxuan30:
> > > I think that if this branch is tested and works fine, and the whole point about it is a bigger shift towards Swift, away from ObjC compatibility - then just go ahead, merge it.
> > > Let people test it as a `beta` or an `rc` for a few days.
> > > I don't see anything that breaks compatibility with the Java counterpart also. The naming is converted to swift-like namings, but the workings stays the same.
> > 
> > 
> > so about deprecating APIs, green light from you? @danielgindi
> 
> I see breaking changes, not deprecations, maybe I missed something. Anyway - that's what major versions are for right? :-)

```swift
    /// The number of y-values this DataSet represents
    @available(*, deprecated, message: "Use `count` instead")
    open override var entryCount: Int { return count }
    
    /// - Throws: out of bounds
    /// if `i` is out of bounds, it may throw an out-of-bounds exception
    /// - Returns: The entry object found at the given index (not x-value!)
    @available(*, deprecated, message: "Use `subscript(index:)` instead.")
    open override func entryForIndex(_ i: Int) -> ChartDataEntry?
    {
        guard i >= startIndex, i < endIndex else {
            return nil
        }
        return self[i]
    }
```
we have 7 APIs marked as `@available(*, deprecated")` in master like above. So it looks like we can remove them now. @jjatie are there any other APIs we should pay attention? like APIs should mark as deprecated or to be removed?

jjatie:
@liuxuan30 They have been marked deprecated, but those warnings have never been released. We should keep them for now, and remove in a subsequent release.

liuxuan30:
@jjatie @danielgindi once it passes CI, seems we can move ahead to prelease 4.0 now

liuxuan30:
@jjatie I created the prelease of 4.0, please add any documentation https://github.com/danielgindi/Charts/releases/tag/v4.0.0 if you think it's important. Thank you for all your efforts!! I know it has been so long to get this done.

PR #4459, opened 2020-08-28
Remove internal use of datasets

jjatie:
### Goals :soccer:
New Collection API is encouraged for access to data sets in ChartData. DataSets is only exposed for custom chart use cases (like PieChartData). We should not be relying on it internally. There are a couple oddities that (have always) exist like down casting elements to a specific type. Those won't be resolved until objc support is either removed, or refactored out of the main chart data types.

### DIFF
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -71,7 +71,7 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
             }
         }
 
-        _buffers = zip(_buffers, barData.dataSets).map { buffer, set -> Buffer in
+        _buffers = zip(_buffers, barData).map { buffer, set -> Buffer in
             let set = set as! BarChartDataSetProtocol
             let size = set.entryCount * (set.isStacked ? set.stackSize : 1)
             return buffer.count == size
@@ -436,17 +436,15 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
                 let barData = dataProvider.barData
                 else { return }
 
-            let dataSets = barData.dataSets
-
             let valueOffsetPlus: CGFloat = 4.5
             var posOffset: CGFloat
             var negOffset: CGFloat
             let drawValueAboveBar = dataProvider.isDrawValueAboveBarEnabled
             
-            for dataSetIndex in barData.dataSets.indices
+            for dataSetIndex in barData.indices
             {
                 guard
-                    let dataSet = dataSets[dataSetIndex] as? BarChartDataSetProtocol,
+                    let dataSet = barData[dataSetIndex] as? BarChartDataSetProtocol,
                     shouldDrawValues(forDataSet: dataSet)
                     else { continue }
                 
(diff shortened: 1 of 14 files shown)


liuxuan30:
CI failed

jjatie:
Appears to be an issue in pre-Swift 5.3 compiler. Moving Travis to Xcode 12 should fix it.

liuxuan30:
let's start this @jjatie

jjatie:
@liuxuan30 This builds fine in Xcode 12. CI is failing in Xcode 12 though. It appears to be an issue with the FBSnapshotTest library again. #4476 should be solved first

liuxuan30:
@jjatie not sure why it has conflicts after merging #4476 , could you take a look?

Source/Charts/Renderers/BubbleChartRenderer.swift
```diff
@@ -153,18 +153,22 @@ open class BubbleChartRenderer: BarLineScatterCandleBubbleRenderer
...
         {
-            let dataSet = dataSets[i]
+            guard let dataSet = bubbleData[i] as? BubbleChartDataSetProtocol,
+                  shouldDrawValues(forDataSet: dataSet)
+            else
+            {
+                continue
+            }
 
             guard shouldDrawValues(forDataSet: dataSet) else { continue }
```
liuxuan30, line 173: @jjatie we have dup check of `shouldDrawValues`?

Source/Charts/Renderers/CandleStickChartRenderer.swift
```diff
@@ -38,7 +38,7 @@ open class CandleStickChartRenderer: LineScatterCandleRadarRenderer
             accessibleChartElements.append(element)
         }
 
-        for set in candleData.dataSets as! [CandleChartDataSetProtocol] where set.isVisible
+        for case let set as CandleChartDataSetProtocol in candleData where set.isVisible
```
liuxuan30, line 41: do you remember when `xxx case let` is introduced? not sure if people would complain it won't build on older swift

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -71,7 +71,7 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
             }
         }
 
-        _buffers = zip(_buffers, barData.dataSets).map { buffer, set -> Buffer in
+        _buffers = zip(_buffers, barData).map { buffer, set -> Buffer in
```
liuxuan30, line 74: @jjatie what you think is the proper time we remove `dataSets` var? Like v4.1?

liuxuan30:
I restarted the failed build and now all green.

Source/Charts/Renderers/CandleStickChartRenderer.swift
jjatie: Pattern matching has been around since at least Swift 2. I believe It was actually in the first public release.

Source/Charts/Renderers/BarChartRenderer.swift
jjatie: I'd say move it to `obsoleted` in 4.1, then transition it out in 5?

jjatie: I think 5 can go a little faster that 4. I think a good goal for 5 would be to move all objc support into a compatibility library, and remove it from the core project. This would allow us to use Swift's generics system, protocols, and structs more effectively in the core lib, and expose more type safe API in the objc compat library too!

liuxuan30 (approved):

PR #4461, opened 2020-08-31
Remove misuse of `count`

jjatie:
`count` is frequently used on collections to infer other semantic meaning. These uses have been replaced with the correct version.

### DIFF
--- a/Source/Charts/Utils/ChartColorTemplates.swift
+++ b/Source/Charts/Utils/ChartColorTemplates.swift
@@ -89,41 +89,37 @@ open class ChartColorTemplates: NSObject
         if colorString.hasPrefix("#")
         {
             var argb: [UInt] = [255, 0, 0, 0]
-            let colorString = colorString.unicodeScalars
-            var length = colorString.count
+            let colorString = colorString.unicodeScalars.dropFirst()
+            let length = colorString.count
             var index = colorString.startIndex
             let endIndex = colorString.endIndex
-            
-            index = colorString.index(after: index)
-            length = length - 1
-            
-            if length == 3 || length == 6 || length == 8
+
+            guard [3, 6, 8].contains(length) else { return .black }
+
+            var i = length == 8 ? 0 : 1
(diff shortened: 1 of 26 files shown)


liuxuan30:
@jjatie any dependency? Been busy this month (family+work). I will start to work on getting `ResolveConflict` merged from now on. only this one and https://github.com/danielgindi/Charts/pull/4459 into `ResolveConflict` branch?

jjatie:
No dependency. Note this targets your WIP ResolveConflict branch

Source/Charts/Utils/ChartColorTemplates.swift
```diff
@@ -89,41 +89,37 @@ open class ChartColorTemplates: NSObject
...
+            let length = colorString.count
             var index = colorString.startIndex
             let endIndex = colorString.endIndex
-            
-            index = colorString.index(after: index)
-            length = length - 1
-            
-            if length == 3 || length == 6 || length == 8
+
+            guard [3, 6, 8].contains(length) else { return .black }
```
liuxuan30, line 97: `.black` is not in the old code right?
if length is not in 3,6,8, it would return 
```
NSUIColor(red: CGFloat(argb[1]) / 255.0, green: CGFloat(argb[2]) / 255.0, blue: CGFloat(argb[3]) / 255.0, alpha: CGFloat(argb[0]) / 255.0)
```

liuxuan30: @jjatie I have updated uisng a `fallbackColor`. If nothing else, we could merge this after CI pass

jjatie: Upon closer inspection, the fallback colour IS `.black`. I'm going to revert your commit (which doesn't compile)

liuxuan30: ah sorry. I'm using the web editor for a quick change, not realize this

PR #4476, opened 2020-09-23
Bump travis ci to use xcode 12

jjatie:
This is just to bump travis to use the latest Swift compiler. Project changes for Xcode 12 will come in a separate PR


liuxuan30:
```
*** Checking out ios-snapshot-test-case at "6.2.0"
*** Cloning ios-snapshot-test-case
*** xcodebuild output can be found in /var/folders/z3/_825pg0s3jvf0hb_q8kzmg5h0000gn/T/carthage-xcodebuild.iiamXF.log
*** Downloading ios-snapshot-test-case.framework binary at "6.2.0"
***  Skipped installing ios-snapshot-test-case.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7) and the local version is 5.3 (swiftlang-1200.0.29.2 clang-1200.0.30.1)."

```
ios-snapshot-test-case is not updated so the binary framework failed. I checked there is no new update on their repo. So seems we have to build from scratch to avoid this at least for now

liuxuan30:
@jjatie seems we have a dependency issue that the source isn't Xcode12 compatible. Either we have to wait or we grab the source code and fix what is broken. idea?

liuxuan30:
@jjatie I'm having a hard time on my side that most of the tests failed even in the Xcode itself. 
I'm seeing the failures because image diff exceeds the tolerance. however I couldn't say who's fault. (likely Apple?)

for example, the diff image on my side is:
image

I also roll back to my commit `fix line chart UT failures`, it still fails. so I think this might because of the new iOS14 SDK? I'm on Xcode 12.0.1 right now.

could you try it on your side to see what will happen?

if you got similar issues like me, do we report or we just flush new images?

liuxuan30:
I have made the image refresh a single commit so we can move on or revert. I'm really confused, but seems there is some hidden changes in the UIKit or CoreGraphics rendering component that causes the slight difference.
@jjatie I still need you to help to take a look at old images and have consistent behaviours like me. thank you

liuxuan30:
let's move ahead

PR #4494, opened 2020-10-26
Add RoundedCornersBarChatRenderer which provides the ability to draw bar charts with rounded corners

ZakharovPavlo:
### Goals :soccer:
Give user ability to draw bar charts with rounded corners and specify corner radius as well as corners to be rounded

The RoundedCornersBarChatRenderer class was created which inherits from BarChartRenderer and overrides 
In BarChartRenderer drawDataSet(context: CGContext, dataSet: IBarChartDataSet, index: Int) method to include cornerRadius and UIRectCorner to which it applies rounding. So the basically use case is:
Use copy constructs RoundedCornersBarChatRenderer from regular BarChartRenderer class.
Then sets corners and radius.
Sets the renderer to the ChartView
Now user is able to draw bar charts with rounded corners.

### Testing Details :mag:
No tests where added, will add if requested


4np:
This would be easy to accomplish by providing your custom renderer, if #4297 was merged.

ZakharovPavlo:
Absolutely agree. Please let me know if it's merged so I will do some refinements to my pr.

pmairoldi:

PR #4497, opened 2020-11-06
Import swift algorithms

jjatie:
Provides convenient algorithms written by the community, curated be the swift core team.

### DIFF
--- a/Charts.xcodeproj/project.pbxproj
+++ b/Charts.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 46;
+	objectVersion = 52;
 	objects = {
 
 /* Begin PBXBuildFile section */
@@ -29,6 +29,7 @@
 		219192CA6B4895319AB49DCA /* BarLineScatterCandleBubbleRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B1C588E9DF6FFD56D7ADF8E /* BarLineScatterCandleBubbleRenderer.swift */; };
 		221CA2922588FCBC00C2DD1E /* Sequence+KeyPath.swift in Sources */ = {isa = PBXBuildFile; fileRef = 221CA2912588FCBC00C2DD1E /* Sequence+KeyPath.swift */; };
 		2243BBFD1FF156EC00B49D0B /* EquatableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2243BBFB1FF156D000B49D0B /* EquatableTests.swift */; };
+		228D56232554F27A00BEE75E /* Algorithms in Frameworks */ = {isa = PBXBuildFile; productRef = 228D56222554F27A00BEE75E /* Algorithms */; };
 		23649EFC635A76022F07FFA6 /* PieChartDataEntry.swift in Sources */ = {isa = PBXBuildFile; fileRef = BD02157CF8CEE1189BF681DA /* PieChartDataEntry.swift */; };
 		23FA50B2730D8C7ACA091C4F /* BarChartRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75F279974FE650E57A061B09 /* BarChartRenderer.swift */; };
 		24151B0729D77251A8494D70 /* LineRadarRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 105FFC9D3773A9C7A60A897F /* LineRadarRenderer.swift */; };
@@ -344,6 +345,7 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				228D56232554F27A00BEE75E /* Algorithms in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -748,6 +750,9 @@
 			dependencies = (
 			);
 			name = Charts;
+			packageProductDependencies = (
+				228D56222554F27A00BEE75E /* Algorithms */,
+			);
 			productName = Charts;
 			productReference = 65AD9E95D9ED4DC0BD73A743 /* Charts.framework */;
 			productType = "com.apple.product-type.framework";
@@ -798,6 +803,9 @@
 				Base,
 			);
 			mainGroup = 865A1CF149F52850CAB7F177;
+			packageReferences = (
+				228D56212554F27A00BEE75E /* XCRemoteSwiftPackageReference "swift-algorithms" */,
+			);
 			productRefGroup = AB2D554102718F209377399E /* Products */;
 			projectDirPath = "";
 			projectRoot = "";
@@ -1050,7 +1058,11 @@
 				INFOPLIST_FILE = "Source/Supporting Files/Info.plist";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@loader_path/Frameworks",
+				);
 				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				MARKETING_VERSION = 4.0.0;
 				MTL_ENABLE_DEBUG_INFO = YES;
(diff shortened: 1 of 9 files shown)


liuxuan30:
sorry guys. been very busy in recent months

Charts.xcodeproj/project.pbxproj
```diff
@@ -1050,7 +1058,11 @@
 				INFOPLIST_FILE = "Source/Supporting Files/Info.plist";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = (
```
liuxuan30, line 1061: @jjatie are these diffs generated by Xcode?

Charts.xcodeproj/project.pbxproj
```diff
@@ -1307,6 +1343,25 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
```
liuxuan30, line 1347: @jjatie could you give some context what are these changes about?

Charts.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
```
liuxuan30, line 1: is this Package.resolved necessary to be commited? so everyone is with the same version of the packages?

Charts.xcodeproj/project.pbxproj
jjatie: Yea, those are generated by Xcode. Looks like the file format used to use a space separated string, and now it uses something like a tuple

Charts.xcodeproj/project.pbxproj
jjatie: Again, autogenned by Xcode when importing a Swift package to an xcproj. Required so that Cocoapods and Carthage build with our SPM dependencies.

This reminds me I forgot to update the Package.swift manifest for SPM build 🤦‍♂️

Charts.xcworkspace/xcshareddata/swiftpm/Package.resolved
jjatie: It is, for the reason you stated. Just like we commit a Gemfile.lock or Cartfile.resolved.

Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
```
jjatie, line 1: @liuxuan30 This is the .resolved for working on the SwiftPM version of Charts.

The one in `/xcsharedata/swiftpm` is for working on the .framework version used by Carthage and Cocoapods.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -195,58 +196,10 @@ open class ChartDataSet: ChartBaseDataSet
...
-                    high = m - 1
-                }
-            }
-        }
-        
-        return entries
+        let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
+        let i = partitioningIndex(where: match)
+        guard i < endIndex else { return [] }
+        return self[i...].prefix(while: match)
```
liuxuan30, line 202: in terms of performance, I came across this https://forums.swift.org/t/performance-of-prefix-while/12993/3

@jjatie do you think we should use `lazy` here?

jjatie: looks like it. Good find

jjatie: pushed

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -260,98 +213,57 @@ open class ChartDataSet: ChartBaseDataSet
...
-                // [m] is closer to xValue
-                // Search in a lower place
-                high = m
+                formIndex(after: &closest)
             }
-            else
+
+        case .down:
+            // If rounding down, and found x-value is upper than specified x, and we can go lower...
+            if closestXValue > xValue && closest > startIndex
```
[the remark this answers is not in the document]
jjatie: In this case they are equivalent. I'm using `startIndex` as it is better to not assume the start is `0`, as slices do not guarantee `startIndex == 0`. It also makes it easier to migrate the algorithm to a generalized extension in the future if desired.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -260,98 +213,57 @@ open class ChartDataSet: ChartBaseDataSet
...
+                formIndex(before: &closest)
             }
-            
-            closest = high
+
+        case .closest:
+            break
         }
-        
-        if closest != -1
```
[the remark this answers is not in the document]
jjatie: The only case where `closest == -1` in the old implementation is if the dataset is empty (`low - high == -1`). This should probably be checked immediately, rather than later on.

Returning `-1` is very non-swift, which is why I've elected to drop that behaviour from everywhere I've found it in 4.0. 

`partitioningIndex` returns `endIndex` if no index is found in `startIndex..<endIndex`

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
liuxuan30: @jjatie https://swiftrocks.com/lazy-sequences-in-swift-and-how-they-work go to "LazyCollection Performance Traps"

I also find this one and point out that `lazy` for Sequence and Collection have sutle difference. Could you remind me what type of `self[i...]` is here? Sequence or Collection?

jjatie: And this is why I shouldn't code review so late at night! `self[i...]` produces a `Slice<ChartDataSet>` which is a `Collection`.

The reason `lazy` performed better in the first link is because `2...` defines a `Range` with bounds from `2` to `Int.max`. Calling and algorithm on this will cause a sequence to be created with every value between those bounds. Calling `lazy` on this will only create elements of the sequence until `prefix(while:)`'s predicate fails.

In our example, all of the values are already instantiated in the data set, so calling `lazy` won't improve performance (in fact it can actually negatively effect performance). I will revert the change.

liuxuan30: sorry :( I was not very familiar with this topic, thank god I read to the bottom...

liuxuan30:
the CI  failed on tvOS very weird, don't understand why. It seems a carthage issue but iOS and macOS build pass.

liuxuan30:
filed https://github.com/Carthage/Carthage/issues/3120 and waiting for some suggestions.

meanwhile, the iOS passed so I assume it's kind of safe to merge? @jjatie

jjatie:
I believe it is.

liuxuan30 (approved):

StefaniOSApps:
@jjatie @liuxuan30 ... 
`import Algorithms` can only be reached via SPM. This means that installation via CocoaPods is no longer possible for older projects. Can you fix that, please? Pull the repo, create a pod and add it to the Podfile by dependency.

ariofrio:
This seems to be causing a crash when the chart is tapped. To reproduce, run this code:

    let chart = LineChartView()
    chart.data = LineChartData(
      dataSet: LineChartDataSet(entries: [
        ChartDataEntry(x: 0, y: 0),
        ChartDataEntry(x: 10, y: 10),
        ChartDataEntry(x: 20, y: 20),
      ])
    )

And tap on the chart (I tapped near the middle). The app crashes.

When I reverted back to the previous commit 857db248f8eab9293e7845a53900ad109dc5cade this crash did not happen anymore.

darbysauter:
@StefaniOSApps Did you figure out a solution for this? I have a fork that I use to add some data race protections but it keeps telling me that its missing Algorithms

StefaniOSApps:
@darbysauter See #4572 ... I think that @liuxuan30 approved that PR too fast .. so wie lost cocoa pods in future

darbysauter:
@StefaniOSApps I ended up using my fork of Charts with SPM instead of (Cocoapods and trying to import Algorithms with SPM) works fine now

StefaniOSApps:
@darbysauter right, the best solution is a wrapper from swift to objc (only a workaround 😜)

NachoSoto:
FYI as others have mentioned this has broken CocoaPods and Carthage support.

PR #4531, opened 2020-12-14
Master branch: BarChartViewRender - remove extra shadow draw pass

bivant:
Leave the one similar to HorizontalBarChartRenderer although it is much longer. The deleted code did draw shadow in height of a bar, not the chartView.

### Goals :soccer:
Remove extra code after the merge with 4.0 branch

### DIFF
--- a/Source/Charts/Renderers/BarChartRenderer.swift
+++ b/Source/Charts/Renderers/BarChartRenderer.swift
@@ -343,18 +343,6 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
 
         let buffer = _buffers[index]
         
-        // draw the bar shadow before the values
-        if dataProvider.isDrawBarShadowEnabled
-        {
-            for barRect in buffer where viewPortHandler.isInBoundsLeft(barRect.origin.x + barRect.size.width)
-            {
-                guard viewPortHandler.isInBoundsRight(barRect.origin.x) else { break }
-
-                context.setFillColor(dataSet.barShadowColor.cgColor)
-                context.fill(barRect)
-            }
-        }
-        
         let isSingleColor = dataSet.colors.count == 1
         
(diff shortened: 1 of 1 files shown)


liuxuan30:
why removing this? I didn't see it's extra?

bivant:
Hello
This code (10 lines below) overwrites the drawing done by the deleted code. Am I wrong?
```swift
        for j in buffer.indices
        {
            let barRect = buffer[j]
            
            guard viewPortHandler.isInBoundsLeft(barRect.origin.x + barRect.size.width) else { continue }
            guard viewPortHandler.isInBoundsRight(barRect.origin.x) else { break }

            if !isSingleColor
            {
                // Set the color for the currently drawn value. If the index is out of bounds, reuse colors.
                context.setFillColor(dataSet.color(atIndex: j).cgColor)
            }
            
            context.fill(barRect)
```

liuxuan30:
I traversed back to some 2018 changes and it seems this logic has been long time existing. 

I think this logic is not duplicated, if you consider
>context.setFillColor(dataSet.barShadowColor.cgColor)

the `barShadowColor` is different, if the single color or more colors have alpha channel, the shadow color is more like a background color. if the colors are opaque, then it is true it's overwritten.

@jjatie anything that I would be wrong about the colors? if this is true, then we don't have to delete the code.

jjatie:
@bivant What is the issue that caused you to look for this? Were you just looking at diffs, or is there a visual glitch or performance issue?

@liuxuan30 Agreed.

bivant:
Hello @jjatie 
I was examining the drawing logic before implementing my own for the shadows/outlines, that probably lead me to a thought that this part is extra.
+was comparing it with the HorizontalBarChartRenderer, it does not have that piece of code.
So no, no visual glitch or performance issue.

PR #4539, opened 2020-12-21
SPM compilation fails due to missing UIKit

radianttap:
### Fixes Issue Link #4533 

- Adds UIKit import where needed.
- Removes obsolete `AxisRendererBase.swift` file.
- Fixes .package file with proper reference to swift-algorithms package

(hopefully this time included just the necessary commits)

PR #4550, opened 2021-01-06
BuildFix: fix build as part of SPM

ArthurValiev:
### Issue Link :link:

https://github.com/danielgindi/Charts/issues/4016

### Goals :soccer:

- During transition from CocoaPods to SPM, I've noticed multiple build failures originating from within Charts project
- The fixes seem trivial enough (missing imports, typos)

### Changes

- The most visible and common failure was that "NSTextAlignment" was failing to be located when building for iOS as target. (missing "import UIKit"). I'e made this import conditional based on the target OS. UIKit is imported for iOS and tvOS, AppKit is imported instead for macOS.
- "AxisRendererBase.swift" had two typos in "roundedToNextSignificant" call. 
- Also "AxisRendererBase.swift" did not conform to "Renderer" protocol properly. Fixed as suggested by Xcode.

### Testing Details :mag:
Performed some very basic testing by replacing the remote SPM dependency with a local one and adding it as a dependency to 3 projects that target macOS, tvOS and iOS. Build now successfully passes.
Tested basic graph cases in my own iOS project. Works fine in my use case.

### DIFF
--- a/Source/Charts/Charts/BarLineChartViewBase.swift
+++ b/Source/Charts/Charts/BarLineChartViewBase.swift
@@ -12,8 +12,10 @@
 import Foundation
 import CoreGraphics
 
-#if !os(OSX)
-    import UIKit
+#if os(iOS) || os(tvOS)
+import UIKit
+#elseif os(macOS)
+import AppKit
 #endif
 
 /// Base-class of LineChart, BarChart, ScatterChart and CandleStickChart.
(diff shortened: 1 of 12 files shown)


ArthurValiev:
Found one open similar PR https://github.com/danielgindi/Charts/pull/4539 which fixes missing imports but that one does not includes fixes for macOS. Build fails with macOS as target with missing AppKit.

jjatie:
Resolved by #4553

PR #4553, opened 2021-01-12
SPM GitHub Action

jjatie:
### Issue Link :link:
Resolves #4539 and #4550

### Goals :soccer:
Fix SwiftPM build, and add a GitHub Action to help ensure it is always buildable

Added typealiases in `Platform.swift` rather than add `#if` in many files for cross-platform support

### Testing Details :mag:
Github Actions run swift build to ensure it builds


andyj-at-aspin:
How about a 4.0.1 pre-release so that the iOS-based users can test for the 4.x changes as well without pulling master?

liuxuan30:
@andyj-at-aspin what is your point here? I'm a little confused as we have 4.0 released so you can test it? or you mean the new merges? like https://github.com/danielgindi/Charts/pull/4497

andyj-at-aspin:
I mean please create a numbered release that includes changes to master such as this specific commit. Before this commit, it was not possible to test Charts 4.x using SPM installation targeting iOS _at all_. After this commit, it's been possible _as long as you tell SPM to use the master branch_. It's not possible to do this if Charts is added as a sub-dependency of another Swift Package.

liuxuan30:
sure. I will address it recently. Need to merge another PR and fix the CI failures

hixfield:
I can build just fine now when pointing the SPM to the master branch. But indeed, an minor 4.0.1 tag would be much nicer.

andyj-at-aspin:
It's been a while since I last verified my Swift Package versions in my Apps. Thank you everyone concerned with the 4.0.1 release - especially Jacob and Xuan.

PR #4560, opened 2021-01-22
Fixed incorrect guard return statement when rendering limit lines

raytso:
### Issue Link :link:
Currently on 4.0.0 the rendering of limit lines has a bug to only render the first one in the `limitLines` array. This is caused by an incorrect early return in the guard statement instead of continuing through the loop

### Goals :soccer:
Fixes limit line rendering so it can render multiple lines correctly

No architectural changes

### Testing Details :mag:
No tests added


raytso:
@liuxuan30 I don't know who to tag but is anyone reviewing this?

jjatie:
Incorrectly targeting 4.0.0. Branch is removed to remove confusion

PR #4563, opened 2021-01-26
Fixed incorrect guard return statement when rendering limit lines

jjatie:
### Issue Link :link:
#4560

### DIFF
--- a/Source/Charts/Renderers/YAxisRenderer.swift
+++ b/Source/Charts/Renderers/YAxisRenderer.swift
@@ -293,7 +293,7 @@ open class YAxisRenderer: NSObject, AxisRenderer
             let label = l.label
             
             // if drawing the limit-value label is enabled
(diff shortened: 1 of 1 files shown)


jjatie:
@liuxuan30 Can you please diagnose the issue with the test when you have time? Appears to be a problem in our script.

awils18:
We are experiencing the same issue - I debugged locally and this is the issue. The other Axis renderers do contain the correct logic to continue if the label is empty or not enabled. 

See - https://github.com/danielgindi/Charts/blob/dea1cc54d11a1eccea9ec98c98c41a00f42b9793/Source/Charts/Renderers/YAxisRendererHorizontalBarChart.swift#L254

liuxuan30:
sorry not seeing this.. @jjatie  shall we release a new version like another thread says, for SPM support? Or moe to merge and release?

PR #4574, opened 2021-02-04
Replace FBSnapshotTestCase with pointfree/swift-snapshot-testing

jjatie:
FBSnapshotTestCase is all but deprecated. [pointfree's snapshotting library](https://github.com/pointfreeco/swift-snapshot-testing) is an excellent replacement that's become a bit of a community standard for snapshot testing. This also removes our Carthage dependency for tests.

Use github actions for testing. They consistent run faster, require less setup and satisfy our needs.


jjatie:
@liuxuan30 It [appears](https://github.com/pointfreeco/swift-snapshot-testing/issues/424) that Apple Silicon Macs create slightly different snapshots than Intel Macs. Could you checkout this branch and recreate the snapshots (and push) when you have time? (Xcode 12.4, iOS 14.4 iPhone 12 Pro, and tvOS 14.3 AppleTV 4K)

liuxuan30:
wow, thank you!!!

liuxuan30:
anything we need to keep in mind for this new framework? like a cheat sheet?

jjatie:
Only issue I've run into (which is true in FBSnapshot too) is that Intel and Silicon Macs generate slightly different snapshots. I've written a method `assertChartSnapshot` that compares to different images based on platform architecture to accommodate this.

There are many more options for snapshotting available in the library, beyond just images. Their GitHub page references the videos where they walkthrough creating the framework and using it.

Issue #4571, opened 2021-02-01
Line chart displaying two datasets leads to fatal error index out of range


poml88:
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Run the below code example

## What did you expect to happen?

A nice plot with two graphs

## What happened instead?

Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444
2021-02-01 21:11:49.285724+0100 testCharts[11213:212535] Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444

## Charts Environment

**Charts version/Branch/Commit Number: 4.0.0 **
**Xcode version: 12.3 **
**Swift version: 5 **
**Platform(s) running Charts: iOS **
**macOS version running Xcode: 10.15.6 **

## Demo Project

I'll post it below.

Hi, I just upgrade from 3.6.0 to 4.0.0 and found out that I have trouble with the below chart.
I have two datasets, and it worked fine with 3.6.0, but with 4.0.0 I get a
```
Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444
2021-02-01 21:20:12.111618+0100 testCharts[11321:217641] Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444
```

Xcode points me to ChartDataSet.swift to this section:
```
public subscript(position: Index) -> Element {
        get {
            // This is intentionally not a safe subscript to mirror
            // the behaviour of the built in Swift Collection Types
            return entries[position]
        }
        set {
            calcMinMax(entry: newValue)
            entries[position] = newValue
        }
    }
```

the error being in line 411 `return entries[position]`.

My example code runs fine, when I replace 
`let chartData = LineChartData(dataSets: [chartDataSet0, chartDataSet1])`
with
`let chartData = LineChartData(dataSets: [chartDataSet0])`
so the problem seems to be the two data sets.

A bug?

```
import UIKit
import Charts

class ViewController: UIViewController {
    
    @IBOutlet weak var chartView2: LineChartView!
    
    //    var dataEntries = [ChartDataEntry]()
    var chartDataSet0 = LineChartDataSet()
    var chartDataSet1 = LineChartDataSet()
    
    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view.
        
[...]

PR #4577, opened 2021-02-07
Fix bounds checks on binary search

jjatie:
### Issue Link :link:
#4571 

Incorrectly returned index beyond the last valid index. This is now corrected

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -214,7 +214,7 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var closest = partitioningIndex { $0.x >= xValue }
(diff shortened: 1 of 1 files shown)


InvaderZim62:
This change and possibly the changes leading up to it caused LineChartData(dataSets:) to stop plotting all data points, if the LineChartDataSet entries are not all the same size.  Only the first point is plotted for the shorter data set.  To reproduce, run this code:
```
let chart = LineChartView()
chart.data = LineChartData(
    dataSets: [
        LineChartDataSet(entries: [
            ChartDataEntry(x: 0, y: 0),
            ChartDataEntry(x: 10, y: 10),
            ChartDataEntry(x: 20, y: 20)
        ]),
        LineChartDataSet(entries: [
            ChartDataEntry(x: 0, y: 5),
            ChartDataEntry(x: 10, y: 15)
        ])
    ]
)
```
When I reverted back to the earlier commit 857db24 this code plots all five data points.

Issue #4049, opened 2019-06-28
EXC_BREAKPOINT on LineChartRenderer.drawCircles


slzrafael:
I need to show Line graphs. I use regular API of Charts to show any graph, and it works perfectly in most case, but i receive in crashlyhtics some errors:

**Crashed: com.apple.main-thread
EXC_BREAKPOINT 0x00000001060e9ad4**

> Crashed: com.apple.main-thread
> 0  Charts                         0x1060e9ad4 LineChartRenderer.drawCircles(context:) + 108 (BarLineScatterCandleBubbleRenderer.swift:108)
> 1  Charts                         0x106052f44 BarLineChartViewBase.draw(_:) + 237 (BarLineChartViewBase.swift:237)
> 2  Charts                         0x106053f9c @objc BarLineChartViewBase.draw(_:) (<compiler-generated>)
> 3  UIKitCore                      0x1edade560 -[UIView(CALayerDelegate) drawLayer:inContext:] + 460
> 4  QuartzCore                     0x1c53395c4 -[CALayer drawInContext:] + 296
> 5  QuartzCore                     0x1c521d970 CABackingStoreUpdate_ + 176
> 6  QuartzCore                     0x1c53400d0 invocation function for block in CA::Layer::display_() + 52
> 7  QuartzCore                     0x1c5338e88 -[CALayer _display] + 1520
> 8  QuartzCore                     0x1c52a1388 CA::Context::commit_transaction(CA::Transaction*) + 532
> 9  QuartzCore                     0x1c52cf330 CA::Transaction::commit() + 640
> 10 QuartzCore                     0x1c52cff20 CA::Transaction::observer_callback(__CFRunLoopObserver*, unsigned long, void*) + 92
> 11 CoreFoundation                 0x1c0e165f8 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 32
> 12 CoreFoundation                 0x1c0e11320 __CFRunLoopDoObservers + 412
> 13 CoreFoundation                 0x1c0e1189c __CFRunLoopRun + 1228
> 14 CoreFoundation                 0x1c0e110b0 CFRunLoopRunSpecific + 436
> 15 GraphicsServices               0x1c301179c GSEventRunModal + 104
> 16 UIKitCore                      0x1ed647978 UIApplicationMain + 212
> 17 Easynvest                      0x104ac72d4 main + 17 (main.swift:17)
[...]

Issue #4046, opened 2019-06-27
Fatal error: Can't form Range with upperBound < lowerBound


danishnaeem57:
hey @danielgindi , first of all, thanks for the awesome library. 
I have recently updated the library and a crash is occurring in 
"extension BarLineScatterCandleBubbleRenderer.XBounds: Sequence" in constructor. line is 

fileprivate init(min: Int, max: Int) {
 ----->           self.iterator = (min...max).makeIterator()
}

Can you please tell me, why is it crashing and what am I doing wrong and how can I prevent it? Its astonishing that graphs have been working fine on previous versions of library.

Error on xcode debugger is:
Fatal error: Can't form Range with upperBound < lowerBound

d0x90:
same here

ali-ehsan:
Same here

Cavech84:
same here

xor22h:
Me too... Seems like min/max values are in wrong order.

t0rn:
same here

Andy1984:
same here

Andy1984:
But you hint me, you recently update the library, thus I can turn to a lower version to solve the problem temporarily.

moosabaloch:
Downgraded the library from **3.3.0** to **3.2.2** and it works fine for now.

pandapancake:
Same here. Even after downgraded the versions to 3.2.2 :/

Andy1984:
> Same here. Even after downgraded the versions to 3.2.2 :/

3.2.2 is fine. I recommend you clean xCode and run again.

psseelman:
@danishnaeem57 @d0x90 @ali-ehsan  @Cavech84  @xor22h @t0rn @Andy1984 @moosabaloch  @pandapancake 

As a temporary fix, when you're creating your data sets before adding them to the chart sort them in place by their x values min < max. It doesn't fix the bug obviously but it'll let you draw your graphs.

`
        var dataEntriesLive: [ChartDataEntry] = []
        
        for i in 0 ..< live.count {
            let liveDataPoint = live[i]
            let power = liveDataPoint.power
            let freq = liveDataPoint.freq
            dataEntriesLive.append(ChartDataEntry(x: Double(freq), y: power))
        }
        
        dataEntriesLive.sort(by: { $0.x < $1.x })
        
        let chartData = LineChartData(dataSets: [])

        chartData.addDataSet(chartDataSetLive)
`

Chunontherocks:
[...]

PR #4592, opened 2021-03-03
Prevent creation of invalid range in XBounds (fixes #4184 and #4049)

### LINKED ISSUE #4184: crash Fatal error: Can't form Range with upperBound < lowerBound
I write a demo with   LineChartView just as your demo code.
if I set the data with follow with crash :

        chartsDataArray.append(ChartDataEntry(x: 10, y: 20))
        chartsDataArray.append(ChartDataEntry(x: 30, y: 12))
        chartsDataArray.append(ChartDataEntry(x: 8, y: 24))
        chartsDataArray.append(ChartDataEntry(x: 5, y: 14))
        chartsDataArray.append(ChartDataEntry(x: 4, y: 4))

I just simple want draw a line with data above,  it crashed with   crash Fatal error: Can't form Range with upperBound < lowerBound.

But if I adjust the order of the data  list , it won't crash.

I use XCode11 and Swift5.0 .

Thank Your !

berendkleinhaneveld:
### Issue Links 🔗 

The issues #4184 and #4049 mention crashes which are caused by an invalid range in the XBounds. This fix might not really solve all underlying problems, but at least it will make it stop crashing. Might need some further investigation.

### Goals :soccer:
This PR should prevent crashes due to invalid ranges in BarLineScatterCandleBubbleRenderer.XBounds (max > min). An negative range will cause a crash when the iterator is called upon.
I traced the problem to the `set(chart, dataset, animator)` function in `BarLineScatterCandleBubbleRenderer.XBounds`. If `entryFrom` is not nil but `entryTo` is, and there are entries outside/below the visible range (or when the entries are not sorted), then the `entryIndex(entry:)` function will/can return an index larger than zero. This results in `min` becoming larger than zero while `max` will default to 0. This creates an invalid range.

I've made sure that when the `entryFrom` value is nil while the `entryTo` values is non-nil, that the min value will be the max value (instead of the default 0). And when `entryFrom` is non-nil but `entryTo` is nil, then the max value will be set to the min value to prevent a negative/invalid range.

### Testing Details :mag:
The test uses a ScatterChartView with a custom `axisMinimum` and `axisMaximum` and a simple dataset where two values are below/outside the range. The last value lies exactly on the lower range. This results in a nil value for `entryTo` and a non-nil value for `entryFrom`. And this would result in an invalid range with the previous code.

The call stack is as follows: `entryForXValue(_:closestToY:rounding:) => entryIndex(x:closestToY:rounding:) => entryIndex(x:closestToY:rounding:) => Collection.partitioningIndex`.
The `partitioningIndex` function relies on the premise that the collection is already partitioned/sorted(!). Would be good to add a notion somewhere in the documentation that it is a good idea to keep datasets sorted, so if anybody has a good idea of where I could put that, I'd be happy to add it in the PR. I don't have a lot of experience with the code base so I don't know exactly the ramifications of the proposed change, but at least it prevents some nasty crashes and I haven't seen any side effects yet in my (limited) testing.

### DIFF
--- a/Charts.xcodeproj/project.pbxproj
+++ b/Charts.xcodeproj/project.pbxproj
@@ -25,6 +25,7 @@
 		135F11CE20425AF600D655A3 /* PieChartTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 135F11CD20425AF600D655A3 /* PieChartTests.swift */; };
 		146EE16342C2BADC92E45BF2 /* LineScatterCandleRadarChartDataSetProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9249AD9AEC8C85772365A128 /* LineScatterCandleRadarChartDataSetProtocol.swift */; };
 		17E994DA88777AA1D8CCFC58 /* BarChartDataSet.swift in Sources */ = {isa = PBXBuildFile; fileRef = C31AA65EA27776F8C653C7E8 /* BarChartDataSet.swift */; };
+		194ECA1825EED8FC00BC6251 /* RangeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 194ECA1725EED8FC00BC6251 /* RangeTests.swift */; };
 		219192CA6B4895319AB49DCA /* BarLineScatterCandleBubbleRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B1C588E9DF6FFD56D7ADF8E /* BarLineScatterCandleBubbleRenderer.swift */; };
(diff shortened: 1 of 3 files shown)


kscheff:
This may also be a fix for #4046

jjatie:
Closing in favour of #4596. @berendkleinhaneveld I would appreciate any help in testing the updated implementation. You can find the behaviour (that always worked as far as we know) [here](https://github.com/danielgindi/Charts/commit/2a1ecb4e6f175f91e91d77d039be342138fe0e3c).

I'm not convinced we should use `StrideThrough.Iterator`, but it guarantees the last known working behaviour.

berendkleinhaneveld:
Cool, I'll check it out!

Issue #4579, opened 2021-02-08
Value highlighting bug


stealdre:
Why does for any simple chart value, highlighting is very laggy and misses most of the chart points ?

sjacs5537:
Same here. When I switch back to v4.0.0, everything works fine.

pcmanik:
Same problem here

roed:
Also v3.6.0 works fine, v4.1.0 is broken/laggy here

PR #4610, opened 2021-03-24
Fix for issue where data points are not highlighted

### LINKED ISSUE #4609: Highlighting buggy / not working consistently
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Added a line chart control to show time series chart data and enabled highlighting so that when the chart is tapped the tapped value is highlighted.

The entry closest to the tap point should be highlighted

## What happened instead?

* The majority of the time no entry is highlighted even after tapping all around the chart
* Rarely an entry actually is highlighted but given hightlighting fails the majority of the time the overall behavior is very poor

## Charts Environment

**Charts version/Branch/Commit Number:** 4.0.1/master/e91ba716190836b013ef8a9ca53e220ad5051e21
**Xcode version:** 12.4
**Swift version:** 5
**Platform(s) running Charts:** simulator iPad Air (4th Generation) 14.4
**macOS version running Xcode:** 10.15.6

## Demo Project

https://github.com/gavynriebau/DemoChartsBug

----

# Details about the demo project

The Chart

_**Figure 1**: The chart that is rendered when this project is run_

Figure 2 shows a stack trace which highlights the problem area.

This stack trace was generated by tapping a point just after the 5th chart entry.

In the debug window it can be seen that the variable `i` is set to `11` indicating the `partitionIndex` func did not correctly find the index of the first item that matches the predicate on line 199.

Given the argument `xValue` having value `6` the expected return for `i` should be `6`.

The values contained within the `ChartDataSet`, along with their indexes, are printed in the bottom right of figure 2.

The overall result is that the `entriesForXValue` func returns an empty array and the chart highlighting / selection fails for that tap.

Stack

_**Figure 2**: Call stack showing where filtering goes wrong_

gavynriebau:
### Issue Link :link:

Fixes #4609 and possibly also #4579 

### Goals :soccer:

Improve the highlighting / selection behavior of line charts

The `entriesForXValue` func finds the data entries for a given x-axis
value. It used an optimization where it would do a binary search to find
the first matching entry then take a slice of data stopping when no more
entries matched.

The optimization broke the highlight feature because data points that
should have matched were not matching resulting in no entries being
returned.

This commit replaces the optimized code with a call to `.filter` which
will be slower for larger data sets but correctly returns filtered
entries.

### Testing Details :mag:

Tested by altering the [demo project](https://github.com/gavynriebau/DemoChartsBug) to use the forked version of the code and confirming the broken behaviour works after the fix is applied

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -196,10 +196,7 @@ open class ChartDataSet: ChartBaseDataSet
     /// An empty array if no Entry object at that index.
     open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
     {
-        let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
-        let i = partitioningIndex(where: match)
-        guard i < endIndex else { return [] }
-        return self[i...].prefix(while: match)
+        return self.filter { $0.x == xValue }
     }
     
     /// - Parameters:


mlatham (commented):
I also encountered this issue, so bump on this PR. I took a different approach to fixing it in our fork that I think addresses the regression while maintaining the optimized approach. The partitioningIndex just wasn't being calculated correctly.

```
/// - Returns: All Entry objects found at the given xIndex with binary search.
/// An empty array if no Entry object at that index.
open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
{
    let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
    let i = partitioningIndex { $0.x >= xValue }
    guard i < endIndex else { return [] }
    return self[i...].prefix(while: match)
}```

gavynriebau:
@mlatham thanks for the code example, I tested out your fix and think you're correct it was just the filter predicate that was wrong.

In light of the above I'm closing this PR and have opened a simpler PR which still fixes the issue here:
#4613

HWS123:
nice

Issue #4641, opened 2021-05-11
Bug: BarChartView Highlight Can Select Some Bar On Chart


SchwartzLizer:
## What did you do?

Create BarChartView

## What did you expect to happen?

Can touch all bar in the chart

## What happened instead?

which bar can select, some can't select

## Charts Environment

Charts version/Branch/Commit Number: 4.0.1 Via SPM (Install Charts Only , Chart-Dynamic Didn't (Is show no no such module algorithms))
Xcode version: 12.5
Swift version: 5
Platform(s) running Charts: iOS
macOS version running Xcode: 11.3.1

## Demo Project

![CleanShot 2021-05-11 at 11 29 04](https://user-images.githubusercontent.com/54163797/117758781-d4a87300-b24c-11eb-89ff-72bb8a5f21af.gif)

## On My Projects

![CleanShot 2021-05-11 at 11 48 24](https://user-images.githubusercontent.com/54163797/117759950-e559e880-b24e-11eb-9f6e-30ad578baa82.gif)

SchwartzLizer:
Update Infomation: When I back to version 3.6.0 via SPM it still worked I hope this bug will be fixed soon (V4.0.0 is still not working too)

orolhawion:
I can confirm this issue and also experience it with line charts. However, markers in pie charts are working as expected.

Lee-Waire-Health:
var entries: [ChartDataEntry] = []
    
    
    var i = 0
    while i < 38 {
        entries.append(ChartDataEntry(x: Double(round(1621858800.0+(Double(i)*60.0))), y: Double(Float.random(in: 70...73))))
        i+=1
    }
    
    let set = ChartDataSet(entries: entries)
    
    let res = set.entriesForXValue(Double(1621860300))
    
    let test1 = entries.partitioningIndex { $0.x == Double(1621860300) }
    let test2 = entries.partition { $0.x >= Double(1621860300) }


^ ^ simplest for of replication think it points to bad use or partitioningIndex, use comments suggest this generally should only be run on a partitioned set.

ARASHz4:
I have same problem in bar chart

hojideli:
I have the same problem too

BeCrecker:
Having same issue too

yusufonderd:
Same issue

ajkiosd:
subclass `ChartDataSet` like below and using it will fixes this issue for now:
```

class CustomBarChartDataSet: BarChartDataSet {

[...]

PR #4647, opened 2021-05-25
Fixes 4641 & 4645

Lee-Waire-Health:
The use of Swift Algorithms  partitioningIndex with exact matching clause == can provide incorrect values.
It has been modfied to use a >= clause as well followed by a check for an exact match from that index.
This should maintain the intended behaviours and I have added a
It has been modfied to use a >= clause as well followed by a check for an exact match from that index.
This should maintain the intended behaviours and I have added a fix for fetching the closest X value index when the value falls between 2 indexs and uses .closest for rounding method.

### Issue Link :link:
[4641](https://github.com/danielgindi/Charts/issues/4641#issue-885984081)

### Goals :soccer:
- Replicate error in Test Suite
- Fix the bug in entriesForXValue

- Fixed the usage of partitioningIndex in entriesForXValue with minimal changes.
- added support for .closest for X values in entryIndex as it seemed to be missing since the Swift Algo port.

### Testing Details :mag:
ChartDataTests -> testEntriesForXValue

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -196,10 +196,11 @@ open class ChartDataSet: ChartBaseDataSet
     /// An empty array if no Entry object at that index.
     open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
     {
-        let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
+        let match: (ChartDataEntry) -> Bool = { $0.x >= xValue }
         let i = partitioningIndex(where: match)
-        guard i < endIndex else { return [] }
-        return self[i...].prefix(while: match)
+
+        guard self[i].x == xValue else { return [] }
+        return self[i...].prefix(while: { $0.x == xValue })
     }
     
     /// - Parameters:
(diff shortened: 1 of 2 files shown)


Lee-Waire-Health:
@danielgindi I'm surprised at the low attention this has gotten as I'd have thought this feature was pretty core to the libraries example usage.

Is it just a time issue or does this PR need to be of better quality in some way?

daSkier:
I can confirm this PR fixes the issues I've been encountering with markers in my line charts.

bohemima:
Also confirming that this fixes issues with highlighting values that I've been seeing, thanks for the PR @Lee-Waire-Health, @danielgindi  would love for this to be incorporated into the main repo.

be-bert:
Agree @bohemima, it would be great to get this incorporated @danielgindi.  Thanks for your work @Lee-Waire-Health!

kscheff:
One issue here is that the partitioningIndex fails when xValue to search is larger than any $0.x, it returns indexEnd. Consequently the guard statement fails and the function returns empty []. This the .closest strategy does not return the nearest entry.

The issue arrises when the data line should get drawn where the highestVisibleX value is larger than any entry.x.
 See #4687 for a fix of not displaying the line.

benck:
@Lee-Waire-Health Thanks for the fix, I am temporarily switching to your branch!

GinesSanchez-TopTracer (approved):

pmairoldi:
@Lee-Waire-Health @GinesSanchez-TopTracer could you validate that https://github.com/danielgindi/Charts/pull/4721 fixes the problem you were having here

PWrzesinski:
@pmairoldi I don't think the PR you've mentioned fixes the incorrect use of the partitioning algorithm. Would love to get this PR merged as I've just discovered the same issue (the hard way)!

Issue #4719, opened 2021-09-20
ChartViewDelegate Function func chartValueSelected(_ chartView: ChartViewBase, entry: ChartDataEntry, highlight: Highlight) is not being called for every tap/touch inside the ChartView


kiranGrg:
* [ ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

I upgraded the Charts(iOS) version to 4.0.1 from 3.6.0 recently. Fixed all the changes in the codebase. Everything seems to be working except for one behavior on the chart. 

## What did you expect to happen?

Enabled the user interaction on the chart. Added the ChartViewDelegate, and implemented the optional delegate function
`public func chartValueSelected(_ chartView: ChartViewBase, entry: ChartDataEntry, highlight: Highlight)`

## What happened instead?

However, apparently, this delegate function `public func chartValueSelected(_ chartView: ChartViewBase, entry: ChartDataEntry, highlight: Highlight)` is not being triggered every time I tap inside the chart view (to be specific I was using LineChartView), instead only for some places in the chart view seems to be responding to the tap/touch.

I am pretty, sure the delegate function used to be called with failing in the Charts 3.6.0.

## Additional Findings:
I tried the official demo version for iOS source codes for both 3.6.0 and 4.0.1, as i couldn't get this issue fixed. Looks like, the demo itself is exhibiting this problem, as I could see the delegate function being called with every tap/touche in 3.6.0 but not in the demo version 4.0.1

## Charts Environment

**Charts version/Branch/Commit Number: 4.0.1**
**Xcode version: 12.5.1**
**Swift version: 5.4.2**
**Platform(s) running Charts: iOS device/simulators**
**macOS version running Xcode: 11.5.2**

esma01:
I see the same issue. The sample project is also not working! Only the middle and last data points can be selected.

SergejLogis:
Same here, 100% accurate description of an issue and same observations (Charts demo app)...

PhillipMaizza:
I'm experiencing the same issue with line charts.

esma01:
@PhillipMaizza @SergejLogis Downgrading the charts to 3.6.0 worked for me!

SergejLogis:
[...]

PR #4721, opened 2021-09-23
Highlighter fix

kcome:
Fixed the use of partitioningIndex(where:) in ChartDataSet.entriesForXValue
Fixed the rounding issue of ChartDataSet.entryIndex()

### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4719

### Goals :soccer:
Having a working highlighter which is broken due to incorrect use of Algorithms `partitioningIndex()`

Partition the entries first, then call `partitioningIndex()`

### Testing Details :mag:
Using the attached `ChartDemo-iOS-Swift` project

Open LineChart1 or LineChart2, tap data points, highlight should work

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -101,8 +101,9 @@ open class ChartDataSet: ChartBaseDataSet
 
         guard !isEmpty else { return }
         
-        let indexFrom = entryIndex(x: fromX, closestToY: .nan, rounding: .down)
-        let indexTo = entryIndex(x: toX, closestToY: .nan, rounding: .up)
+        let indexFrom = entryIndex(x: fromX, closestToY: .nan, rounding: .closest)
+        var indexTo = entryIndex(x: toX, closestToY: .nan, rounding: .up)
+        if indexTo == -1 { indexTo = entryIndex(x: toX, closestToY: .nan, rounding: .closest) }
         
         guard indexTo >= indexFrom else { return }
         // only recalculate y
@@ -216,7 +217,7 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var closest = partitioningIndex { $0.x >= xValue }
-        guard closest < endIndex else { return -1 }
+        guard closest < endIndex else { return rounding == .closest ? (endIndex-1) : -1 }
 
         let closestXValue = self[closest].x
 
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -197,9 +197,11 @@ open class ChartDataSet: ChartBaseDataSet
     open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
     {
         let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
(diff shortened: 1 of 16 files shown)


krzysztof-grobelny (approved):

kcome:
@krzysztof-grobelny The issue #4719 happens on both Xcode 13 and 12.

There are two fixes together solves #4719 :

1 is in `class ChartDataSet`, function  open override `func entryIndex(x xValue: Double, closestToY yValue: Double, rounding: ChartDataSetRounding) -> Int`. When using "closest" rounding method, there was no checking present. E.g. in data-set [(1,1), (2,2), (3,3)], when trying to find closest point to xValue 2.1 with closest rounding, previous solution gives (3, 3) but actually the closes point is (2,2).

2 is in `open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]`. According to swift-algorithm [document](https://github.com/apple/swift-algorithms/blob/main/Guides/Partition.md):

> The partitioningIndex(where:) method returns the index of the start of the second partition when called on an already partitioned collection.

Previous implementation didn't call `.partition(by: match)` so the result of `.partitioningIndex(where: match)` is not what we want.

lewis-smith (approved):

kcome:
@liuxuan30 could we merge this PR?

kcome:
Hi @pmairoldi , I resolved a conflict in this PR. Could we start merge? Let me know if there is anything else need to be done.

dudenamedjune:
Think I have a proper fix open for this https://github.com/danielgindi/Charts/pull/4817
https://github.com/danielgindi/Charts/issues/4719#issuecomment-1107433833

417-72KI:
I tried with #4817 and seems to work `chartValueSelected(_:entry:highlight:)`,
but `chartValueNothingSelected(_:)` still not worked.

This PR works both and seems to be expected specs ☺️

417-72KI (approved):

pmairoldi (approved):
Thanks! Seems to work fine.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -197,9 +198,11 @@ open class ChartDataSet: ChartBaseDataSet
     open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
     {
         let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
```
PWrzesinski, line 200: I believe this line should be:

    let match: (ChartDataEntry) -> Bool = { $0.x >= xValue }

Note the >= instead of ==.

pmairoldi: Can you explain why?

PWrzesinski: Actually nevermind, sorry for bothering you! Please disregard this and other comments - I'll delete them.

My thinking was that `partitioningIndex` function requires that all values after the desired index match the expression, but after looking more carefully I see you've added the `partition` function before running `partitioningIndex` - I'm guessing your intention is to make it possible to search in non-ordered data. I'm a little worried about the performance here, but otherwise I think the implementation is correct, except one detail: comparing Doubles directly. This is probably bigger discussion about the library design, but basically due to Double representation in memory, we should never compare Doubles using the equality operator. This will break for example in a simple case: 0.3 != 0.1 + 0.2. But as I said, that's probably a separate discussion.

wlxo0401:
chartValueNothingSelected not work

Issue #4668, opened 2021-06-22
entriesForXValue() not finding point with matching xValue when it should


darbysauter:
* [x ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Tapped a point on my LineChart 

## What did you expect to happen?

Highlight a point 

## What happened instead?

Failed to highlight point

## Charts Environment

**Charts version/Branch/Commit Number:master**
**Xcode version:Latest**
**Swift version:Latest**
**Platform(s) running Charts:iPhoneX**
**macOS version running Xcode:BigSur**

## Demo Project

Unable to produce

## Reason for this to happen

In entriesForXValue(_ xValue: Double) -> [ChartDataEntry] of ChartDataSet
it uses partitioningIndex() to find the first index that matches the xValue.

The xValue is from:
closest = set.entryForXValue(xValue, closestToY: .nan, rounding: rounding)

Which I have verified is working properly:
<img width="1016" alt="image" src="https://user-images.githubusercontent.com/40215561/123013096-c2a80d00-d380-11eb-9c82-611ea8241b33.png">

As you can see xValue matches the x value of the entry at index 3. using firstIndex() also produces index 3, but partitioningIndex() returns the the index after the last of entries.

So to debug this, after the case where it could not find the index I manually check every entry and Int round them, If I find a match I then test firstIndex() using the same predicate as partitioningIndex() (so no Int rounding), and this works.

partitioningIndex() does work sometimes but I couldn't figure out a common cause.

Assuming my assumption that partitioningIndex() and firstIndex() should do the same thing then partitioningIndex() is broken.

Any ideas how to properly fix this? for the meantime I may just use firstIndex() on my fork.

ggorlim:
I have same issue, so as workaround i rewrite method entriesForXValue, using code from older lib version:

`override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
    {
        var entries = [ChartDataEntry]()
        var low = startIndex
        var high = endIndex - 1        
        while low <= high...`

This work, but we have some other issues, i m not think they are related to this, like rendering only first from multiple graphs..

I think this issue also have same problem : #4645 Markers not appearing as expected

lancelot9527:
> I have same issue, so as workaround i rewrite method entriesForXValue, using code from older lib version:
> 
[...]

Issue #4645, opened 2021-05-21
Markers not appearing as expected


orolhawion:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?
In my project I updated Charts with Swift Package Manager from 3.6.0 to 4.0.1 and found that markers on line charts and bar charts are not appearing reliably.

## What did you expect to happen?
When I tap an entry in a bar chart or line chart, the marker appears reliably in v3.6.0 but not in v4.0.1. However I could not find this issue with pie charts, this seems to be working as before.

## What happened instead?
Nothing, no errors in the log

## Charts Environment
**Charts version/Branch/Commit Number:**
4.0.1

**Xcode version:**
12.5 (12E262)

**Swift version:**
5

**Platform(s) running Charts:**
iOS 14.5+

**macOS version running Xcode:**
Big Sur 11.3.1

## Demo Project
Since I did not change any charts related code in my project, I expect this to be reproducible with any project that you guys have that use markers.

This might be related to https://github.com/danielgindi/Charts/issues/4641#issue-885984081

Lee-Waire-Health:
var entries: [ChartDataEntry] = []
        
        
        var i = 0
        while i < 38 {
            entries.append(ChartDataEntry(x: Double(round(1621858800.0+(Double(i)*60.0))), y: Double(Float.random(in: 70...73))))
            i+=1
        }
        
        let set = ChartDataSet(entries: entries)
        
        let res = set.entriesForXValue(Double(1621860300))
        
        let test1 = entries.partitioningIndex { $0.x == Double(1621860300) }
        let test2 = entries.partition { $0.x >= Double(1621860300) }

astarsoftware:
I also ran into this bug when updating to 4.0.1.

samdogg7:
I ran into this same bug as well on 4.0.0

drewster99:
I have this issue also on 4.0.1.  None of the markers appear correctly when tapped.  Either they don't appear at all or they appear int he wrong place entirely (or perhaps it's a different marker that appears).  Updated from 3.6.0 to 4.0.1.

samdogg7:
@drewster99 I ended up downgrading to 3.6.0 on SPM as a workaround. Still no word on a fix.

PR #4731, opened 2021-10-06
Fixes markers not appearing as expected (#4645) and entriesForXValue not finding match (#4668)

drewster99:
Resolves Issue 4668 "entriesForXValue() not finding xValue when it should (https://github.com/danielgindi/Charts/issues/4668).

Resolves Issue 4645 "Markers not appearing as expected" (https://github.com/danielgindi/Charts/issues/4645).

The entriesForXValue function was incorrectly partiioning on ==, and also sometimes incorrectly returning an array of values that don't match at all.
The entryIndex function's switch statement assumes that the closest X value had already been found, which would often not be the case, since we are partitioning on >=.

### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4668
https://github.com/danielgindi/Charts/issues/4645

The entriesForXValue function was incorrectly partiioning on ==, and also sometimes incorrectly returning an array of values that don't match at all.
The entryIndex function's switch statement assumes that the closest X value had already been found, which would often not be the case, since we are partitioning on >=.

### Testing Details :mag:
Tested in my own app which uses vertical bar charts, stacked vertical bars, and use annotations/markers for everything when tapped.  This functionality was completely broken when I transitioned from 3.6.0 to 4.0.1.   Also ran app while testing and running both the old (pre-Swift Numerics) implementation of these two functions along side the new, asserting on any differing results.  Should be solid.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -196,10 +196,9 @@ open class ChartDataSet: ChartBaseDataSet
     /// An empty array if no Entry object at that index.
     open override func entriesForXValue(_ xValue: Double) -> [ChartDataEntry]
     {
-        let match: (ChartDataEntry) -> Bool = { $0.x == xValue }
-        let i = partitioningIndex(where: match)
-        guard i < endIndex else { return [] }
-        return self[i...].prefix(while: match)
+        let i = partitioningIndex(where:  { $0.x >= xValue })
+        guard i < endIndex, self[i].x == xValue else { return [] }
+        return self[i...].prefix { $0.x == xValue }
     }
     
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


hidollars-ios:
This is working good and should be merged.

myihsan:
badge
It may be better to revert #4497 and any other related to it, since `partitioningIndex(where:)` is not designed for search according to the [document](https://github.com/apple/swift-algorithms/blob/main/Guides/Partition.md).
Also, it's slower than the original implementation, since they are using the same algorithm but `partitioningIndex(where:)` will not stop when the target is found.

@drewster99  @jjatie @liuxuan30
What do you think?

4brunu:
This PR changes fixed the MakerView behaviour.

@pmairoldi @liuxuan30 @jjatie @nuomi1 could you please help to land this fix?

Thanks

ajisaputraajaib:
Please merge this PR and add it to next release

shahryar2001:
Yes please merge it. It is working...

mylovelycodes:
good job, it works for me.
thanks @drewster99

pmairoldi:
Closed in favour of https://github.com/danielgindi/Charts/pull/4721

drewster99:
> badge It may be better to revert #4497 and any other related to it, since `partitioningIndex(where:)` is not designed for search according to the [document](https://github.com/apple/swift-algorithms/blob/main/Guides/Partition.md). Also, it's slower than the original implementation, since they are using the same algorithm but `partitioningIndex(where:)` will not stop when the target is found.
> 
> @drewster99 @jjatie @liuxuan30 What do you think?

Good question.  In my fork I considered reverting all of that as well for the same reason and did some quick tests, like, "call this function 1000 times and see how long it takes".  I don't remember the specific results, but decided it was innocuous, so I kept my fix.  (Also, of course, I wanted to minimize the changes between my fork and the official current version.)  

Was anything else changed with the use of the swift algorithms?  Are we losing something by reverting it?

Issue #4780, opened 2022-02-24
Y axis labels improperly sized causing labels to be cut off in some cases


drewster99:
Y axis labels (such as leftAxis) are sometimes cut off (horizontally).  This is due to a bug in YAxis.swift.  The current code takes the longest label string, where length is determined in terms of characters, and uses its size to determine the amount of width allocated to y axis labels.  In some cases, such as is often the case with custom fonts, there are multiple labels of the same number of characters, but where one of them may take up more screen space than the other.

* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Displayed Y axis labels on the left axis

## What did you expect to happen?

I expected that they would not be clipped, but rather sized for the largest label.

## What happened instead?

They are not always sized for the largest label, which results in sometimes clipping part of the label.

## Charts Environment

**Charts version/Branch/Commit Number: 85cfba96eb3492124d645ac0901b66f0cb266267**
**Xcode version: 13.2.1**
**Swift version: 5.5**
**Platform(s) running Charts: iPadOS 15.1**
**macOS version running Xcode: 12.2.1**

## Demo Project

n/a

PR #4781, opened 2022-02-24
Issue 4780 Y Axis Labels properly sized - clean fix

drewster99:
### Issue Link :https://github.com/danielgindi/Charts/issues/4780:
Y axis labels are improperly sized in some cases

Old broken logic: Label size was determined by the CGSize of the label with the largest number of characters
New logic: Label size is determined by the CGSize of the label with the largest CGSize


drewster99:
@berbaspin @pmairoldi Any chances we can get this merged?

drewster99:
how about now?

PR #4801, opened 2022-03-25
Properties on axis renderers to Objc (Fixes #4800)

### LINKED ISSUE #4800: Properties on XAxisRenderer and YAxisRenderer are not marked Objc
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?
Tried to use XAxisRenderer.axis property from ObjC

Property to be available.

## What happened instead?

Property is not available from Objc, but is from Swift

## Charts Environment
**Charts version/Branch/Commit Number:** 4.0.2

SLambrakis:
This makes the three public properties on XAxisRenderer and YAxisRenderer, available for access from Objc.

### Goals :soccer:
To make it possible to use these classes from Objc

These classes and many of the functions in them are already accessible from Objc. I think that the properties should also be (and indeed, they were until the move from `AxisRendererBase` to `AxisRenderer` in [#4456](https://github.com/danielgindi/Charts/pull/4456,))

### DIFF
--- a/Source/Charts/Renderers/XAxisRenderer.swift
+++ b/Source/Charts/Renderers/XAxisRenderer.swift
@@ -16,9 +16,9 @@ import CoreGraphics
 @objc(ChartXAxisRenderer)
 open class XAxisRenderer: NSObject, AxisRenderer
 {
-    public let viewPortHandler: ViewPortHandler
-    public let axis: XAxis
-    public let transformer: Transformer?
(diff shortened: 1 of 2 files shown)


pmairoldi (approved):

pmairoldi:
Thanks

Issue #4798, opened 2022-03-21
Crash building iterator. Range requires lowerBound <= upperBound


sebastienhannay:
* [ x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Upgrade from 3.6.0 to 4.0.2

## What happened ?

Most drawing of charts leads to a crash when building iterator in 
`extension BarLineScatterCandleBubbleRenderer.XBounds: Sequence`
because Range requires lowerBound <= upperBound

`let entryTo = dataSet.entryForXValue(high, closestToY: .nan, rounding: .up)` is nil. It looks like the closest found is < endIndex in the dataset.

I use custom minimum and maximum axis values and visible range.

The same code was working perfectly in v3.6.
I also tried to rollback to 3.6 as a temporary workaround but the project doesn't compile anymore with XCode 13.3.

## Charts Environment

**Charts version/Branch/Commit Number: 4.0.2**
**Xcode version:  13.3 (13E113)**
**Swift version: 5**
**Platform(s) running Charts: iOS**
**macOS version running Xcode: 12.0.1 (21A559)**

pavel-zlotarenchuk:
Some problem

<img width="1114" alt="image" src="https://user-images.githubusercontent.com/36473231/161632361-a3f79945-7862-4953-8bb6-8b39abbaa730.png">

k1ran-ak:
This can happen when max values are 0 or negative values. make sure the x values are not 0 or negative in the data entry x value array you add to the Chart Data entry.

sebastienhannay:
hello @k1ran-ak, I just checked and I don't have any negative value in my data :
```
Charts.ChartDataSet	Charts.ChartDataSet	
Charts.ChartBaseDataSet	Charts.ChartBaseDataSet	
entries	[Charts.ChartDataEntry]	60 values	
_yMax	Double	80.135000000000005
_yMin	Double	49.600000000000001
_xMax	Double	17737.416666666668
_xMin	Double	16777.458333333332
```
Tried to swap min and max before building iterator if min > max and I got the chart display without a crash but when I scroll to the very end of it, the line disappear.

sebastienhannay:
Found a quick fix (available in above pull request) that is not perfect and doesn't solve the source of the issue but at least it doesn't crash anymore

k1ran-ak:
[...]

PR #4808, opened 2022-04-06
fix crash max < min when entryTo is not found

sebastienhannay:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4798

### Goals :soccer:
Fix the crash when entry to is not found max can be lower than min.
It's not a perfect fix as I still have issue with fill under charts line.
Would probably be better to fix the `entryForXValue` returning nil instead

As `entryFrom` is found but `entryTo` is not min may be higher than max. Just use the last index of datset instead of 0 to make sure max is always bigger than min

### Testing Details :mag:
Run integration into my app and the app doesn't crash anymore while building the chart.


playtopro:
Nice work, solved the issue for me :)

PR #4812, opened 2022-04-18
Remove Cocoadocs reference from README

rph8:
Cocoadocs does not exist anymore.
So I suggest remove the link from the README which could mislead a beginner.


pmairoldi:
Thanks!

Issue #4819, opened 2022-05-06
Invisible Line Chart When xAxis.axisMaximum > Data Maximum, Only on 4.x


giantramen:
* [ x ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Update to 4.0.2 from 3.6.

## What did you expect to happen?

Set an axis maximum and still have my line chart appear. 

## What happened instead?

Chart does not appear when xAxis.axisMaxiumum > data maximum.

xAxis.axisMaximum set to 200 -> 
![Simulator Screen Shot - iPhone 13 - 2022-05-06 at 09 11 02](https://user-images.githubusercontent.com/10697351/167150131-10cc06b4-56ec-46d6-b3eb-c6ff7bc30680.png)
![Simulator Screen Shot - iPhone 13 - 2022-05-06 at 09 10 56](https://user-images.githubusercontent.com/10697351/167150143-387093dc-d7e7-47c3-8d14-711249708ee2.png)

## Charts Environment

**Charts version/Branch/Commit Number:** 4.0.2
**Xcode version:** 13.3
**Swift version:** 5
**Platform(s) running Charts:** iOS
**macOS version running Xcode:** 12.3.1 (21E258)

## Demo Project

https://github.com/giantramen/Charts/commit/aea80980947b6033c97aa94489704149dc830f5a

Issue #4792, opened 2022-03-13
Candlestick Chart Issue with Version 4.0.2 (Pod was updated from 3.6.0)


Spencer-Villarreal:
I run 'pod update' about every single day just to make sure I am up-to-date on the latest releases. Today the Charts pod got updated from version 3.6.0 to 4.0.2 and now my candlestick chart is unusable.

When I was using the 3.6.0 version I followed the swift demo to get everything set up, but once I updated to 4.0.2 the whole chart does not show data correctly anymore. As of right now, only one candlestick is showing on the far left side and no other candles are shown on the screen. I cannot tell if all of the candles are being rendered on the far left side or if only one candle is being shown.

There are 19 chart values but only 1 value is being shown on the far left side. I set the x-axis to be the same amount of chart values + 2 for some spacing.

I have downgraded back to 3.6.0 and it goes back to normal/works completely fine. Does anyone know what could possibly be happening from the update? 

If it is possible could you create a new candlestick chart demo for the new 4.0.0 update?

dborelli:
I have the same situation with a multiple-set line chart. There was some sort of breaking change that's not apparent to me in the 4.x update. The demo version of the multiple line charts still works correctly, but it's the same example code I used in my original implementation.

technocraticOptimist:
Same issue. When will we see a fix for this?

blazarz:
Here's the issue and (not thoroughly tested) fix. Sorry, I'm not a Swift guy, so I'm not comfortable pulling/posting this fix...

The drawDataset method in CandleStickRenderer.swift sets the min, max and range values for _xBounds
```
_xBounds.set(chart: dataProvider, dataSet: dataSet, animator: animator)
```
The XBounds setter is found in BarLineScatterCandleBubbleRenderer.swift.  The entryTo value is what's not getting set correctly. 
```
let entryTo = dataSet.entryForXValue(high, closestToY: .nan, rounding: .up)
```
This method calls the entryIndex method in ChartDataSet.swift that tries to find the closest X value. 
```
open override func entryForXValue(
        _ xValue: Double,
        closestToY yValue: Double,
        rounding: ChartDataSetRounding) -> ChartDataEntry?
    {
        let index = entryIndex(x: xValue, closestToY: yValue, rounding: rounding)
        if index > -1
        {
            return self[index]
        }
        return nil
    }
```

The problem is that it's greater than the number of dataset elements, so it returns -1.
[...]

Issue #4753, opened 2021-11-17
spaceMin and spaceMax (xAxis) not working when setting new data 


francyrho11:
## What did you do?

I'm using `xAxis.spaceMin` and `xAxis.spaceMax` to add some extra spacing. 
axisMaximum are calculated automatically. 

## What did you expect to happen?

spaceMin and spaceMax will add extra spacing even when setting new data

## What happened instead?

Spacing is correct first time i set some data inside the graph.
Once I set new data inside the graph (e.g. `lineChartView.data = chartData`) the additional space disappear.

Rajneesh071:
Is it solved ? Because It's not working for me in latest code.

Only spaceMin working for me.
e.g. chartView.xAxis.spaceMin = 10

If I add Max, line in not drawing.

Issue #4739, opened 2021-10-15
Scatter chart:  Only the first point is showing a value, none of the other ones are


MichaelMeinberg:
* [ X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

_I have a scatter chart. with many points (maybe 20), and no matter what data I send it only shows the value on the chart for the first dot._ 

_I have tried dataSet.drawValuesEnabled = true, did not make a difference
Some values are large, but some are small, even down to 1 digit._

## What did you expect to happen?

_Values showing on every dot._

## What happened instead?

_Value shows on first dot each time, no others._

## Charts Environment

Charts:  4.0.1 - master branch
Xcode:  13.0
Swift: 5.0
Plateform:  iOS - iPhone
Mac OS:  Big Sure (11.4)

## Demo Project

_Can't do that, corporate code._

![image](https://user-images.githubusercontent.com/4966291/137518207-908e06e9-9651-433f-b2e8-7cc2761396e8.png)

MichaelMeinberg:
![image](https://user-images.githubusercontent.com/4966291/137516594-da245ae5-8b47-4ae4-9966-bc6cae13e3a0.png)

PR #4829, opened 2022-05-26
Fix chart only drawing first entry

FelixHerrmann:
### Issue Link :link:
#4819 #4792 #4753 #4739 #4722 #4699 (and possibly a lot more)

### Goals :soccer:
Fixing line chart only drawing first entry when spaceMax on xAxis is not 0.
This will fix the candle stick chart only drawing first value as well!

With the addition of the swift-algorithms package in v4.0.0 many search algorithms have changed, including the one in [ChartDataSet.entryIndex(x:closestToY:rounding:)](https://github.com/danielgindi/Charts/blob/b38b8d45a8cbda9f0f2a3566778ed114f06056b7/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift#L214).

It turns out that [partitioningIndex](https://github.com/apple/swift-algorithms/blob/195e0316d7ba71e134d0f6c677f64b4db6160c46/Sources/Algorithms/Partition.swift#L188) works different in one situation: when the given value is greater than the greatest value in the array.
Screen Shot 2022-05-26 at 23 39 32

Because of that, we need some kind of check that ensures we don't get an index out of bounds.
#4577 #4721 and the initial implementation #4497 do those checks, but wrong. The initial implementation is the best, the index is just one too high as shown in the screenshot above. We simply need to return the last possible index if a higher value is requested. Returning -1 in this situation will result in only drawing the first value!

In addition to that, I replaced all the raw index manipulation with the standard lib APIs and fixed Package.resolved to reflect the versions from the Package.swift!

### Testing Details :mag:
I tested in the demo projects and my personal one.

Before Fix | After Fix
:---------:|:---------:
Simulator Screen Shot - iPhone 13 Pro - 2022-05-26 at 22 24 17 | Simulator Screen Shot - iPhone 13 Pro - 2022-05-26 at 22 25 51
Simulator Screen Shot - iPhone 13 Pro - 2022-05-26 at 22 24 52 | Simulator Screen Shot - iPhone 13 Pro - 2022-05-26 at 22 25 32

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -240,10 +240,13 @@ open class ChartDataSet: ChartBaseDataSet
             // The closest value in the beginning of this function
             // `var closest = partitioningIndex { $0.x >= xValue }`
             // doesn't guarantee closest rounding method
-            if closest > 0 {
+            if closest > startIndex {
                 let distanceAfter = abs(self[closest].x - xValue)
-                let distanceBefore = abs(self[closest-1].x - xValue)
-                distanceBefore < distanceAfter ? closest -= 1 : ()
+                let distanceBefore = abs(self[index(before: closest)].x - xValue)
+                if distanceBefore < distanceAfter
+                {
+                    closest = index(before: closest)
+                }
                 closestXValue = self[closest].x
             }
         }
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -217,7 +217,7 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var closest = partitioningIndex { $0.x >= xValue }
-        guard closest < endIndex else { return rounding == .closest ? (endIndex-1) : -1 }
+        guard closest < endIndex else { return index(before: endIndex) }
 
         var closestXValue = self[closest].x
 
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/apple/swift-algorithms",
         "state": {
           "branch": null,
-          "revision": "bb3bafeca0e164ece3403a9de646b7d38c07dd49",
-          "version": "0.0.2"
+          "revision": "b14b7f4c528c942f121c8b860b9410b2bf57825e",
+          "version": "1.0.0"
         }
       },
       {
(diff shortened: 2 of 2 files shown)


TKZwo:
I've tested our apps' graphs with the results from this PR and I can confirm this solves our problem. Good job! I'll pull the updated version in our app as soon as it get available in a release.

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -217,7 +217,7 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var closest = partitioningIndex { $0.x >= xValue }
-        guard closest < endIndex else { return rounding == .closest ? (endIndex-1) : -1 }
+        guard closest < endIndex else { return index(before: endIndex) }
```
fumoboy007, line 220: Hmm this doesn’t respect the rounding mode though. For example, if `rounding == .up`, there should be no match, right?

TKZwo: I'm sure the rounding should not be applied when we are looking for an x higher than the highest in the set. 

The rounding is used for choosing between two x's when in range. Looking for x = 1.0 in [0, 1.5, 3, 4.5, 6, 7.5 ] should return index 0 when down, index 1 when up and index 1 for closest. Looking for 9.9 should return index 5

Returning a '-1' for both up or down when we hit the last in the range seems invalid.

artalejo:
It would be highly appreciate it if this fix can make it to the next release.

bzeeman:
Any thoughts on when this fix might be merged?

FelixHerrmann:
Maybe @pmairoldi can tell us when we can expect this fix to be merged?

morluna:
Any word on when this can move forward @pmairoldi ? We ran into this same issue and can confirm this change fixes it.

FelixHerrmann:
Awesome @cbalsalobre, thanks for confirming!

FelixHerrmann:
A note for the Collaborators: #4839 and #4844 try to fix the same problem but they don't resolve the root cause of the issue

417-72KI:
We are waiting for merging this for a long time.

417-72KI (approved):

TKZwo (approved):
This fix should be in before anything else.

TKZwo:
@fumoboy007 I think if you edit your comment to 'approve' the 'open discussion' will be closed and this can will be pulled in.

pmairoldi:
Thanks for the contribution. Will be out with the next version for Xcode 14

santieduardo:
Hello @pmairoldi 

Do you know when the next release of this lib will be available?

Thanks

PR #4839, opened 2022-06-12
Fix bug in `XBounds` calculation of minimum/maximum visible entry index.

fumoboy007:
`XBounds` looks for data entries on the boundary of or outside of the visible range. However, there may not be any such entries for a particular data set.

Imagine a chart with two data sets, one of which has a bigger range of x-values than the other. The chart’s zoomed-out visible range would be the range of the wider data set. When the renderer tries to render the narrower data set, it would fail to correctly calculate the starting and ending entry indices because the starting/ending entries are strictly inside of the visible range.

The fix is to add fallback logic to look for data entries inside of the visible range if the initial algorithm does not find anything.

The `XBounds` bug fix also requires a bug fix in `ChartDataSet`’s binary search algorithm.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -217,7 +217,7 @@ open class ChartDataSet: ChartBaseDataSet
         rounding: ChartDataSetRounding) -> Int
     {
         var closest = partitioningIndex { $0.x >= xValue }
-        guard closest < endIndex else { return rounding == .closest ? (endIndex-1) : -1 }
+        guard closest < endIndex else { return [.down, .closest].contains(rounding) ? (endIndex-1) : -1 }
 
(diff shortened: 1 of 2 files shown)


k1ran-ak (commented):
Reviewed. No crashes. I didn't see any Lower bound <= upper bound errors anymore

FelixHerrmann:
This conflicts with #4829 where I fixed the partitioningIndex algorithm change

fumoboy007:
@FelixHerrmann Cool! Confirmed that #4829 also partially fixes the `XBounds` calculation. However, I [commented](https://github.com/danielgindi/Charts/pull/4829/files#r902186856) on part of the change, which I believe is incorrect.

truongtantri:
Please merge this commit soon. thanks.

lundjrl:
Hey all, any updates on merging this in?

FelixHerrmann:
@lundjrl should be fixed with #4829

PR #4841, opened 2022-06-15
fixed no used warning

eren-celik:
### Goals :soccer:

there will be no warning in demo project

As I noticed in the demo project, it was giving a warning that the addEntryOrdered(_ e: ChartDataEntry) function is not used. That's why I made the development


pmairoldi (changes requested):
I think we should just change the function to return void instead. I can't think of a case where we need to know if the result had been added successfully.

PR #4844, opened 2022-06-15
CandleStickChart Correction of maximum element index

pppiroto:
When retrieving the element that approximates the maximum value on the x-axis, the index range is exceeded at the maximum value.

### Issue Link :link:
CandleStickChart, including the demo, only shows the first data.

### Testing Details :mag:
Demo application, visual confirmation only by moving the slider left and right.


FelixHerrmann:
Duplicate and a bit different implementation of #4829

PR #4861, opened 2022-07-07
Update ChartDataSet.swift

batikansosun:
Missing method.

Xcode 14 Beta 3 issues. 
Unavailable instance method 'replaceSubrange(_:with:)' was used to satisfy a requirement of protocol 'RangeReplaceableCollection'

### Issue
Xcode 14 Beta 3 issues. 
Unavailable instance method 'replaceSubrange(_:with:)' was used to satisfy a requirement of protocol 'RangeReplaceableCollection'

### Goals :soccer:
Instance method 'replaceSubrange(_:with:)' need to adding.

Added unavailable instance method 'replaceSubrange(_:with:)'.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -435,6 +435,11 @@ extension ChartDataSet: RandomAccessCollection {
 
 // MARK: RangeReplaceableCollection
 extension ChartDataSet: RangeReplaceableCollection {
+    public func replaceSubrange<C>(_ subrange: Swift.Range<Index>, with newElements: C) where C : Collection, Element == C.Element {
+        entries.replaceSubrange(subrange, with: newElements)
+        notifyDataSetChanged()
+    }
+    
     public func append(_ newElement: Element) {
         calcMinMax(entry: newElement)
         entries.append(newElement)


dingtianran:
Nice work
🤞

magneroald:
Thanks for the fix.
I also get this compilation erro ron xcode 14 beta 3:
Referencing instance method 'makeIterator()' on 'Collection' requires the types 'IndexingIterator<ChartData>' and 'IndexingIterator<RadarChartData>' be equivalent

Any solution to this?

pmairoldi:
Thanks for the contribution!

grangej:
Can we please have this one fix back ported to 3.x ?

pmairoldi:
We will not patch older versions of the library for newer Xcode versions. Your best bet is to update or fork the library.

linxiangyang112:
Can you change the minimum version to 10.0

pmairoldi:
I believe there is a technical reason behind it so we cannot. Is there a reason you still need to support such an old iOS version? The market share seems to indicate very low to 0 usage of this version https://developer.apple.com/support/app-store/

EarthHeroOrg:
@linxiangyang112 we use 10.3 and found a fix in the conversation here:

https://github.com/danielgindi/Charts/issues/4903

simson2010:
Hi, did anyone know about this change, I believe that this code works before without this update.

PR #4869, opened 2022-07-22
Update README.md

heydoy:
Fix little typos


pmairoldi:
I think it is grammatically correct like it is also. I don't think we need to use the oxford comma.

Issue #486, opened 2015-10-20
Cannot distribute inhouse build with ios-charts library embedded to my project


andrejbroncek:
Hello,

first of all, I would like to thank you for creating this library. I have embedded it in my project (as written in the README file) by dragging it into the project, embedding it in the binaries and so on. Target iOS version is 8.0.

Everything worked great for dev version of the app, but as soon as I built the inhouse version, app crashes after initial welcome screen shows. Logs for the crash are as following:

```
Oct 20 15:54:15 iPhone SpringBoard[54] <Error>:  SecTrustEvaluate  [leaf IssuerCommonName SubjectCommonName]
Oct 20 15:54:15 iPhone SpringBoard[54] <Error>:  SecTrustEvaluate  [leaf IssuerCommonName SubjectCommonName]
Oct 20 15:54:15 iPhone kernel[0] <Notice>: xpcproxy[700] Container: /private/var/mobile/Containers/Data/Application/6D451583-EE0E-4F0A-9CA9-2E33F83D6138 (sandbox)
Oct 20 15:54:15 iPhone kernel[0] <Notice>: AMFI: energy_challenge(pid 700) - [deny-mmap] mapped executable file has no team identifier in its signature: /private/var/mobile/Containers/Bundle/Application/84AEA7F2-4987-4119-B3CC-A4EAFC9989DC/energy_challenge.app/Frameworks/Charts.framework/Charts
Oct 20 15:54:15 iPhone com.apple.xpc.launchd[1] (UIKitApplication:com.edge5.nativeMenuEnergieInhouse[0x1f2e][700]) <Notice>: Service exited due to signal: Trace/BPT trap: 5
Oct 20 15:54:15 iPhone assertiond[63] <Warning>: Unable to obtain a task name port right for pid 700: (os/kern) failure (5)
Oct 20 15:54:15 iPhone SpringBoard[54] <Warning>: Unable to register for exec notifications: No such process
Oct 20 15:54:15 iPhone SpringBoard[54] <Warning>: Unable to obtain a task name port right for pid 700: (os/kern) failure (5)
Oct 20 15:54:15 iPhone SpringBoard[54] <Warning>: Unable to obtain a task name port right for <FBApplicationProcess: 0x13226dc00; com.edge5.nativeMenuEnergieInhouse; pid: 700>
[...]

PR #4881, opened 2022-08-14
Fix Type 'ChartDataSet' does not conform to protocol 'RangeReplaceabl…

kanstantsin-bucha:
For some reason I got this error on xCode - beta.

Fix implemented by:
* implementing method of RangeReplaceableCollection protocol for ChartDataSet 
* renaming of Range(objc ChartsRange) class cause it Shadows swift Range class that leads to compile errors.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
@@ -527,7 +527,7 @@ extension ChartDataSet: RandomAccessCollection {
 }
 
 // MARK: RangeReplaceableCollection
-extension ChartDataSet: RangeReplaceableCollection {
+extension ChartDataSet: RangeReplaceableCollection {    
     public func append(_ newElement: Element) {
         calcMinMax(entry: newElement)
         entries.append(newElement)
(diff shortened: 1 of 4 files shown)


josipbernat:
Can somebody merge this PR? iOS 16 is near and this error breaks my app. Thanks!

emadhegab:
@danielgindi  I think this one is top priority if you don't mind

richardsimkins:
This issue is also happening on 3.6.0 with Xcode 14 beta 6. Will merging this fix also fix it for people not on version 4.x.x?

Source/Charts/Data/Implementations/Standard/ChartDataSet.swift
```diff
@@ -571,4 +571,12 @@ extension ChartDataSet: RangeReplaceableCollection {
         entries.removeAll(keepingCapacity: keepCapacity)
         notifyDataSetChanged()
     }
+    
+    public func replaceSubrange<C>(
+        _ subrange: Range<Index>,
```
pmairoldi, line 576: I think if we just use `Swift.Range<Index>` here we don't have to rename the other Range class. I think it would be a better fix for the problem.

pmairoldi (changes requested):
The change works but I'd rather not change the `Range` class's name.

pmairoldi:
Close in favour of https://github.com/danielgindi/Charts/pull/486. Thanks for the contribution.

zygoat:
#486 is seven years old and unrelated. I presume that superseding issue number is a typo?

pmairoldi:
Ahhhh sorry about that it was https://github.com/danielgindi/Charts/issues/4861

Issue #4895, opened 2022-09-04
Trying to compile via SPM  on ios 16 beta swift 5.7 xCode 14 beta 5- fails


benchuk:
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag [`ios-charts`.](https://stackoverflow.com/questions/tagged/ios-charts)
>
> Please remove this line and everything above it before submitting.

* [ ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?
Try to compile via SPM on xcode 14 beta

## What did you expect to happen?

will be able to compile

## What happened instead?

does not compile 
getting error 'Type 'ChartDataSet' does not conform to protocol 'RangeReplaceableCollection'

## Charts Environment

**Charts version/Branch/Commit Number:*master*
**Xcode version:*xcode14 beta 5*
**Swift version:*5.7 beta on xcode 14 ios 16*
**Platform(s) running Charts:*ios 16*
**macOS version running Xcode:* macOS ventura beta *

## Demo Project

N/A

funkenstrahlen:
I can confirm this. I have the same issue.

gk-brown:
Same here - just installed Xcode 14 RC and can no longer build Charts. I was on 3.6.0 and thought that updating to 4.0.3 would help, but it did not.

MyHZ:
`// MARK: RangeReplaceableCollection
extension ChartDataSet: RangeReplaceableCollection {
    public func replaceSubrange<C>(_ subrange: Swift.Range<Index>, with newElements: C) where C : Collection, Element == C.Element {
        entries.replaceSubrange(subrange, with: newElements)
        notifyDataSetChanged()
    }
`

xjki:
See https://github.com/danielgindi/Charts/issues/4897#issuecomment-1240652977

jaytrisw:
This issue was fixed here: https://github.com/danielgindi/Charts/pull/4861

I can confirm that specifying master is a work around until there is a new release.

itayAmza:
[...]

PR #4902, opened 2022-09-12
Update ChartDataSet.swift

billbonney:
Fixes the issue of the Missing method when compiling in Xcode 14 for Charts 3.6.0 by cherry-picking

Will need to be merged against a release branch based on tag 3.6.0

**NOTE: This is for the version 3 release of Charts to avoid the need to migrate to 4.0** 

### Issue Link :link:
see #4895


pmairoldi:
We do not want to support 2 major version at once. We do not have enough resources to do so. I'd also like to point out that the fix you are making is on master not v3.6.0.

billbonney:
Thanks. I've just posted this back here as a 'hot-fix' for critical build issue. Will be migrating to 4.0 in future. 👍

PR #4942, opened 2022-10-05
Fix RangeReplaceableCollection protocol implementation for 3.6 version

flaviosuardi:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/4873

### Goals :soccer:
I fix problem related to RangeReplaceableCollection protocol implemented on ChartDataSet adding replaceSubrange method for 3.6 version.

### Testing Details :mag:
I just rebuild project and no build errors occur on Xcode 14.0.

PR #4974, opened 2022-11-23
Fixing errors in xcode 14

saddam-csk:


liuxuan30:
dup to https://github.com/danielgindi/Charts/pull/4942

Issue #4960, opened 2022-11-03
LineChartRenderer drawHighlighted(context: CGContext, indices: [Highlight]) crash


AlexandrKovtunenkoDevX:
## What did you do?

[presumably] Using `HorizontalBarChartView` with live updates  user taps chart view

## What did you expect to happen?

All bars from chart view top to touch location.y becomes highlighted

## What happened instead?

App crashes on `guard let set = lineData[high.dataSetIndex] as? LineChartDataSetProtocol` with "index out of range" exception

## Charts Environment

Charts (4.1.0)
Xcode v.14.0.1
Swift 5
Platform: iOS
macOS v.12.5.1

## Demo Project

Wasn't able to reproduce the crash myself, just received crash reports from Firebase Crashlytics and through XCode reports.
Reproduce rate: [presumably] low, 6 events during first week after release
Crash Log example:
`Date/Time:           2022-10-16 20:48:11.7358 -0500
Launch Time:         2022-10-16 20:43:46.4317 -0500
OS Version:          iPhone OS 15.6.1 (19G82)
Release Type:        User
Baseband Version:    1.70.01
Report Version:      104

Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x000000010151ecb4
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: SIGNAL 5 Trace/BPT trap: 5
Terminating Process: exc handler [72164]

Triggered by Thread:  0

Kernel Triage:
VM - pmap_enter failed with resource shortage

Thread 0 name:
Thread 0 Crashed:
0   Charts                        	0x000000010151ecb4 Swift runtime failure: Index out of range + 0 (<compiler-generated>:0)
1   Charts                        	0x000000010151ecb4 specialized _ArrayBuffer._checkInoutAndNativeTypeCheckedBounds(_:wasNativeTypeChecked:) + 4 (<compiler-generated>:0)
2   Charts                        	0x000000010151ecb4 specialized Array._checkSubscript(_:wasNativeTypeChecked:) + 4 (<compiler-generated>:0)
3   Charts                        	0x000000010151ecb4 specialized Array.subscript.getter + 4 (ChartData.swift:452)
4   Charts                        	0x000000010151ecb4 ChartData.subscript.getter + 4 (<compiler-generated>:0)
5   Charts                        	0x000000010151ecb4 LineChartRenderer.drawHighlighted(context:indices:) + 1312
[...]

PR #4981, opened 2022-12-06
Added Safe index subscript logic

tringbhagya:
### Issue Link :link:
[Crash Issue](https://github.com/danielgindi/Charts/issues/4960
)

- Added safe subscript logic.

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/ChartData.swift
+++ b/Source/Charts/Data/Implementations/Standard/ChartData.swift
@@ -452,6 +452,10 @@ extension ChartData: MutableCollection
         get { return dataSets[position] }
         set { self._dataSets[position] = newValue }
     }
+    
+    public subscript(safe index: Index) -> Element? {
+           return indices.contains(index) ? self[index] : nil
+    }
 }
 
 // MARK: RandomAccessCollection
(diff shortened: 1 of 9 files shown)


liuxuan30:
what's the effect that adding safe? just return nil?

I would think adding such will only hide the issue and harder to find and make a fix for calculating highlight bounds correctly. Crashes always mean there is a defect especially OOB. Getting a real fix sometimes is better than hiding it. If there is no crash, then it's harder to notice then highlight is broken or mis-displaying in the field without any notice.

But I'm open to add this.

liuxuan30:
we are working on release 5.0 to fix a bunch of compile errors and renaming the library. Probbaly have to revisit this PR after it's merged, or, you can target the release/5.0.0 branch and let @pmairoldi carry it in

tringbhagya:
@liuxuan30 Thank you for the response. I'll check release/5.0.0 branch and include my changes.

tringbhagya:
Hi @liuxuan30 
It seems this crash is still occurring with latest release code. I've created a new pull request with latest release code. Please check it.
https://github.com/ChartsOrg/Charts/pull/5200

PR #5005, opened 2023-01-20
iOS 11 Support

furkansandal:


pmairoldi:
Sorry we do not have the bandwidth to also support iOS 11.

Issue #4845, opened 2022-06-15
Library name conflict with Apple Charts


remlostime:
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag [`ios-charts`.](https://stackoverflow.com/questions/tagged/ios-charts)
>
> Please remove this line and everything above it before submitting.

* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

There's apple Charts now. If we import Charts , when does the Xcode know it's apple one or third party one?

## What did you expect to happen?

ℹ Please replace this with what you expected to happen.  

## What happened instead?

ℹ Please replace this with of what happened instead.  

## Charts Environment

**Charts version/Branch/Commit Number:**
**Xcode version:**
**Swift version:**
**Platform(s) running Charts:**
**macOS version running Xcode:**

## Demo Project

ℹ Please link to or upload a project we can download that reproduces the issue.

waterskier2007:
Following for status updates. Personally what I did is fork the repo and then rename the library so that it didn't conflict with the native implementation.

liuxuan30:
ok.. this sucks, @jjatie @danielgindi @pmairoldi thoughts? What would be the appropriate name now?

waterskier2007:
> Also one question- does this happen also when _not_ adding the Swift Charts framework to the project?

It does not occur if you do not use the Swift Charts framework

waterskier2007:
Perhaps a name such as UIKitCharts (just off the top of my head) or something else to indicate that the implementation is in UIKit?

waterskier2007:
> I doubt someone would use both at the same time. This seems to be implemented in swift 5.7 so maybe this could work https://github.com/apple/swift-evolution/blob/main/proposals/0339-module-aliasing-for-disambiguation.md

I disagree. We have a large app we support that has some "legacy" views using UIKit that have charts in them but we would also like to start adding Swift Charts to our SwiftUI views once iOS 16 rolls out. I imagine that use case could be common.

alanzeino:
[...]

PR #5009, opened 2023-02-07
Update library name

waterskier2007:
### Issue Link https://github.com/danielgindi/Charts/issues/4845

Fixes naming conflict with Apple's Charts framework

### Goals :soccer:

Update the Swift package name
Update the podspec
Update the directory structure to match
Update the Charts xcode project to use a folder-based structure instead of Groups

Implemented each step in a single commit so it's easier to follow


liuxuan30:
Thanks for your contritution! this is what we exactly needed!

however, UIChart seems unpossible since it's reserved by Apple already. Would mind change to DGCharts ?

liuxuan30:
let's make sure it biuld on xcode 14 as well

rineek:
Shouldn't the directory names also be updated from `UICharts` to `DGCharts`?

waterskier2007:
Wow, yep that was a big mistake on my part - updated.

zygoat:
What about the `xcodeproj` and `podspec` filenames; wouldn't it be idiomatic for them to match the project name too?

waterskier2007:
I am taking a look at why the snapshot test resources were removed. That was not intentional

pmairoldi:
I am looking at this today and I do not think we need to rename all the source files. We can just change the product name. I will take some of these changes but leave the code reorg to the side. It will make maintainability easier for us to not reorg everything.

pmairoldi:
Closed by #5027

Issue #5023, opened 2023-03-17
MarkerView.viewFromXib possible crash


RCCoop:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Added a `LineChartViewController` to a Mac app, with a `MarkerView` subclass as its marker. 

## What did you expect to happen?

The marker would show up when a value on the chart was selected.

## What happened instead?

App crashed due to trying to force-unwrap a nil value.

## Charts Environment

**Charts version/Branch/Commit Number:** v4.1.0
**Xcode version:** 14.2
**Swift version:** .57
**Platform(s) running Charts:** MacOS
**macOS version running Xcode:** 13.2.1

## Solution:

I found the offending line of code in `MarkerView`'s `viewFromXib(in:)`:
```
if bundle.loadNibNamed(
    NSNib.Name(String(describing: self)),
    owner: nil,
    topLevelObjects: &loadedObjects)
{
    return loadedObjects?[0] as? MarkerView
}
```

What I found was that `loadedObjects` contains the `MarkerView`, but not always at index 0 (in my tests, there are always two elements in the array, one of which is an instance of `NSApplication`). So the issue can be fixed by changing the `return` line to:
```
return loadedObjects?.compactMap({ $0 as? Self }).first
```
(I also changed the return value of `viewFromXib` to `Self`, but that change could be omitted for a smaller change)

I can make a PR for this. Not sure if I should make the same change to the `#if !os(OSX)` section of the function, though. Any guidance on that would be appreciated.

yongfrank:
Same!

PR #5030, opened 2023-03-23
Marker view mac

RCCoop:
### Issue Link :link:
https://github.com/danielgindi/Charts/issues/5023

### Goals :soccer:
Make `MarkerView` work in MacOS, and add an example of it into the MacOS sample app.

- Several non-breaking changes in `MarkerView`
  - Added `didSet` in `chartView` property, which calls a new function `didAddToChart(_:)`.
  - `didAddToChart(_:)` only does anything in OSX. It adds the `MarkerView` as an off-screen subview of the `chartView`'s containing view. This is necessary because the `MarkerView` would not render in OSX without it having been added to a view already. See this [SO answer](https://stackoverflow.com/a/75773084/1275947) for more context.
  - Changed the implementation of `viewFromXib(in:)` to return `Self` rather than `MarkerView` so that subclasses will return as their own type rather than the base class.
  - Also changed the implementation of `viewFromXib(in:)` to find the first instance of `Self` in the loaded objects from the xib rather than assuming that the first object in that array was the one we'd be looking for. In MacOS this was a problem because the view was often second in the xib's loaded objects rather than the first.
  - Finally, in the MacOS implementation of `viewFromXib(in:)`, set `wantsLayer` to true for the resulting view, which was also necessary to render the view during `draw(context:point:)` as referenced in previous StackOverflow answer.

### Testing Details :mag:

- Created `RadarMarkerViewMac` in the MacOS sample app, which is nearly identical to the `RadarMarkerView` in the iOS sample app.
- Added the new `RadarMarkerViewMac` as the marker in the MacOS sample's `RadarDemoViewController`.
- Ran the MacOS sample app and made sure the `RadarDemoViewController` displayed the marker as expected.


yongfrank:
Thank you for your commit. It very useful for me to develop with DGCharts on Mac!

It's strange that your PR fixing the bug has not been merged for a long time.

yongfrank:
@liuxuan30 Hi, I find the same issue with Marker View and xib on Mac when I was developing. #5023

PR #5031, opened 2023-03-23
better

gsbernstein:
### Goals :soccer:

Jagged lines stick out from behind the data points

Before | After
--- | ---
jagged|Simulator Screen Shot - iPhone 14 - 2023-03-23 at 11 10 00

Because this disconnects each line segment, I limited it only to the case where there's a data point circle covering the line join. If we want to be stricter about that check, we could also make sure that the circle size is greater than the line width, and that the color of the circle is fully opaque.

I also have a fix for the case where the disjoint isn't covered up, see `improve-jagged-lines-extra`

### Testing Details :mag:
 - demo app


gsbernstein:
@pmairoldi — what is the best way to get a review for this? Should I update my commits? Do I need an issue?

PR #5043, opened 2023-04-10
Fix Node 12 deprecation warning

FelixHerrmann:
### Goals :soccer:

The GitHub checkout action v2 uses Node.js 12. This version of Node is deprecated and will be fully removed in summer 2023, see [this blog post](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).
Screenshot 2023-04-10 at 15 55 30

I have updated all checkout actions to us v3 instead of v2.

### Testing Details :mag:
I couldn't test it with this repo, but I have not experienced any issues with the other workflows where I have done the upgrade.

_Edit_: CI is not failing and the warnings are gone: https://github.com/danielgindi/Charts/actions/runs/4658235640


pmairoldi:
Nice catch. Thanks.

Issue #4938, opened 2022-09-27
Building via Carthage with BUILD_LIBRARY_FOR_DISTRIBUTION = YES fails


djs-code:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/jjatie/Charts/blob/master/CONTRIBUTING.md).

## What did you do?

Attempted to build Charts 4.1.0 via Carthage with `BUILD_LIBRARY_FOR_DISTRIBUTION` set to `YES`.  The following script was used:

```
# carthage.sh
# Usage example: ./carthage.sh build --platform iOS

set -euo pipefail

xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
trap 'rm -f "$xcconfig"' INT TERM HUP EXIT

# For Xcode 12 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
# the build will fail on lipo due to duplicate architectures.

CURRENT_XCODE_VERSION=$(xcodebuild -version | grep "Build version" | cut -d' ' -f3)
echo "EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_$CURRENT_XCODE_VERSION = arm64 arm64e armv7 armv7s armv6 armv8" >> $xcconfig

echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200 = $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_$(XCODE_PRODUCT_BUILD_VERSION))' >> $xcconfig
echo 'EXCLUDED_ARCHS = $(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT)__XCODE_$(XCODE_VERSION_MAJOR))' >> $xcconfig

echo 'BUILD_LIBRARY_FOR_DISTRIBUTION = YES' >> $xcconfig

export XCODE_XCCONFIG_FILE="$xcconfig"
carthage "$@"
```

[...]

Issue #4919, opened 2022-09-15
SwiftAlgorithms Bitcode_disabled


afridikhalid:
## What did you do?

ℹ I did a pod update with charts '~> 4.1.0`

## What did you expect to happen?

ℹ I expected that everything should work smoothly. but found out that Charts 4.1.0 has many breaking changes. which I have now updated everything. no more errors regarding Charts it self

## What happened instead?

ℹ When I use Pods install it installs SwiftAlgorithms together with Charts 4.1.0 and that's when I try to run the project I get the error that SwiftAlgorithms should enable bitcode and I don't know how to change that.

## Charts Environment

**Charts version/Branch/Commit Number:** 4.1.0
**Xcode version:**  14.0
**Swift version:** 5
**Platform(s) running Charts:**  iOS

afridikhalid:
I had to iterate through all my targets and set ENABLED_BITCODE  to true and now everything works :)

areejsadaqa66:
I am still facing a  problem with SwiftAlgorithms some files showing errors as: 'let' property 'base' may not be initialized directly; use "self. init(...)" or "self = ..." instead

and I tried to get all BITCODE to enable true and it still didn't work

mugikhan:
> 

I am having the same issue as bove

tejua:
I am having the same issue 
did you find any solutions? @mugikhan @afridikhalid

mounou19:
This should help bypassing the issue until SwiftAlgorithms get updated to support new swift 5.7 I think !
To add at the end of your Podfile

```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      if target.name == "SwiftAlgorithms"
        config.build_settings['BUILD_LIBRARY_FOR_DISTRIBUTION'] = 'NO'
      end
    end
  end
end
```

PR #5069, opened 2023-06-08
remove swift-algorithms package in favor of manully importing needed …

pmairoldi:
Removed the need for swift-algorithms.

We were using only 2 functions so I duplicated the source code into our codebase. We now own this and do not have dependencies.

Addresses #4919 & #4938


liuxuan30:
@pmairoldi, just asking, it seems builds fine, are you having problems still and have to remove the packages?

pmairoldi:
Don't think so. I removed everything.

liuxuan30:
what don't compile?

pmairoldi:
Nothing but it's more to fix the Bitcode and build for distribution errors. We use only 2 functions. I'd rather own the code than have a dependency for such a small thing.

PR #5097, opened 2023-08-28
starting from 2023

KostiantynKaniuka:
- Just recalculated the DayFormatter in the demo app for the Swift section to start from 2023.


pmairoldi:
These are just examples, it is ok to know have the most recent dates.

PR #5101, opened 2023-09-08
优化ViewPortHandler对isInBoundsTop的判断

big-zZ:
do the same as 'isInBoundsLeft' function, change 'y' to 'y + 1.0'


big-zZ:
to avoid something like  contentRect.y = 36.4 and tapPostion.y = 36.39999998

PR #5105, opened 2023-09-15
Shorten type checking times by shortening expressions and avoiding type conversions

idrougge:
I noticed that DGCharts amounted to a large portion of the compile times of our app, so I measured the times for evaluating expressions and function bodies by adding the below build flags:
```
-Xfrontend -warn-long-function-bodies=<ms>
-Xfrontend -warn-long-expression-type-checking=<ms>
```

It turned out that certain functions and expressions had evaluation times close to one second on an M1 Mac, so I added some help for the compiler to shorten those checking times.

As a result of these changes, the compile times have gone down from 6 s for the entire package to only 4 s — a 33 % decrease. Certain files still make out the bulk of the total compile time so a further refactoring of those files could enable higher parallelisation of compilation, but that would make this PR too big to review.

### Goals :soccer:
Shorter compile times.

1. Bracketing parts of certain long expressions such as `-0.5 * (sqrt(1.0 - position * position) - 1.0)`
2. Removing type conversions from `Double` to `Double` such as `Double( sqrt(1 - position * position) )`
3. Adding a `guard` to functions that consisted of just one big `if` statement
4. Adding a type specifier to variable declarations that instead used a type initialiser on a literal such as `var maxW: CGFloat = 0.0`

### Testing Details :mag:
The tests still run as before.
To ease code review, it helps to turn on "hide white spaces" in Github's code review view.

### DIFF
--- a/Source/Charts/Animation/ChartAnimationEasing.swift
+++ b/Source/Charts/Animation/ChartAnimationEasing.swift
@@ -211,25 +211,25 @@ internal struct EasingFunctions
     
     internal static let EaseInSine = { (elapsed: TimeInterval, duration: TimeInterval) -> Double in
         var position: TimeInterval = elapsed / duration
-        return Double( -cos(position * Double.pi / 2) + 1.0 )
+        return -cos(position * Double.pi / 2) + 1.0
     }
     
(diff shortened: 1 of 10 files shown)


drewster99:
Oof.  Similar work done in https://github.com/danielgindi/Charts/pull/5124.

PR #5109, opened 2023-09-29
Support for rounded bar chart and grid edges

swatiyadav009:

### DIFF
--- a/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift
+++ b/Source/Charts/Data/Implementations/Standard/BarChartDataSet.swift
@@ -51,6 +51,9 @@ open class BarChartDataSet: BarLineScatterCandleBubbleChartDataSet, BarChartData
     /// the overall entry count, including counting each stack-value individually
     private var _entryCountStacks = 0
     
+    /// the corner radius applied to each data set
+        public var cornerRadius: CGFloat = 0.0
+    
     /// Calculates the total number of entries this DataSet represents, including
     /// stacks. All values belonging to a stack are calculated separately.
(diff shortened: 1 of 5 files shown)


swatiyadav009:
Please do check my open PR for rounded slices for PieChart as well.
Link - https://github.com/danielgindi/Charts/pull/5114

gabors:
@swatiyadav009 Love this PR, few comments / wishes however:

- the corner radius is a percentage of the bar width, so it doesn't work for horizontal charts, as they end up with super rounded corners (see example)
- would it be possible to supply instead an absolute value for the corner radius in points?


image

crash481:
1) bottom of Bars also with CornerRadius, when usually it is only on top. Need to make for every edge
2) Selection highlight without corners
Screenshot 2024-03-28 at 20 46 28

gabors:
I actually don't mind the rounded edges on the bottom but yes the lack of rounding on the highlighted back has been annoying.

crash481:
@gabors it is popular design when bars in bottom starts without corner radius, and on the top corners is rounded

crash481:
I made this pull request, it solves problem better
https://github.com/ChartsOrg/Charts/pull/5163

gabors:
Agreed it's better to have options.

crash481:
@gabors oh, also your wish about implement corner radius in absolute values also there. ✌️

gabors:
@crash481 amazing. I can get rid of my fork then 😊

I'll test it out today. Thanks 🙏

gabors:
@crash481 on a different subject. How did you add that popup value ?

pmairoldi:

PR #5114, opened 2023-10-10
Support for rounded edges for pie chart slices

swatiyadav009:


swatiyadav009:
Please do check my open PR for rounded edges for BarChart and Grid as well.
Link - https://github.com/danielgindi/Charts/pull/5109

minhDTNHust:
@swatiyadav009: how to make it work with half pie chart? =>  two ends of the first and last slices are not be effected.
And, can I choose the end of slice to apply effect, not all left and right ends should be effected?

Issue #4414, opened 2020-07-09
Minimum angle for slices


emreond:
I wanted to have a minimum angle for the slices which are too small to be seen. I modify the values when creating `PieChartDataEntry` but It requires to modify all values. Is there any short way to achieve this? 

Android library `MPAndroidChart` has a `setMinAngleForSlices` function to achieve this. Is there a similar function in Charts too?

Xcode version: Xcode 12 - B2
Platform(s) running Charts: iOS

ManuEspeso:
Yes please, it will be awesome

chaaarly:
Any news about the `setMinAngleForSlices` method ?

Rajankumarchicmic:
please share any info about this.

DevanshVyasTR:
Any update on `setMinAngleForSlices`?

PR #5136, opened 2023-12-15
[issue: #4414 ] Added: minAngleForSlices

ibrahimakpinar:
it's a similar functionality to setMinAngleForSlices at Android.

When the percentages of some data are very small, they are either not visible or difficult to touch. For this, it is necessary to determine the minimum slice angle.

### Issue Link :link:
#4414 

### Testing Details :mag:


ibrahimakpinar:
when given an array like [50000, 1000, 10, 10, 10, 10]

when minAngleForSlices is 0 
image

after set minAngleForSlices as 5:

image

PR #5180, opened 2024-07-02
Fix min greater than max iterator

nikolascane:
Sometimes, the min value is greater than the max value. This discrepancy causes a crash.


Naveen-C-Ramachandrappa (approved):

ankushkushwaha (approved):

danielgindi:
It looks like you are only swapping local variables...

PR #5185, opened 2024-07-12
Add corner radius in Barchart

Khanak1521:

### DIFF
--- a/DGCharts.podspec
+++ b/DGCharts.podspec
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = "12.0"
   s.tvos.deployment_target = "12.0"
   s.osx.deployment_target = "10.13"
-  s.source = { :git => "https://github.com/danielgindi/Charts.git", :tag => "#{s.version}" }
+  s.source = { :git => "https://github.com/Khanak1521/Charts.git", :tag => "#{s.version}" }
   s.default_subspec = "Core"
   s.swift_version = '5.7'
   s.cocoapods_version = '>= 1.5.0'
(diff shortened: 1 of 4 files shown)


pelodi:
when this feature will be added and merged?

liuxuan30:
looks unlikely, the implementation of this will has an issue: when the bar is too thin, it will look like an Oval rather than a bar.

DGCharts.podspec
```diff
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = "12.0"
   s.tvos.deployment_target = "12.0"
   s.osx.deployment_target = "10.13"
-  s.source = { :git => "https://github.com/danielgindi/Charts.git", :tag => "#{s.version}" }
+  s.source = { :git => "https://github.com/Khanak1521/Charts.git", :tag => "#{s.version}" }
```
ankushkushwaha, line 11: I think we need to undo this git url change

Source/Charts/Renderers/BarChartRenderer.swift
```diff
@@ -379,7 +382,10 @@ open class BarChartRenderer: BarLineScatterCandleBubbleRenderer
                 context.setFillColor(dataSet.color(atIndex: j).cgColor)
             }
             
-            context.fill(barRect)
+            let rectCorners: UIRectCorner = [.topLeft, .topRight]
+            let bezierPath = UIBezierPath(roundedRect: barRect, byRoundingCorners: rectCorners, cornerRadii: CGSize(width: self.barChartCornerRedius, height: self.barChartCornerRedius)) //(roundedRect: barRect, byRoundingCorners: rectCorners, cornerRadi: 5)
```
ankushkushwaha, line 386: Do we need the commented code ?

pmairoldi:

PR #5198, opened 2024-09-02
Charts/issues/5197 - fixed host app crash by adding a check before ca…

sorinmiroiu97:
### Issue Link :link:
Link to issue: https://github.com/ChartsOrg/Charts/issues/5197

### Goals :soccer:
Fix the crash of the host app from within the library code whenever you get in a situation where the CGFloat calculated in the renderers' isDrawingValuesAllowed function are equal to CGFloat.infinity or to CGFloat.nan, as type casting those causes a crash.

Honestly it's pretty self explanatory and it's rather a dumb-value check, nothing crazy or complicated. The issue explains it all already.

### Testing Details :mag:
I didn't add any tests, as there's no point in testing something that i added because if fails at the compiler level so i don't think any tests are required, but please let me know.

### DIFF
--- a/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
+++ b/Source/Charts/Renderers/BarLineScatterCandleBubbleRenderer.swift
@@ -66,7 +66,9 @@ open class BarLineScatterCandleBubbleRenderer: NSObject, DataRenderer
     open func isDrawingValuesAllowed(dataProvider: ChartDataProvider?) -> Bool
     {
         guard let data = dataProvider?.data else { return false }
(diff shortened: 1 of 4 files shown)


Naveen-C-Ramachandrappa (approved):

sorinmiroiu97:
@Naveen-C-Ramachandrappa thanks for the approval. Do you know when we can expect the PR to be merged in? Thanks a lot 😄

danielgindi (approved):

sorinmiroiu97:
thanks a lot for approving and merging this in @danielgindi ! 😄

PR #5231, opened 2025-03-18
Fix crash in LineChartRenderer.swift when animating data set changes

StainlessStlRat:
When updating the data set for the chart and then animating it, sometimes highlights are generated for the
            // removed sets. This guard prevents OOB in the case where the new set has less data than the old set

### Goals :soccer:

Fix crash when updating datasets and animating


Naveen-C-Ramachandrappa (approved):

StainlessStlRat:
@chen1987lei can you make a new PR?

